Issue #11, opened 2023-01-09
Add jest as a unit testing suite for testing react components and code


phantomjinx:

phantomjinx:
Jest is currently missing and not configured so going to complete scaffolding to begin executing unit tests.

tadayosi:
@phantomjinx Thanks for the reporting.

Doesn't `yarn test` work for you?  Create React App uses [React Testing Library](https://testing-library.com/docs/react-testing-library/intro/) and Jest should be available as its dependency. For example, we already have something like these:
https://github.com/hawtio/hawtio-next/blob/main/src/hawtio/Hawtio.test.tsx
https://github.com/hawtio/hawtio-next/blob/main/src/hawtio/util/strings.test.ts

phantomjinx:
It originally complained that `jest` was not installed, despite a `yarn install`. I have added it for the moment and got most of the test suites passing (issues with `global.fetch` mocking mostly). I'll remove it and see if a `yarn install` does bring it back in.

PR #14, opened 2023-01-10
(#11): Small changes to jest configuration

phantomjinx:
* craco.config.js
 * Removes need for config parameter
 * Adds src/hawtio/test module to directories
 * Adds moduleNameMapper for react-markdown which was throwing an error on use of "export" token"

* package.json
 * Adds script for test coverage

* **/.ts
 * Adds support module for testing. Include any functions required by number of tests
 * Moves the mocking of global.fetch to its own function and adds handling for replacing with any default after tests are completed

### DIFF
--- a/craco.config.js
+++ b/craco.config.js
@@ -8,12 +8,32 @@ module.exports = {
     },
   },
   jest: {
-    configure: config => {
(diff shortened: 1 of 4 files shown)


tadayosi (approved):
thanks!

PR #37, opened 2023-01-27
Refactoring for better sharing of artifacts / Unit testing

phantomjinx:


src/hawtio/plugins/shared/mbean/node.ts
```diff
@@ -1,73 +1,12 @@
...
-    return this.tree.length === 0
-  }
-}
-
-const Icons = {
-  folder: React.createElement(FolderIcon),
-  folderOpen: React.createElement(FolderOpenIcon),
-  mbean: React.createElement(CubeIcon),
-} as const
+const log = Logger.get(`mbean-node`)
```
tadayosi, line 9: We use the naming convention `hawtio-<plugin>-<area>` for loggers. Let's follow the convention.

src/hawtio/plugins/shared/mbean/tree.ts
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    return filteredTree
+  }
+
+  private tree: MBeanNode[] = []
+
+  private constructor(id: string) {
+    this.id = id
+    this.log = Logger.get(`${id}-tree`)
```
tadayosi, line 45: We use the naming convention `hawtio-<plugin>-<area>` for loggers. Let's follow the convention.

src/hawtio/plugins/shared/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './components'
+export * from './mbean'
```
tadayosi, line 2: I'd prefer to name this folder as `tree` as the name is used in the legacy hawtio-integration as well.

src/hawtio/plugins/shared/workspace.ts
```diff
@@ -31,10 +31,11 @@ class Workspace {
       }
     }
     const value = await jolokiaService.list(options)
+
     // TODO: this.jolokiaStatus.xhr = null
     const domains = this.unwindResponseWithRBACCache(value)
     log.debug("JMX tree loaded:", domains)
-    return new MBeanTree(domains)
+    return MBeanTree.createMBeanTreeFromDomains('workspace', domains)
```
tadayosi, line 38: We use the naming convention `hawtio-<plugin>-<area>` for loggers. Let's follow the convention.

src/hawtio/plugins/shared/mbean/node.ts
phantomjinx: Yeah. Was not happy with this after posting the PR yesterday. Since added an id to the tree so that the logger gets the pluginName where the tree is used rather than just "shared". That was we can separate, for example, 'jmx-tree-logger' from 'camel-tree-logger'.

src/hawtio/plugins/shared/workspace.ts
phantomjinx: Not sure that this necessarily works for the workspace anymore since it is outside the plugins. As a singleton, it is always available as not tied to one plugin. Would `hawtio-plugin-workspace` or `hawtio-shared-workspace` work?

tadayosi: Originally it was `hawtio-jmx-workspace` and `hawtio-jmx-workspace-tree`. You can easily see the original set of logger names by launching the legacy hawtio SB example and then looking at the console logs preferences and hitting the `add` button. For this case, `hawtio-shared-workspace` looks good.

phantomjinx:
Hopefully, all comments fixed, commits rebased and good to go.

tadayosi (approved):
thanks!

PR #63, opened 2023-01-31
First stages of camel plugin

phantomjinx:


packages/hawtio/src/plugins/camel/context.ts
```diff
@@ -0,0 +1,78 @@
+import { createContext, useEffect, useState } from 'react'
+import { TreeViewDataItem } from '@patternfly/react-core'
+import { workspace, MBeanNode, MBeanTree } from '@hawtio/plugins/shared'
+import { pluginName } from './globals'
+
+export const jmxCamelDomain = 'org.apache.camel'
```
tadayosi, line 6: You can refer to it from `globals.ts`

tadayosi (approved):
One minor comment but overall it looks great!

packages/hawtio/src/plugins/camel/TreeProcessor.ts
```diff
@@ -0,0 +1,105 @@
+import React from 'react'
```
tadayosi, line 1: Code styling: I would reserve a capitalised file name for React components. We could change it to `treeProcessor.ts` or `tree-processor.ts`.

tadayosi (approved):
thx!

PR #70, opened 2023-02-01
chore: mandate the minimum version of yarn to use for development

phantomjinx:
* Default installed versions of yarn (1.22-19) do not understand to use the local package directory. They prefer to install published package from npm. This causes issues if the development is futher forward than the published @hawtio/react package.

* package.json
 * Mandate a minimum version of yarn 3.3.1. If lower than this is installed then yarn will display an error message.

* README.md
 * Provides information in the event that the installed version of yarn does not meet the minimum.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,22 @@ You need to have installed the following tools before developing the project.
 - [Node.js >= 16](https://nodejs.org/en/)
 - [Yarn v3](https://yarnpkg.com/getting-started/install)
 
+#### Minimum Version of Yarn is 3.3.1
+The default installation version of yarn on many operating systems is *1.22-19* (the classic version). This causes a problem as the development app downloads the `@hawtio/react` package rather than using
+the project directory. As a result, the mandated minimum version has been set to *3.3.1*.
+
+If `yarn install` is attempted with a version lower than *3.3.1* then an error message is displayed, eg.
+> $ /usr/bin/yarn install
+> yarn install v1.22.19
+> [1/5] Validating package.json...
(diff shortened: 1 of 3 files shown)


tadayosi (commented):
Honestly I don't see why this is needed. In README, requirement has been Yarn v3 and I never expect Yarn v1 (classic) to work well with the project.
https://github.com/hawtio/hawtio-next#prerequisites

And if you are using Node >=16.10, corepack is enabled by default so I thought it picks up a correct yarn version automatically.

Actually, recommendation from Yarn v3 seems to be that the version of yarn used in the project should also be committed under `.yarn/*`.
https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored

So `.gitignore` should be:
```
.pnp.*
.yarn/*
!.yarn/patches
!.yarn/plugins
!.yarn/releases
!.yarn/sdks
!.yarn/versions
```
but I intentionally omitted it because I was not sure if it's really a good practice. If there is a problem with the team using the same version of yarn then we can try it and push an exact version of yarn v3 to the repo.

tadayosi:
Went ahead and created a pull req #71. Does it solve the issue?

phantomjinx:
> Went ahead and created a pull req #71. Does it solve the issue?

I think that #71 is a good alternative, as documented here [1]. However, I did this PR to avoid committing binaries (though .js is technically not a binary ?)  to the repository. Anyway, happy to do either.

That's said, I think the README.md still needs updating since the default install on Fedora, Ubuntu .... is `yarn 1.22-19` so users simply install yarn and not really notice the version they have. Therefore, something along the lines of "yarn required is v3 and note if you have less than this installed then it will default to using yarn 3.4.1 that comes in the .yarn project directory". WDYT?

PR #71, opened 2023-02-02
chore(build): use fixed version of yarn

tadayosi:
Use fixed versoin of Yarn v3 in the project. See #70


lhein:
this is doing a similar thing as in #70 , right?

tadayosi:
@lhein This is what I think should be done instead of #70.

phantomjinx:
See comment here -> https://github.com/hawtio/hawtio-next/pull/70#issuecomment-1413355143

phantomjinx (approved):

PR #77, opened 2023-02-03
fix(connect): deactivate Connect plugin only when connection name is provided

### LINKED ISSUE #75: Connect - Deactivate the plugin only when connection name is provided

tadayosi:
Fix #75

### DIFF
--- a/packages/hawtio/src/Hawtio.test.tsx
+++ b/packages/hawtio/src/Hawtio.test.tsx
@@ -4,7 +4,7 @@ import { Hawtio } from './Hawtio'
 describe('Hawtio', () => {
   test('renders page', async () => {
     render(
-      <Hawtio basepath='/hawtio' />
+      <Hawtio basepath='/' />
     )
     await waitFor(() => {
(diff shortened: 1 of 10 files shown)


oscerd (approved):

phantomjinx (approved):
Happy to approve it as I think I understand the reasoning behind the changes but would appreciate maybe a fuller description as some changes here appear to be unconnected but are actually associated?

tadayosi:
@phantomjinx Sorry, there are a lot of small issues to file to move things forward and I'm tempted to get things done rather than spending my time to give a full description for each issue. I'll try my best to give fuller descriptions for future issues so that everyone can understand them better.

For now, I think it'd be still really helpful if reviewers could check the pull req and see if it doesn't do something really wrong code-wise, regarding for example code structure and how jest tests are written, etc.

About this issue, it's fixing the issue found in the integration with hawtio/hawtio project, where you can see the Connect plugin doesn't appear while it should when you run hawtio-app.jar like this:
```
java -jar hawtio-app.jar
```

PR #93, opened 2023-02-08
Camel plugin: introduction of route-service and schema-service

phantomjinx:
Provides the ability to shred the Camel DSL and add additional child nodes to the camel profile routes in the nav trees


phantomjinx:
Converted to draft atm due to look into alternative XML parser

packages/hawtio/src/plugins/camel/icons/create-icon-index.sh
```diff
@@ -0,0 +1,74 @@
+#!/bin/bash
```
tadayosi, line 1: if we go this direction, at least I'd like to have these scripts in the top level of the `@hawtio/react` package, not under `src/`, and make them runnable from yarn scripts.

and considering we'll need to maintain the scripts it might be better to have them in javascript/typescript instead of shell because we can then utilise the toolset such as linter and formatter for them as well.

packages/hawtio/src/plugins/camel/icons/components/AggregateIcon.tsx
```diff
@@ -0,0 +1,13 @@
+// === THIS FILE WAS AUTO-GENERATED BY create-route-icon.sh ===
```
tadayosi, line 1: Looking at this again, I wonder why is it necessary to generate a separate React component file for each svg icon?  If you look at the contents of the `XxxIcon` component, they are all the same, just the `svg` and `size` parameters are different. If so, why not just create generic `CamelIcon` components and receive the `svg` and `size` parameters in `CamelIconProps`?

Note in the `camel/icons/index.ts` you can also pass a React element instead of component like this:
```
componentMap.set('AggregateIcon', <CamelIcon svg={aggregateIcon} size={16} />)
```

There are already many svg files and if we generate a React component for each svg, the number of files we'd need to maintain is doubled. Plus we'd need to maintain the scripts to generate those files.

Please see how we handle multiple image files in a different place:
https://github.com/hawtio/hawtio-next/blob/main/packages/hawtio/src/ui/page/HawtioBackground.tsx

tadayosi (changes requested):

.eslintignore
```diff
@@ -4,6 +4,8 @@
 /app/*.cjs
 /**/ignore/**
 
+# Ignore generated camel model (need to fix generator to fix all the linting errors)
+/packages/hawtio/src/plugins/camel/model/camelModel.ts
```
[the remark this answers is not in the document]
phantomjinx: This is `eslintignore`. My intent is to ignore it when linting since the model is generated by
the mvn-generator. We can fix the generator but that's a little down the road at the moment. The current model flags up about ~150 errors in the linting.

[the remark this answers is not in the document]
phantomjinx: Added https://github.com/hawtio/hawtio-next/issues/95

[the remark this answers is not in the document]
tadayosi: You are right. My bad, I thought it was .gitignore but it is .eslintignore ... Please ignore this comment. We won't even need to fix lint issues for the generated file.

packages/hawtio/src/plugins/camel/icons/create-icon-index.sh
phantomjinx: No removed as a result of refactoring the icons module

packages/hawtio/src/plugins/camel/icons/components/AggregateIcon.tsx
phantomjinx: Refactored, as you suggested, with IconNames enum and generic CamelImageIcon component. Single generator javascript located in top-level package utils directory (added to linting).

packages/hawtio/src/plugins/camel/CamelTreeView.tsx
```diff
@@ -10,19 +10,28 @@ import {
 } from '@patternfly/react-core'
 import { CamelContext } from './context'
 
+//
```
tadayosi, line 13: we can use jsdoc format for the comments in the enum.

packages/hawtio/src/plugins/camel/icons/components/CamelImageIcon.tsx
```diff
@@ -0,0 +1,23 @@
+
```
tadayosi, line 1: I don't see value in putting this component under `components/`.

Generally, in this project I try to avoid structuring packages based on functionality/kind but rather based on feature/domain:
https://medium.com/sahibinden-technology/package-by-layer-vs-package-by-feature-7e89cde2ae3a

So I try to avoid creating directories such as `components`, `services`, `types`, etc under each plugin. Rather we can organise files with directories like `featureA`, `featureB`, ... I think `icons` or `images` are fine as they stand on its own.

tadayosi (commented):
Overall, it looks great. For me, the remaining issue is to make camel-model a separate package under `packages/`.

phantomjinx:
> Overall, it looks great. For me, the remaining issue is to make camel-model a separate package under `packages/`.

Still working it. :-)

packages/hawtio/src/util/objects.ts
```diff
@@ -6,3 +6,37 @@ export function isObject(value: unknown): value is object {
...
+ * and returns the last most value of the path or null if it could not be found.
+ *
+ * @method pathGet
+ * @static
+ * @param {Object} obj the start object to start navigating from
+ * @param {Array} paths an array of path names to navigate or a string of dot separated paths to navigate
+ * @return {*} the last step on the path which is updated
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function pathGet(obj: any, paths: string[]): object {
```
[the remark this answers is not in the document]
phantomjinx: Removed.

packages/hawtio/src/util/objects.ts
```diff
@@ -6,3 +6,37 @@ export function isObject(value: unknown): value is object {
 export function isEmpty(obj: object): boolean {
   return Object.keys(obj).length === 0
 }
+
+/**
```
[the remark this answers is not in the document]
phantomjinx: Think this still works as a utility function. At least according to
https://www.freecodecamp.org/news/check-if-an-object-is-empty-in-javascript/

phantomjinx:
> > Overall, it looks great. For me, the remaining issue is to make camel-model a separate package under `packages/`.
> 
> Still working it. :-)

Done.

packages/camel-model/tsconfig.json
```diff
@@ -0,0 +1,23 @@
+{
+  "compilerOptions": {
```
tadayosi, line 2: we might want to revisit these options, especially on "jsx" where we shouldn't need react & jsx support here.

packages/hawtio/package.json
```diff
@@ -24,18 +24,21 @@
...
     "build:watch": "tsup --watch",
+    "generate:camel-icons": "node ./utils/generate-camel-svg-index.js",
     "test": "jest",
     "test:watch": "jest --watch",
     "test:coverage": "jest --coverage",
     "replace-version": "replace __PACKAGE_VERSION_PLACEHOLDER__ $npm_package_version ./dist/index.js",
     "prepack": "yarn build && yarn replace-version"
   },
   "dependencies": {
+    "@hawtio/camel-model-react": "^0.1.0-dev",
```
tadayosi, line 35: you can specify the version as `"workspace:*"`. it should be resolved at build time by yarn workspace.

packages/hawtio/package.json
```diff
@@ -24,18 +24,21 @@
...
     "@patternfly/react-core": "^4.267.6",
     "@patternfly/react-table": "^4.112.6",
     "@testing-library/jest-dom": "^5.16.5",
     "@testing-library/react": "^13.4.0",
     "@testing-library/user-event": "^14.4.3",
+    "@types/clone": "^2.1.1",
```
tadayosi, line 41: do we still need this?

packages/hawtio/jest.config.ts
```diff
@@ -10,7 +10,7 @@ export default {
     '\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga|md)$':
       '<rootDir>/src/__mocks__/fileMock.js',
     '\\.(css|less)$': '<rootDir>/src/__mocks__/styleMock.js',
-    '@hawtio/(.*)': '<rootDir>/src/$1',
+    '@hawtiosrc/(.*)': '<rootDir>/src/$1',
```
tadayosi, line 13: hmm, we might need to revisit the usage of this custom path alias... for the time being it's ok, but we should reconsider if we really need this or not and if we do what's the proper name for the alias.

packages/hawtio/src/plugins/camel/CamelTreeView.tsx
```diff
@@ -1,21 +1,30 @@
 import React, { ChangeEvent, useState, useContext } from 'react'
-import { PluginTreeViewToolbar, MBeanNode, MBeanTree } from '@hawtio/plugins/shared'
+import { PluginTreeViewToolbar, MBeanNode, MBeanTree } from '@hawtiosrc/plugins/shared'
 import { TreeView, TreeViewDataItem } from '@patternfly/react-core'
 import { CamelContext } from './context'
 
+/*
```
tadayosi, line 6: this should be jsdoc, starting from `/**`

packages/hawtio/src/plugins/camel/CamelTreeView.tsx
```diff
@@ -1,21 +1,30 @@
...
+ * Expansion requires more than 2 states since the expandAll
+ * must be removed completely to defer to the expanded state
+ * of each data node
+ */
+enum ExpansionValue {
+  Default = 0, // should revert to the expanded state of the data
```
tadayosi, line 12: these line comments can be jsdoc as well. and I think the value assignment is superfluous for enums.

packages/hawtio/src/plugins/camel/icons/Icons.test.tsx
```diff
@@ -0,0 +1,22 @@
...
+    render(icon1)
+    render(icon2)
+    render(icon3)
+    render(icon4)
+
+    expect(screen.getByAltText(IconNames.AggregateIcon)).toBeInTheDocument()
+    expect(screen.getByAltText(IconNames.GenericIcon)).toBeInTheDocument()
+    expect(screen.getByAltText(IconNames.RouteIcon)).toBeInTheDocument()
+    expect(screen.getByAltText(IconNames.EndpointsNodeIcon)).toBeInTheDocument()
+    cleanup()
```
tadayosi, line 20: is this cleanup necessary?  according to https://testing-library.com/docs/react-testing-library/api/#cleanup
> Please note that this is done automatically if the testing framework you're using supports the afterEach global and it is injected to your testing environment (like mocha, Jest, and Jasmine). If not, you will need to do manual cleanups after each test.

packages/hawtio/src/plugins/camel/routes-service.test.ts
```diff
@@ -0,0 +1,114 @@
...
+  let simpleRouteNode: MBeanNode
+
+  const testRouteId = 'simple'
+  const routesXmlPath = path.resolve(__dirname, 'testdata', 'camel-sample-app-routes.xml')
+  const sampleRoutesXml = fs.readFileSync(routesXmlPath, { encoding: 'utf8', flag: 'r' })
+  const doc: XMLDocument = $.parseXML(sampleRoutesXml as string)
+  const simpleRouteXml = $(doc).find("route[id='" + testRouteId + "']")[0]
+
+  jolokiaService.execute = jest.fn(
+    async (mbean: string, operation: string, args?: unknown[] | undefined): Promise<unknown> => {
```
tadayosi, line 26: in `args?: unknown[] | undefined`, no need for `| undefined` since `args` is assigned with `?`.

packages/hawtio/src/plugins/camel/routes-service.ts
```diff
@@ -0,0 +1,145 @@
+import React from 'react'
+import { Logger } from '@hawtiosrc/core'
+import { MBeanNode } from '@hawtiosrc/plugins/shared'
+import { jolokiaService } from '@hawtiosrc/plugins/connect/jolokia-service'
+import { jmxDomain } from './globals'
+import $ from 'jquery'
+import { schemaService } from './schema-service'
+import * as icons from './icons'
+
+const log = Logger.get(`hawtio-${jmxDomain}-routes-service`)
```
tadayosi, line 10: I think it's too much for logging. Here we can just use the global logger named 'hawtio-camel' from `./globals`. For starters, I'd prefer to keep the same level of logger granularity as the legacy hawtio-integration does.

packages/hawtio/src/util/objects.ts
```diff
@@ -3,6 +3,14 @@ export function isObject(value: unknown): value is object {
...
 
+export function cloneObject<T>(value: T): T {
+  return JSON.parse(JSON.stringify(value))
+}
+
 export function isEmpty(obj: object): boolean {
   return Object.keys(obj).length === 0
 }
+
+export function isString(value: unknown) {
```
tadayosi, line 14: let's declare return type as `value is string` to make our type-safe programming easier with it.

packages/hawtio/src/util/objects.ts
```diff
@@ -3,6 +3,14 @@ export function isObject(value: unknown): value is object {
   return value != null && (type === 'object' || type === 'function')
 }
 
+export function cloneObject<T>(value: T): T {
+  return JSON.parse(JSON.stringify(value))
```
tadayosi, line 7: at some point in the future we should replace this with `structuredClone()` builtin fn.

packages/hawtio/src/plugins/connect/__mocks__/jolokia-service.ts
```diff
@@ -23,7 +23,7 @@ class MockJolokiaService implements IJolokiaService {
     return null
   }
 
-  async execute(mbean: string, operation: string, args?: unknown[]): Promise<unknown> {
+  async execute(mbean: string, operation: string, args?: unknown[] | undefined): Promise<unknown> {
```
tadayosi, line 26: why do we need this?  does it show type error in your editor?

tadayosi (changes requested):
There are a few comments, but the most important part is to update to the latest Camel version. and then remove `@types/clone`.

packages/hawtio/jest.config.ts
phantomjinx: Agree. It ended up bringing more trouble with the introduction of an additional package. Created https://github.com/hawtio/hawtio-next/issues/105.

packages/hawtio/src/util/objects.ts
phantomjinx: Captured with issue https://github.com/hawtio/hawtio-next/issues/106

tadayosi (approved):
Looks all good to me, thanks for the great work!  Let's merge it.

PR #96, opened 2023-02-10
chore: use prettier for consistent code formatting #92

### LINKED ISSUE #92: Use prettier dependency for consistent formatting and styling across IDEs

tadayosi:
fix #92

From this commit, we use Prettier to format most of the project code in order to keep the same formatting rules across different IDEs such as VS Code and WebStorm the team use.

- You can run format check with `yarn format:check` and fix with `yarn format:fix`
- Format check is also integrated into the `Lint` CI workflow
- Please follow https://prettier.io/docs/en/editors.html to integrate Prettier settings with your editor. I recommend enabling auto formatting upon save files.

The Prettier rules I apply as the default:
```
module.exports = {
  printWidth: 120,
  semi: false,
  singleQuote: true,
  jsxSingleQuote: true,
  trailingComma: 'all',
  bracketSpacing: true,
  bracketSameLine: false,
  arrowParens: 'avoid',
}
```

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -8,25 +8,23 @@ module.exports = {
     'plugin:react/jsx-runtime',
     'plugin:react-hooks/recommended',
     'plugin:testing-library/react',
+    'prettier',
   ],
   parser: '@typescript-eslint/parser',
-  plugins: [
-    '@typescript-eslint',
-    'import',
-  ],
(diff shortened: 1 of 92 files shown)


mmelko (approved):
Looks good! Formatting seems to be consistent now in webstorm too. 
LGTM

PR #115, opened 2023-02-18
Camel Plugin Node Pages

phantomjinx:


phantomjinx:
No where near finished but want to share progress on this to give others chance to comment on how I am doing this.
Highlights so far:
- Moving JMXContentBeans to shared components for use in camel and jmx plugins
- Moving Attributes, Operations, Chart to shared components for use in camel and jmx plugins
  - Necessitates removal of useContext and passing in of props containing selected node

tadayosi:
>   * Necessitates removal of useContext and passing in of props containing selected node

I'd avoid using props instead of useContext, as it would require deep chaining of props passing. That should be exactly the reason useContext is introduced to React. Rather, why can't you just move the context in JMX plugin into the shared and reuse from different plugins?  Note multiple contexts can be used in a React component.

I think what's important is to design the scope of a context -- what information should be conveyed down to which React components?  If we think there should be some context that are shared across the entire console instead of just a subset of plugins, we can create one somewhere either under `/plugins/` or a dedicated directory under root src.

tadayosi:
Another possible approach would be to just make Camel plugin depend on JMX plugin. I think it's ok for plugins to have inter-dependencies; I regard `connect` and `jmx` as *core* plugins that should be essentially shared across other downstream plugins such as Camel, Artemis, etc. I'm not sure how much we could gain by moving every sharable component into `shared`. I think it's ok to put some reusable React components into `shared`, just like we put some common code into `utils`.

phantomjinx:
> Another possible approach would be to just make Camel plugin depend on JMX plugin. I think it's ok for plugins to have inter-dependencies; I regard `connect` and `jmx` as _core_ plugins that should be essentially shared across other downstream plugins such as Camel, Artemis, etc. I'm not sure how much we could gain by moving every sharable component into `shared`. I think it's ok to put some reusable React components into `shared`, just like we put some common code into `utils`.

Think I will explore `useContext` approach first as I feel the concept of core plugins has historically been soemthing of a design anti-pattern (to me). In that if something is a plugin then it can be removed and everything else in the app still works. If it is core then it is fundamental and necessary so removing it will break stuff. This may not be sustainable and eventually we have to consolidate with `core` plugins but if you wouldn't mind I'd like to try and continue with this mindset for the time being.

tadayosi:
> Think I will explore `useContext` approach first as I feel the concept of core plugins has historically been something of a design anti-pattern (to me). In that if something is a plugin then it can be removed and everything else in the app still works. If it is core then it is fundamental and necessary so removing it will break stuff.

I can see the point of core plugins being anti-pattern. But in fact Hawtio.next indeed works without those core plugins; it just provides blank workspace with help, preferences, about, and authentication features. So they are still *plugins*. I think it's somewhat similar to the architecture of Eclipse; everything is implemented as inter-dependent plugins.

Maybe the next important question is whether we should allow a plugin to depend on other plugins. Ideally it's better to make them independent of each other, but that would make implementing a plugin much harder in exchange for a cleaner plugin design.

> This may not be sustainable and eventually we have to consolidate with `core` plugins but if you wouldn't mind I'd like to try and continue with this mindset for the time being.

Actually, the main reason why I've chosen the architecture of everything being a plugin is because the legacy Hawtio has been historically based on the same architecture; everything starting from JVM, JMX, and Jolokia-related features has been implemented as a plugin. I think that the important goal of the Hawtio.next project is quick win. We should migrate Hawtio to React + PF4 as fast as possible, so following the existing architecture and code design seems to be the quickest way to achieve that. We should be able to have a plenty of time to improve the code design once we have a solid basis of a working project, but right now we should focus on completing features.

tadayosi:
Another angle of discussion about whether JMX-related plugins should be really *core* is that JMX is really a technology that's getting obsolete in the cloud environments. In the future, we might want to pivot Hawtio to support other management protocols than JMX, such as GraphQL, K8s API, etc. If we make it *core* then Hawtio would be permanently bound to JMX and would lose the opportunity to adopt other new technologies unless we start over Hawtio.next.next again.

phantomjinx:
> > Think I will explore `useContext` approach first as I feel the concept of core plugins has historically been something of a design anti-pattern (to me). In that if something is a plugin then it can be removed and everything else in the app still works. If it is core then it is fundamental and necessary so removing it will break stuff.
> 
> I can see the point of core plugins being anti-pattern. But in fact Hawtio.next indeed works without those core plugins; it just provides blank workspace with help, preferences, about, and authentication features. So they are still _plugins_. I think it's somewhat similar to the architecture of Eclipse; everything is implemented as inter-dependent plugins.
> 
> Maybe the next important question is whether we should allow a plugin to depend on other plugins. Ideally it's better to make them independent of each other, but that would make implementing a plugin much harder in exchange for a cleaner plugin design.
> 
> > This may not be sustainable and eventually we have to consolidate with `core` plugins but if you wouldn't mind I'd like to try and continue with this mindset for the time being.
> 
> Actually, the main reason why I've chosen the architecture of everything being a plugin is because the legacy Hawtio has been historically based on the same architecture; everything starting from JVM, JMX, and Jolokia-related features has been implemented as a plugin. I think that the important goal of the Hawtio.next project is quick win. We should migrate Hawtio to React + PF4 as fast as possible, so following the existing architecture and code design seems to be the quickest way to achieve that. We should be able to have a plenty of time to improve the code design once we have a solid basis of a working project, but right now we should focus on completing features.

Sorry, to clarify. Am happy with the plug-in architecture but just prefer to keep the inter-dependencies to a minimum.

Have moved back to 'useContext' in the shared components by creating a shared plugin context provider. This has had the benefit that the selection in the JMX plugin is now maintained, where applicable, in the Camel plugin and vice-versa. Moving on to porting other pages.

packages/hawtio/src/plugins/camel/contexts/context-actions.component.ts
```diff
@@ -0,0 +1,104 @@
+/// <reference path="contexts.service.ts"/>
```
tadayosi, line 1: Maybe it's a leftover from the legacy code?

packages/hawtio/src/plugins/camel/contexts/contexts.component.ts
```diff
@@ -0,0 +1,146 @@
+/// <reference path="context.ts"/>
```
tadayosi, line 1: leftover

packages/hawtio/src/plugins/camel/contexts/contexts-service.ts
```diff
@@ -0,0 +1,99 @@
...
+import { IRequest, IResponseFn } from 'jolokia.js'
+import { jolokiaService, AttributeValues } from '@hawtiosrc/plugins/connect/jolokia-service'
+import { MBeanNode } from '@hawtiosrc/plugins/shared'
+
+export interface ContextAttributes {
+  Context: string,
+  [key: string]: any
+}
+
+class ContextsService {
```
tadayosi, line 11: By the way, starting from Camel 3 we no longer allow multiple contexts in a Camel app. The plural treatment of Camel contexts is legacy handling of the code, so we should revisit it and try to remove the presumption that the app can have multiple contexts.

packages/hawtio/src/plugins/camel/contexts/Contexts.tsx
```diff
@@ -0,0 +1,138 @@
...
+import { IResponse } from 'jolokia.js'
+import { AttributeValues } from '@hawtiosrc/plugins/connect/jolokia-service'
+import { useEffect, useState, useContext } from 'react'
+import { PluginNodeSelectionContext } from '@hawtiosrc/plugins'
+import { log } from '../globals'
+import { contextsService, ContextAttributes } from './contexts-service'
+import { moveElement } from '@hawtiosrc/util/arrays'
+import { isObject } from '@hawtiosrc/util/objects'
+
+export const Contexts: React.FunctionComponent = () => {
```
tadayosi, line 13: Same here. We should revisit if we can rename it to singular "Context" and drop presumption that we may have multiple contexts if it makes sense.

packages/hawtio/src/plugins/jmx/contexts/Contexts.tsx
```diff
@@ -0,0 +1,135 @@
...
+import { Table, TableBody, TableHeader, TableProps, wrappable } from '@patternfly/react-table'
+import { IResponse } from 'jolokia.js'
+import { AttributeValues } from '@hawtiosrc/plugins/connect/jolokia-service'
+import { useEffect, useState, useContext } from 'react'
+import { PluginNodeSelectionContext } from '@hawtiosrc/plugins'
+import { log } from '../globals'
+import { contextsService, ContextAttributes } from './contexts-service'
+import { moveElement } from '@hawtiosrc/util/arrays'
+
+export const Contexts: React.FunctionComponent = () => {
```
tadayosi, line 12: What is the "Context" in JMX plugin?  I'm a bit confused as there's no such concept of "context" in JMX.

packages/hawtio/src/plugins/shared/components/JmxContentBeans.tsx
```diff
@@ -0,0 +1,50 @@
+import { useContext } from 'react'
```
tadayosi, line 1: The file name should be `JmxContentMBean` instead of `JmxContentBean` (note 'M').

packages/hawtio/src/plugins/shared/components/attributes/index.ts
```diff
@@ -0,0 +1,2 @@
+export { Attributes } from './Attributes'
```
tadayosi, line 1: If we move `attributes` to `shared`, I wouldn't like to move it to `shared/components` for the same reason as features over functions. Can we move it to `shared/attributes`?

packages/hawtio/src/ui/page/HawtioPage.tsx
```diff
@@ -18,10 +18,12 @@ import { HawtioBackground } from './HawtioBackground'
 import { HawtioHeader } from './HawtioHeader'
 import { HawtioLoading } from './HawtioLoading'
 import { HawtioSidebar } from './HawtioSidebar'
+import { PluginNodeSelectionContext, usePluginNodeSelected } from '@hawtiosrc/plugins'
```
tadayosi, line 21: hmm, with this dependency, now we have a circular dependency between the core ui part and the plugins. Is it what we really need?

packages/hawtio/src/util/arrays.ts
```diff
@@ -0,0 +1,8 @@
+export function moveElement<T>(target: T[], value: T, dest: number) {
```
tadayosi, line 1: no tests?

packages/hawtio/src/plugins/camel/contexts/contexts-service.ts
lhein: What does that mean in terms of backwards compatibility? Does it mean we only support Camel 3+ with the new hawtio?

tadayosi: Good point. I understand that backward compatibility is really important from the product point of view, but at the same time I'm not sure if it's good idea to maintain the old design for a new project. Maybe we should keep the design of accepting multiple Camel contexts for the time being?

By the way, we already started to be based on Camel 3.20.2 model for the Camel plugin. To really achieve backward compatibility, we'd need to publish an old Camel 2.x model and cut a branch that consumes the 2.x model for the product.

lhein: I will try to figure out if we need the backwards compatibility.

packages/hawtio/src/plugins/camel/contexts/context-actions.component.ts
phantomjinx: Shouldn't be there at all. Have committed an old angular file. Also spotted a WIP commit. Will scrub and rebase.

packages/hawtio/src/ui/page/HawtioPage.tsx
phantomjinx: Not quite. The `PluginNodeSelectionContext is not in the JMX plugin but in the `plugin` module. I thought hard about this. If we want to move it to a core module then that should be ok I think.

packages/hawtio/src/plugins/jmx/contexts/Contexts.tsx
phantomjinx: When the `contexts` node is clicked on within the JMX tree ....

jmx-contexts-screenshot-old

Old version ^

jmx-contexts-screenshot-new

New version ^

lhein: Its about Camel Contexts and nothing to do with JMX really other than that we grab information via JMX :)

packages/hawtio/src/ui/page/HawtioPage.tsx
phantomjinx: https://github.com/hawtio/hawtio-next/pull/115/files#diff-4b7e788f0291a4786b206a46509a458911884d644d2bc9e245bda4d449b0344f

packages/hawtio/src/plugins/jmx/contexts/Contexts.tsx
phantomjinx: So, I started creating it originally as I mistook it for the `expected` contexts page in the camel plugin. Only when I had completed it did I realise that the camel page was indeed different. Therefore, rather than waste it I moved it to the JMX plugin.

I believe the original Attributes components provided this functionality somehow, ie. the rotated table with the attributes as columns, but the react version certainly doesn't. Therefore, for the moment it may be worth keeping just for providing that alternative `view` for any other JMX node that perhaps requires it?

tadayosi: For the JMX plugin, I don't think we need to provide a specialised view for some specific MBeans such as Camel contexts. I don't think it was by design in the old Hawtio version, probably it was just a collateral side effect from the specific way of implementation. So long as we can provide a specific view at the dedicated plugin (i.e. Camel plugin for Camel Contexts) it should be fine. Rather it should be better to keep JMX plugin as generic and not to bring any assumptions about specific application context.

packages/hawtio/src/plugins/camel/context.ts
```diff
@@ -39,39 +27,58 @@ export function useCamelTree() {
...
       } else {
-        console.log('TODO: reroute back to jmx view')
         setTree(wkspTree)
-        // // No camel contexts so redirect to the JMX view and select the first tree node
-        // if (tree.children && tree.children.length > 0) {
-        //   const firstNode = tree.children[0]
-        //   this.$location.path('/jmx/attributes').search({ 'nid': firstNode['id'] })
-        // }
+        // No camel contexts so redirect to the JMX view and select the first tree node
+        navigate('/jmx')
```
tadayosi, line 41: no need for `/` with react router v6.

packages/hawtio/src/plugins/camel/contextsdash/ContextsDash.tsx
```diff
@@ -0,0 +1,345 @@
...
+        return id && ctx.state === 'Suspended'
+      })
+      .forEach(ctx => {
+        contextsService
+          .startContext(ctx)
+          .then(() => {
+            eventService.notify({
+              type: 'success',
+              message: 'Camel context started successfully',
+              duration: 3000,
```
tadayosi, line 73: let's follow the same duration settings with the legacy version, unless there's specific reason not to do so. For these parts, I don't think we need to set special durations. Default should just work.

packages/hawtio/src/plugins/camel/contextsdash/ContextsDash.tsx
```diff
@@ -0,0 +1,345 @@
...
+import { IResponse } from 'jolokia.js'
+import { AttributeValues } from '@hawtiosrc/plugins/connect/jolokia-service'
+import React, { useEffect, useState, useContext } from 'react'
+import { PluginNodeSelectionContext } from '@hawtiosrc/plugins'
+import { log } from '../globals'
+import { contextsService, ContextDashAttributes } from './contexts-service'
+import { eventService } from '@hawtiosrc/core'
+import { workspace } from '@hawtiosrc/plugins/shared'
+
+export const ContextsDash: React.FunctionComponent = () => {
```
tadayosi, line 26: what is `ContextsDash`?

packages/hawtio/src/ui/page/HawtioPage.tsx
tadayosi: But the concept of `selectedNode` comes from JMX plugin's MBean tree, so that's why I think the page depends on JMX plugin.

packages/hawtio/src/plugins/jmx/contexts/Contexts.tsx
tadayosi: So, indeed the old Hawtio has a certain feature that shows a table view of MBean attributes on the folder with homogeneous  MBeans. We need to implement the same feature in Hawtio.next. But we don't call it "contexts" but just "attributes" if I'm not mistaken.

image

phantomjinx: Right. I'll remove the JMX `Contexts` component (in the git history then in case we need similar functionality) and create an issue for the `homogenous attributes` page view. If you could point out the `old` functionality I'd be grateful as I dug around for ages and failed to find it.

packages/hawtio/src/ui/page/HawtioPage.tsx
phantomjinx: The MBean tree is no longer in the JMX plugin. It is fetched and provided by [workspace](https://github.com/hawtio/hawtio-next/blob/main/packages/hawtio/src/plugins/shared/workspace.ts#L14), which is in shared.

packages/hawtio/src/plugins/camel/contextsdash/ContextsDash.tsx
phantomjinx: I'll remove JMX plugin version and rename.

packages/hawtio/src/plugins/jmx/contexts/Contexts.tsx
phantomjinx: https://github.com/hawtio/hawtio-next/issues/125

packages/hawtio/src/plugins/camel/contexts/Contexts.tsx
phantomjinx: Issue posted to track: https://github.com/hawtio/hawtio-next/issues/126

packages/hawtio/src/plugins/camel/contexts/contexts-service.ts
phantomjinx: Tracking issue posted -> https://github.com/hawtio/hawtio-next/issues/126

packages/hawtio/src/ui/page/HawtioPage.tsx
tadayosi: Shared is still part of the plugins. I don't think it improved the situation. For now, let's accept the circular dependency and go ahead.

tadayosi (approved):
There are small things that can be improved, but we can do so later. Let's merge this for now.

tadayosi:
@phantomjinx If there are any issues that can be closed by this pull req, please close.

Issue #72, opened 2023-02-02
UX - Tree selections in the JMX tree are not highlighted


lhein:
When selecting an item from the JMX tree that item is not highlighted as selected so you end up not seeing what you selected in the tree. The Camel Plugin tree has a solution to this already.

joshiraez:
I'll be doing this one together with the search on JMX tree as it should be quite easy and I can do it together with the functionality i'm working on

phantomjinx:
So I spent some time going down a rabbit hole ...

I noticed that clicking on a folder in the JMX TreeView displays a generic Attribute Page and, if a folder, lists the contents of it. One can then click on each row and effectively move the selection in the JMX TreeView to that node. I experimented with updating the tree to highlight selections in blue, in the same way as the CamelTreeView (using the prop `hasSelectableNodes={true}`). However, there is a problem ....

If the parent folder is collapsed, the external selection from the Attribute Page fires, the node is selected but the parent folder is NOT automatically expanded. I spent a good few hours trying to get it to expand using the property `defaultExpanded` but nothing seemed to work. So just to be clear ...

1. In JMXTreeView, click on `jolokia` to display its Attribute Page;
2. On the Attribute Page, click on `Config`, `Discovery` or `ServerHandler`. The Attribute Page will refresh to the contents of that node and selection in the JMXTreeView will reflect the same node;
3. Repeat step 1 then collapse `jolokia` in the JMXTreeView then repeat step 2. Despite the selection working correctly, `jolokia` does NOT re-expand to display the selection.

Looking back at the angular version of the app, the expansion did occur.

If you want to experiment, I can post a branch with my experiments on it. Otherwise, it may be a separate issue that we have to take up with Patternfly.

lhein:
Lets keep it simple for the moment. We can look at that once we are feature complete and start to polish the UI.

tadayosi:
Yes, UX improvements can be addressed later after we complete the feature set. I moved this ticket to 0.x and 'no status' from 'to do'.

PR #119, opened 2023-02-27
Search on JMX tree (Fixes #24 and #72)

### LINKED ISSUE #24: JMX - Search on JMX tree
This feature is not yet implemented.

image

joshiraez:
Search on JMX tree and highlight of selected node.

Tree search functionality is as follows:

- If parent node is a hit bring the parent node completely.
- If parent node is not a hit but any subnodes are hits, return the parent with only the subnodes that are hits. (Subnodes follow the same rule as above).
- If there are no hits in parent or subnodes, it's not shown.

jmxsearch

@phantomjinx  I saw that you did one search functionality and put it in the class JMBeanTree. Would you like me to update it? Right now yours wont retrieve children nodes on parent hits.

|-> Will be setting up a PR in the future refactoring this code to the general case for both JMX and Camel

Also another gif showing the highlight in action 

highlighting

Fixed the issue with the broken yarn build with @tadayosi [suggestion](https://github.com/hawtio/hawtio-next/pull/119#issuecomment-1446097088). Using yarn build:all or yarn build:camel-model does the trick :)

### DIFF
--- a/packages/hawtio/src/plugins/jmx/JmxTreeView.tsx
+++ b/packages/hawtio/src/plugins/jmx/JmxTreeView.tsx
@@ -8,16 +8,15 @@ import './JmxTreeView.css'
 export const JmxTreeView: React.FunctionComponent = () => {
   const { tree, node, setNode } = useContext(MBeanTreeContext)
   const [expanded, setExpanded] = useState(false)
-  const [filteredTree, updateTree] = useState(tree.getTree())
+  const [filteredTree, setFilteredTree] = useState(tree.getTree())
 
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const onSearch = (event: ChangeEvent<HTMLInputElement>) => {
-    if (!event.target.value) updateTree(tree.getTree())
+    if (!event.target.value) setFilteredTree(tree.getTree())
 
     const treeElements = lookupSearchInTree(event.target.value, tree.getTree())
 
-    if (treeElements.length === 0) updateTree(tree.getTree())
-    else updateTree(treeElements)
+    if (treeElements.length === 0) setFilteredTree(tree.getTree())
+    else setFilteredTree(treeElements)
   }
 
   const onSelect = (event: React.MouseEvent<Element, MouseEvent>, item: TreeViewDataItem) => {
@@ -25,11 +24,11 @@ export const JmxTreeView: React.FunctionComponent = () => {
   }
 
   const lookupSearchInTree = (search: string, tree?: MBeanNode[]): MBeanNode[] => {
-    if (tree?.length === 0) return []
+    if (!tree || tree?.length === 0) return []
 
     let results: MBeanNode[] = []
 
-    for (const parentNode of tree || []) {
+    for (const parentNode of tree) {
       if (parentNode.name.toLowerCase().includes(search.toLowerCase())) {
         results = results.concat(parentNode)
       } else {
--- a/packages/hawtio/src/plugins/jmx/JmxTreeView.tsx
+++ b/packages/hawtio/src/plugins/jmx/JmxTreeView.tsx
@@ -1,7 +1,7 @@
 import { TreeView, TreeViewDataItem } from '@patternfly/react-core'
 import React, { ChangeEvent, useContext, useState } from 'react'
-import { PluginTreeViewToolbar } from '@hawtio/plugins/shared'
-import { MBeanNode } from '@hawtio/plugins/shared'
+import { PluginTreeViewToolbar } from '@hawtiosrc/plugins/shared'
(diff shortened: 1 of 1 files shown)


packages/hawtio/src/plugins/jmx/JmxTreeView.tsx
```diff
@@ -6,23 +6,54 @@ import { MBeanTreeContext } from './context'
 import './JmxTreeView.css'
 
 export const JmxTreeView: React.FunctionComponent = () => {
-  const { tree, setNode } = useContext(MBeanTreeContext)
-  const [expanded, setExpanded] = useState(false)
+  const { tree, node, setNode } = useContext(MBeanTreeContext);
+  const [expanded, setExpanded] = useState(false);
+  const [filteredTree, updateTree] = useState(tree.getTree());
```
tadayosi, line 11: Normally, with React `useState` hook a state is accompanied with its `set` function, so instead of naming it `updateTree` it should be `setFilteredTree` according to the convention:
https://beta.reactjs.org/reference/react/useState#setstate

packages/hawtio/src/plugins/jmx/JmxTreeView.tsx
```diff
@@ -6,23 +6,54 @@ import { MBeanTreeContext } from './context'
...
 
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
tadayosi, line 13: please remove it if you complete it.

packages/hawtio/src/plugins/jmx/JmxTreeView.tsx
```diff
@@ -6,23 +6,54 @@ import { MBeanTreeContext } from './context'
...
   const onSelect = (event: React.MouseEvent<Element, MouseEvent>, item: TreeViewDataItem) => {
     setNode(item as MBeanNode)
   }
 
+  const lookupSearchInTree = (search: string, tree?: MBeanNode[]) : MBeanNode[] => {
+    if (tree?.length == 0) return [];
+    
+    let results : MBeanNode[] = [];
+
+    for (let parentNode of tree || []) {
```
tadayosi, line 32: This should be `const`. Also, do we need `|| []` part?  As emptiness of the list should be checked at the beginning of the function.

packages/hawtio/src/plugins/jmx/JmxTreeView.tsx
```diff
@@ -6,23 +6,54 @@ import { MBeanTreeContext } from './context'
...
+    if (treeElements.length == 0) updateTree(tree.getTree())
+    else updateTree(treeElements);
   }
 
   const onSelect = (event: React.MouseEvent<Element, MouseEvent>, item: TreeViewDataItem) => {
     setNode(item as MBeanNode)
   }
 
+  const lookupSearchInTree = (search: string, tree?: MBeanNode[]) : MBeanNode[] => {
+    if (tree?.length == 0) return [];
```
tadayosi, line 28: It should also check `tree`'s existence.

tadayosi (approved):
There are small issues but otherwise it looks great. Thanks!

tadayosi:
@joshiraez Please fix lint errors.

Regarding the build error at main, sorry, I've changed the build workflow a bit. That must be the reason why it didn't work for you. If you clone the project and run `yarn install` and still you cannot run `yarn build` successfully, try running `yarn build:all` first. Now the project has three packages: `packages/hawtio`, `packages/camel-model`, and `app`, and `yarn build` only runs build on `packages/hawtio` as we rarely need to build the others. Also, now `packages/hawtio` doesn't depend on the workspace `packages/camel-model`, so it shouldn't require full build with `yarn build:all` normally.

If you still fail to build, sometimes cleaning up `node_modules` might solve the problem :wink:

packages/hawtio/src/plugins/jmx/JmxTreeView.tsx
joshiraez: If tree is null or undefined, the ?. operator will return undefined instead of following.

It's equivalent to `if(tree && tree.length ==0)`

packages/hawtio/src/plugins/jmx/JmxTreeView.tsx
joshiraez: Yes, if we remove the || [] part TS will throw an error stating that parentNode.children is of type MBeanNode | undefined, even though we are checking if it's null or not at the top.

joshiraez: Actually the above comment is wrong. You were right that I wasn't checking if tree existed.

phantomjinx (approved):
LGTM. Please go ahead and move to the general case under MNodeBean

Issue #28, opened 2023-01-26
Connect - Preferences


tadayosi:
Connect prefenreces is not yet fully implemented. See `// TODO` comments in `ConnectPreferences.tsx`.

What should be implemented:

- Apply the parameters to Jolokia instance
- Reset the connections settings persisted in local storage

![Connect preferences](https://user-images.githubusercontent.com/156692/214837685-660649ff-466f-4083-b021-3a782d0aa4e2.png)

joshiraez:
I'll take this

joshiraez:
I'm starting with this

PR #120, opened 2023-02-27
#55 Home preferences, saving to local storage

### LINKED ISSUE #55: Core - Home preferences
The Home preferences is not fully implemented yet.

What should be implemented:

- Switch to show/hide the vertical navigation by default
- Reset button to clear all Hawtio settings in the local storage

Home

joshiraez:
Fixes #55 

Quick ticket, although I'm not sure if we are to implement the actual behaviour of the preference in this ticket as well.

Not so elegant solution, as I'd like to be able to have a list/dictionary configuring all the fields and their defaults, local storage keys and update functions, but I need to compute the initial value before retrieving the function, and I need the function before I can build the dictionary. So in the end I went with a simpler solution just adding all the data to an array

The array is used by the reset function to make sure everything goes back into place.

Even though this is a rather small preferences page, I'll be using a similar pattern in #28 

Gif showing it in action

Preferences

### DIFF
--- a/packages/hawtio/src/preferences/HomePreferences.tsx
+++ b/packages/hawtio/src/preferences/HomePreferences.tsx
@@ -1,13 +1,49 @@
 import { Button, CardBody, Form, FormGroup, FormSection, Switch } from '@patternfly/react-core'
 import React, { useState } from 'react'
+import { string } from 'superstruct'
 import { log } from './globals'
 
+type LocalStorageFieldData<T> = {
+  localStorageKey: string,
+  defaultValue: T,
+  updateFunction: React.Dispatch<React.SetStateAction<T>>
+}
+
 export const HomePreferences: React.FunctionComponent = () => {
-  const [defaultVerticalNavState, setDefaultVerticalNavState] = useState(true)
+
+  const LOCAL_STORAGE_VERTICAL_NAVIGATION = "local_storage_vertical_navigation"
+  const DEFAULT_VALUE_VERTICAL_NAVIGATION = true;
+  
+  const VERTICAL_NAVIGATION_INITIAL_VALUE : boolean = 
+    localStorage.getItem(LOCAL_STORAGE_VERTICAL_NAVIGATION) !== null
+    ? localStorage.getItem(LOCAL_STORAGE_VERTICAL_NAVIGATION) === "true"
+    : DEFAULT_VALUE_VERTICAL_NAVIGATION;
+
+  const [defaultVerticalNavState, setDefaultVerticalNavState] = useState(VERTICAL_NAVIGATION_INITIAL_VALUE)
+
+  const VERTICAL_NAVIGATION_FIELD_PARAMETERS : LocalStorageFieldData<boolean> = {
+    localStorageKey: LOCAL_STORAGE_VERTICAL_NAVIGATION,
+    defaultValue: true,
+    updateFunction: setDefaultVerticalNavState
+  }
+
+  const FIELDS_TO_RESET = [VERTICAL_NAVIGATION_FIELD_PARAMETERS]
 
   const reset = () => {
-    // TODO: impl
-    log.info('TODO - Reset settings')
+    FIELDS_TO_RESET.forEach(
+      field => {
+        localStorage.removeItem(field.localStorageKey)
+        field.updateFunction(field.defaultValue)
+      }
+    )
+  }
+
+  const savingToLocalStorage = <T,>(keyToSave : string, updateFunction : React.Dispatch<React.SetStateAction<T>>) 
+    : React.Dispatch<React.SetStateAction<T>> => {
+    return (value) => {
+      localStorage.setItem(keyToSave, String(value))
+      return updateFunction(value)
+    };
   }
 
   const UIForm = () => (
@@ -16,7 +52,7 @@ export const HomePreferences: React.FunctionComponent = () => {
         label='Show vertical navigation'
         labelOff='Hide vertical navigation'
         isChecked={defaultVerticalNavState}
-        onChange={setDefaultVerticalNavState}
+        onChange={savingToLocalStorage(LOCAL_STORAGE_VERTICAL_NAVIGATION, setDefaultVerticalNavState)}
       />
     </FormGroup>
   )


packages/hawtio/src/preferences/HomePreferences.tsx
```diff
@@ -1,13 +1,46 @@
...
 
 export const HomePreferences: React.FunctionComponent = () => {
-  const [defaultVerticalNavState, setDefaultVerticalNavState] = useState(true)
+  const LOCAL_STORAGE_VERTICAL_NAVIGATION = 'local_storage_vertical_navigation'
+  const DEFAULT_VALUE_VERTICAL_NAVIGATION = true
+
+  const VERTICAL_NAVIGATION_INITIAL_VALUE: boolean =
+    localStorage.getItem(LOCAL_STORAGE_VERTICAL_NAVIGATION) !== null
+      ? localStorage.getItem(LOCAL_STORAGE_VERTICAL_NAVIGATION) === 'true'
+      : DEFAULT_VALUE_VERTICAL_NAVIGATION
```
joshiraez, line 17: I didn't find a clever way to put this. I need to first check if it's null to know if I need to use the DEFAULT_VERTICAL_NAVIGATION.

Then, on top of that, TS doesn't have a Boolean.parseBoolean like Java, so it ends being a bit ugly :D

packages/hawtio/src/preferences/HomePreferences.tsx
```diff
@@ -1,13 +1,46 @@
...
+
+  const [defaultVerticalNavState, setDefaultVerticalNavState] = useState(VERTICAL_NAVIGATION_INITIAL_VALUE)
+
+  const VERTICAL_NAVIGATION_FIELD_PARAMETERS: LocalStorageFieldData<boolean> = {
```
joshiraez, line 21: I'd like to have this at the top, instead of having to set all those consts first before hand, make this kind of field the "const" for each variable to be saved in local storage.

But I didn't find any good way, as I need to compute the initialization value to retrieve the update function, and then I can create this object.

If any one has any good ideas please chime in

phantomjinx (approved):

joshiraez:
I'm going to see if it's worth creating a custom hook on wednesday. If not, i'll rewrite it in a more simple way and merge

packages/hawtio/src/preferences/HomePreferences.tsx
```diff
@@ -1,13 +1,46 @@
...
+type LocalStorageFieldData<T> = {
+  localStorageKey: string
+  defaultValue: T
+  updateFunction: React.Dispatch<React.SetStateAction<T>>
+}
 
 export const HomePreferences: React.FunctionComponent = () => {
-  const [defaultVerticalNavState, setDefaultVerticalNavState] = useState(true)
+  const LOCAL_STORAGE_VERTICAL_NAVIGATION = 'local_storage_vertical_navigation'
+  const DEFAULT_VALUE_VERTICAL_NAVIGATION = true
```
tadayosi, line 12: The values `true` or `false` with this variable name doesn't convey the real meaning of what the flag does. It should be an enum with 'show' | 'hide' if you don't change the var name, or `true` | `false` with the var name something like `showVerticalNavByDefault`.

packages/hawtio/src/preferences/HomePreferences.tsx
tadayosi: You can always look at the code in the other places that interact with local storage.

packages/hawtio/src/preferences/HomePreferences.tsx
tadayosi: I don't see why we need this. It seems it just makes a simple thing a bit more complicated. You can simply persist the vertical nav state to local storage via the preferences service.

packages/hawtio/src/preferences/HomePreferences.tsx
```diff
@@ -1,13 +1,46 @@
...
+    defaultValue: true,
+    updateFunction: setDefaultVerticalNavState,
+  }
+
+  const FIELDS_TO_RESET = [VERTICAL_NAVIGATION_FIELD_PARAMETERS]
 
   const reset = () => {
-    // TODO: impl
-    log.info('TODO - Reset settings')
+    FIELDS_TO_RESET.forEach(field => {
```
tadayosi, line 30: why can't we just use `localStorage.clear()`?

It's always recommended to consult with the legacy impl to see if we might miss anything.
https://github.com/hawtio/hawtio-core/blob/main/src/app/preferences/reset-preferences/reset-preferences.controller.ts#L11-L29

packages/hawtio/src/preferences/HomePreferences.tsx
```diff
@@ -1,13 +1,46 @@
...
+      localStorage.removeItem(field.localStorageKey)
+      field.updateFunction(field.defaultValue)
+    })
+  }
+
+  const savingToLocalStorage = <T,>(
```
tadayosi, line 36: I don't think we need to do this in such a complex way.

tadayosi (changes requested):
Can you please look around the rest of the code and harmonise with its implementation styles?

Also, I'd like to follow the conventional commits style for commit messages. It's because in the future we might want to introduce some releasing tool that is based on the conventional commits and automate releasing of versions and changelogs.
https://github.com/hawtio/hawtio-next/blob/main/docs/developing.md#styles-conventions-and-guidelines

Also, please always consult with the legacy impl to see if we might miss anything. In this case, it's here:
https://github.com/hawtio/hawtio-core/tree/main/src/app/preferences

tadayosi:
Also, as this pull req is going to close #55, please implement the actual behaviours around the default vertical nav state flag. That means, you should read the default state from local storage here.
https://github.com/hawtio/hawtio-next/blob/main/packages/hawtio/src/ui/page/HawtioHeader.tsx#L29

joshiraez:
Im on holiday today but I will get onto it tomorrow first time.

No idea how the linter is passing my build here too considering Im using react hooks "ilegally" but, anyway, Ill get to see everything tomorrow. Lots of useful comments, thanks @tadayosi  (and Ill try to keep with the conventional styles too)

packages/hawtio/src/preferences/HomePreferences.tsx
joshiraez: Clear would remove all the settings, moreover, everything on the localstorage, not just the one on that page of the settings, so its generally recommended to avoid it.

Ill check it tomorrow anyway and show a gif too if it is as I think

packages/hawtio/src/preferences/HomePreferences.tsx
joshiraez: Clear would remove all the settings, moreover, everything on the localstorage, not just the one on that page of the settings, so its generally recommended to avoid it.

Ill check it tomorrow anyway and show a gif too if it is as I think

tadayosi: Yes, clear would remove all the settings for the same domain e.g. `http://localhost:3000` but not others. Since the `Reset` button is supposed to reset all the Hawtio settings, it should be what is expected to do. The old Hawtio actually has been with the behaviour. It's kind of hard reset. Do you still think we need to avoid this?
https://github.com/hawtio/hawtio-core/blob/main/src/app/preferences/reset-preferences/reset-preferences.controller.ts#L11-L29

When you look at the legacy impl, there are two things to consider:

First one is which keys should be reserved. Looking at the legacy impl, you'll see `jvmConnect` and `osAuthCreds` are kept. If we follow the same behaviour we should at least reserve the connection settings, i.e. `connect.connections`.

Second one is that we'd need to reload the page once you reset the local storage, since the operation should also clear Jolokia settings which require reload to apply the new settings.

joshiraez: Looking at the old implementation I think you're right that there is another way to do this with a preferences service for example. Let me think it through.

Also, yep, you are correct about resetting and I was wondering about that the other day.

I guess I'm going to rewrite it in a more simple way and taking into account all of this

packages/hawtio/src/preferences/HomePreferences.tsx
```diff
@@ -1,13 +1,46 @@
 import { Button, CardBody, Form, FormGroup, FormSection, Switch } from '@patternfly/react-core'
 import React, { useState } from 'react'
-import { log } from './globals'
+
+type LocalStorageFieldData<T> = {
```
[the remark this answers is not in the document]
joshiraez: It was just to comply with TS types because if I wanted to do a decorator for savingToLocalStorage function I do need to save the type if I don't want to just use any and remove all type checking.

I mean, there is a much easier way to do this just by doing something like

(value) => {saveToStorage(key, value); updateValue(value)} in the component but I was thinking in generalizing that. (Although I don't know if that would make react compain... any way, I'll be rewriting this using a service as you said below so it will be much simpler)

joshiraez:
I converted it as draft because I'm not entirely convinced and I need to iterate a bit more through it. I see the point of trying to simplify it but in the process it I feel like there is some hidden complexity that really shouldn't be there (or are fixed by side effects which I really don't see the point of having sort of fixed like that when we are updating it from scratch. Better to have a solid foundation). I'll keep working on this in monday.

I understand that watching some of the old code to see the old implementation should be a guideline, but I don't like some of the hidden complexity that I have seen in the old app and I think we can do better without too much effort

packages/hawtio/src/util/localStorage.ts
```diff
@@ -0,0 +1,37 @@
...
+    //Use this function to reset preferences. Some preferences of the local storage shouldn't
+    //be resetted. Add their fields to the list below so they can be backup and 
+    //restored into the local storage.
+
+    const LOCAL_STORAGE_FIELDS_TO_PRESERVE = [
+        STORAGE_KEY_JOLOKIA_OPTIONS,
+        STORAGE_KEY_UPDATE_RATE,
+        STORAGE_KEY_CONNECTIONS,
+        STORAGE_KEY_CURRENT_CONNECTION
+    ]
```
tadayosi, line 23: Among these keys, the only really necessary key is `STORAGE_KEY_CONNECTIONS`. Anything else shouldn't be preserved. I don't think we need to have `LOCAL_STORAGE_FIELDS_TO_PRESERVE` to manage the preserved key list, as it shouldn't grow in the near future.

packages/hawtio/src/util/localStorage.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    localStorage.clear();
+
+    backupFields.forEach(
+        (key, value) => localStorage.setItem(key, value)
+    )
+
+    //TODO: reload page on reset on connection preferences
+
+}
```
tadayosi, line 37: Overall, I don't think it's good idea to locate the code that manipulates local storage in `util`. We don't do so in the other places of the code base.

tadayosi (changes requested):

Issue #20, opened 2023-01-18
Camel - Routes view


tadayosi:
Camel plugin should provide the view for managing Camel routes from the web console.

It should provides the following features:

- [x] Start, stop, resume, and delete Camel routes
- [x] View ~and update~ the source of the routes on the fly (Updating routes on the fly is no longer supported in Camel 3 hawtio/hawtio#2779)
- [x] https://github.com/hawtio/hawtio-next/issues/144
- [x] https://github.com/hawtio/hawtio-next/issues/145

### Routes

![Routes](https://user-images.githubusercontent.com/156692/213158904-f47dda6f-88c9-415e-be91-e63979b3a0f6.png)

### Source

![Source](https://user-images.githubusercontent.com/156692/213165492-89c27991-9fe1-4a3f-abff-84ea3861fc52.png)

### Exchanges

![Exchanges](https://user-images.githubusercontent.com/156692/213165621-b737ca54-fccb-4957-960a-8ae78260e9bc.png)

### Type converters

![Type converters](https://user-images.githubusercontent.com/156692/213165725-130974d1-5f79-40ed-9e53-9b690826b810.png)

tadayosi:
@mmelko Can we close this?

mmelko:
@tadayosi yes I think so. Now we have source and routes views in place.

PR #137, opened 2023-03-08
Camel plugin

phantomjinx:
~Preparation work added for addition of Exchange Page~

The work now includes the completed Exchange Page as well as updates from the comments of this PR.


packages/hawtio/src/plugins/shared/tree/node.ts
```diff
@@ -113,15 +121,14 @@ export class MBeanNode implements TreeViewDataItem {
   create(name: string, folder: boolean): MBeanNode {
     // this method should be invoked on a folder node
     if (this.children === undefined) {
-      log.warn(`node "${this.name}" should be a folder`)
```
tadayosi, line 116: Just for checking my understanding, why is this warning removed?  When I added this `if` block, it was the assumption that this `create` method is not expected to be invoked on a leaf mbean node and thus `children` should be always there. That's why I added this log as warning. Has that assumption changed in any way?

packages/hawtio/src/plugins/shared/tree/node.ts
```diff
@@ -139,6 +146,11 @@ export class MBeanNode implements TreeViewDataItem {
 
     const remove: MBeanNode[] = this.children
     this.children = []
+
+    for (const r of remove) {
```
tadayosi, line 150: Nitpicky style comment but I'd prefer to use `Array.forEach` or similar functional methods over `for` statement in most cases. We'd use `for (const x of xs)` only when we need to execute some async/await functions in loop but sequentially.
https://zellwk.com/blog/async-await-in-loops/

tadayosi (approved):
There are a few comments but overall it looks great. This can be turned from draft to 'ready for push'.

packages/hawtio/src/plugins/shared/tree/node.ts
phantomjinx: The assumption has not changed but rather the logic of node creation does not quite work. I noticed that in the `sampleapp` a node (folder) is created but it has no children close by hence it is created as a node rather than a folder. However, later in the tree reading sequence the children are then added.

Consquently, this warning represents a false positive since the node is correctly converted to a folder and the children added correctly. If you need further details, I can did out which node it is occurring on.

packages/hawtio/src/plugins/shared/tree/node.ts
phantomjinx: I would as well but in this case, the for-loop must be synchronous in order that the `remove` children are returned correctly (with no parent). The `forEach` function cannot guarantee that and would require the unit test in node.test.ts to manage the asynchronousness.

phantomjinx:
Related issues -> https://github.com/hawtio/hawtio-next/issues/20 https://github.com/hawtio/hawtio-next/issues/144

packages/hawtio/src/plugins/shared/tree/node.ts
tadayosi: Ah, I see. I think it is the case when we have two mbeans with names like `domain:aaa=xxx,bbb=yyy` and `domain:aaa=xxx,bbb=yyy,ccc=zzz`. In this case, at the first mbean `bbb=yyy` node is created as a leaf, but then at the second one it is considered as a folder since it has a child `ccc=zzz`.

The whole process of tree generation needs to be revisited with such an edge case at some point. For now, I agree with removing the log line.

packages/hawtio/src/plugins/shared/tree/node.ts
tadayosi: > The forEach function cannot guarantee that

I don't think so. JavaScript is single-thread model so it always runs synchronously unless you explicitly use setTimeout or async/await in the `forEach` callback. So it should guarantee the order. What I said in the first comment is about when you need to execute some async function or setTimeout inside a loop. Otherwise you should always be able to use `Array.forEach`.

tadayosi: But for now let's mark it as resolved and move on.

tadayosi (approved):
LGTM. Thanks!

Issue #145, opened 2023-03-13
Camel - Routes - Type Converters tab


tadayosi:
Enable and view statistics for type converters in the routes

phantomjinx:
Cannot provide data for the TypeConverter table due to #151

tadayosi:
OK, then we can push this feature from 0.2.0 to future releases. We can work on it once we come up with an alternative way to implement it, or decide not to do it if it doesn't make sense to provide the feature for Camel 3 at all.

tadayosi:
According Claus, it doesn't make sense to provide this feature for Camel 3 https://github.com/hawtio/hawtio-next/issues/151#issuecomment-1471428156. Let's just remove the feature from Hawtio NG.

Issue #151, opened 2023-03-15
TypeConverter View cannot populate type converters table


phantomjinx:
The TypeConverter view is not able to populate the table of TypeConverter from/to types since the JMX operation "listTypeConverters" was removed in the Camel upgrade from [Camel 3.6 -> 3.7](https://camel.apache.org/manual/camel-3x-upgrade-guide-3_7.html#_type_converters).

This also affects current releases of hawtio. The table will only display "No data available in table" and the console will show the error:
```
Possibly unhandled rejection: java.lang.IllegalArgumentException : No operation listTypeConverters found on MBean org.apache.camel:context=SampleCamel,name=DefaultTypeConverter,type=services
```

-----
related to #145

tadayosi:
@davsclaus Is there an alternative way to list Type Converters defined in a Camel context for the recent versions of Camel 3?  Considering the optimisation done for type converters on Camel 3, doesn't it even make sense for a console like Hawtio to provide the Type Converters view at all?

davsclaus:
No just remove that view

tadayosi:
Thanks Claus, then let's remove this feature from Hawtio NG.

PR #155, opened 2023-03-16
feat(#145): Provides TypeConverters view

phantomjinx:
* jolokia-service
 * View that requires a writeAttribute function so implemented in same style as the 'read' function

* Contexts
 * Narrows the context to use the CamelContext rather than the more general plugin context. Allows for fewer imports when running jest unit tests

### DIFF
--- a/packages/hawtio/src/plugins/camel/CamelContent.tsx
+++ b/packages/hawtio/src/plugins/camel/CamelContent.tsx
@@ -20,6 +20,7 @@ import { CamelContext } from './context'
 import { Attributes, Operations, Chart, JmxContentMBeans, MBeanNode } from '@hawtiosrc/plugins/shared'
 import { Contexts } from './contexts'
 import { Exchanges } from './exchanges'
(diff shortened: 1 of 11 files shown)


tadayosi (changes requested):
So according to Claus https://github.com/hawtio/hawtio-next/issues/151#issuecomment-1471428156 we don't need Type Converters view for Hawtio.next. If there are reusable parts for other features in this pull req, just extract them and create another pull req.

phantomjinx:
> So according to Claus [#151 (comment)](https://github.com/hawtio/hawtio-next/issues/151#issuecomment-1471428156) we don't need Type Converters view for Hawtio.next. If there are reusable parts for other features in this pull req, just extract them and create another pull req.

But the TypeConverters statistics portion of the view is still applicable. Should this element be not retained?

Screenshot_20230317_140718

tadayosi (approved):
@phantomjinx Yes, you're right. Let's merge this.

PR #156, opened 2023-03-16
feat: Add routes and sources pages

mmelko:
Add route controls for start,stop,delete
Add monaco-editor dependencies
Related to #20 
Routes view: 
Screenshot 2023-03-16 at 22 38 55

I have to figure out why Xml isn't "colorized", probably missing webpack plugin: 
Screenshot 2023-03-16 at 22 38 29


packages/hawtio/src/plugins/camel/routes-service.ts
tadayosi, line 1: If this service is supposed to be used mainly from `CamelRoutes.tsx`, this should be put under `routes/` directory as well.

packages/hawtio/src/plugins/camel/routes/CamelRoutes.tsx
```diff
@@ -0,0 +1,313 @@
+import React, { useContext, useEffect, useState } from 'react'
+import { CamelContext } from '@hawtiosrc/plugins/camel/context'
+import { CamelRoute, routesService } from '@hawtiosrc/plugins/camel/routes-service'
```
tadayosi, line 3: Use relative path if you put `routes-service.ts` under `routes/`.

tadayosi (commented):
Thanks. Overall it looks great. There are a few minor comments, plus CI check failures that should be fixed.

packages/hawtio/src/plugins/camel/routes/CamelRoutes.tsx
```diff
@@ -0,0 +1,313 @@
...
+import { eventService } from '@hawtiosrc/core'
+import { workspace } from '@hawtiosrc/plugins/shared'
+
+export const CamelRoutes: React.FunctionComponent = () => {
+  const { selectedNode } = useContext(CamelContext)
+  const [routes, setRoutes] = useState<CamelRoute[]>([])
+  const [selected, setSelected] = useState<Map<string, boolean>>(new Map())
+  const [isOpen, setIsOpen] = useState<boolean>(false)
+  const [isConfirmDeleteOpen, setIsConfirmDeleteOpen] = useState(false)
+  const getRouteAttributes = () => {
```
tadayosi, line 30: You can make this `async` and inside the definition you can await for `routesService.getRoutesAttributes(selectedNode)` instead of chaining with `then`.

packages/hawtio/src/plugins/camel/routes/CamelRoutes.tsx
```diff
@@ -0,0 +1,313 @@
...
+  ToolbarItem,
+} from '@patternfly/react-core'
+import { AsleepIcon, InfoCircleIcon, PlayIcon, Remove2Icon } from '@patternfly/react-icons'
+import { eventService } from '@hawtiosrc/core'
+import { workspace } from '@hawtiosrc/plugins/shared'
+
+export const CamelRoutes: React.FunctionComponent = () => {
+  const { selectedNode } = useContext(CamelContext)
+  const [routes, setRoutes] = useState<CamelRoute[]>([])
+  const [selected, setSelected] = useState<Map<string, boolean>>(new Map())
```
tadayosi, line 27: We haven't used `Map` anywhere else yet. Perhaps you can instead use `Record<string, boolean>` for consistency with other code.

packages/hawtio/src/plugins/camel/routes/CamelRoutes.tsx
```diff
@@ -0,0 +1,313 @@
...
+    selected.forEach((value, key) => {
+      const route = routes.find(r => r.RouteId === key)
+      if (value && route && route.State === state) res = value
+    })
+    return res
+  }
+
+  function onStopClicked() {
+    let stoppedCount = 0
+    selected.forEach(async (value, key) => {
```
tadayosi, line 80: does `forEach` work with `asycn` callback?  I remember it doesn't for `Array.forEach()`. Is it different for `Map`?

packages/hawtio/src/plugins/camel/routes/CamelRoutes.tsx
```diff
@@ -0,0 +1,313 @@
...
+        type: 'success',
+        message: `${stoppedCount} routes stopped successfully`,
+      })
+    }
+    // workspace.refreshTree()
+  }
+
+  function onStartClicked() {
+    let startedCount = 0
+    selected.forEach(async (value, key) => {
```
tadayosi, line 108: same here. does `async` call work for `forEach`?

packages/hawtio/src/plugins/camel/routes/Source.tsx
```diff
@@ -0,0 +1,48 @@
+import React, { useContext, useEffect, useState } from 'react'
+import { jolokiaService } from '@hawtiosrc/plugins'
+import { CodeEditor, Language } from '@patternfly/react-code-editor'
+import { isRoutesFolder } from '@hawtiosrc/plugins/camel/camel-content-service'
+import { CamelContext } from '@hawtiosrc/plugins/camel/context'
+import { MBeanNode } from '@hawtiosrc/plugins/shared'
+
+export const findParentContext = (child: MBeanNode, root: MBeanNode): MBeanNode | undefined => {
```
tadayosi, line 8: why export this? is it supposed to be used anywhere else?

packages/hawtio/src/plugins/camel/routes-service.ts
phantomjinx: Not the case. The routes-services contains methods use by the tree-processor when re-aligning the JMX tree into the Camel plugin configuration. I wonder whether these two 'different' routes services should be kept separate despite the collision (?) of names? What I mean is that
1. the routes-service dedicated to shredding the routes XML and converting it to tree nodes;
2. the routes-service dedicated to contacting the jolokia service and retrieving the data necessary for populating the Routes View.

packages/hawtio/src/plugins/camel/routes/CamelRoutes.tsx
```diff
@@ -0,0 +1,313 @@
...
+      routes.forEach(route => {
+        selected.set(route.RouteId, selected.get(route.RouteId) ?? false)
+      })
+      setSelected(tmpSelected)
+      setRoutes(routes)
+    })
+  }
+  useEffect(() => {
+    getRouteAttributes()
+  }, [])
```
phantomjinx, line 42: `getRoutesAttributes()` should probably be placed inside the `useEffect` since the linter will complain that it is not included as a dependency. By simply adding it to the dependencies array, an infinite loop will result and it will be necessary to surround the function in a `[useCallback](https://react.dev/reference/react/useCallback)`. However, this has ended up being a lot of extra work when the entire function can be moved into the `useEffect`.

The dependency array should contain the `selectedNode`.

packages/hawtio/src/plugins/camel/routes/CamelRoutes.tsx
```diff
@@ -0,0 +1,313 @@
...
+  function onStopClicked() {
+    let stoppedCount = 0
+    selected.forEach(async (value, key) => {
+      if (value) {
+        const route = routes.find(r => r.RouteId === key)
+        if (route && route.State === 'Started') {
+          try {
+            await routesService.stopRoute(route.objectName as string)
+            stoppedCount++
+            getRouteAttributes()
```
phantomjinx, line 87: With this probably being the reason why `getRouteAttributes()` is outside the `useEffect`, may I suggest in this case that a flag is adopted instead that triggers the `useEffect` to run again rather than calling the function directly?

packages/hawtio/src/plugins/camel/routes/CamelRoutes.tsx
```diff
@@ -0,0 +1,313 @@
...
+  function onStartClicked() {
+    let startedCount = 0
+    selected.forEach(async (value, key) => {
+      if (value) {
+        const route = routes.find(r => r.RouteId === key)
+        if (route && route.State === 'Stopped') {
+          try {
+            await routesService.startRoute(route.objectName as string)
+            startedCount++
+            getRouteAttributes()
```
phantomjinx, line 115: Same as above comment.

packages/hawtio/src/plugins/camel/routes/CamelRoutes.tsx
```diff
@@ -0,0 +1,313 @@
...
+        <Card>
+          <CardBody>
+            <Text component='p'>
+              <InfoCircleIcon /> This context does not have any routes.
+            </Text>
+          </CardBody>
+        </Card>
+      )}
+      {routes.length > 0 && (
+        <TableComposable aria-label='Simple table' variant={'compact'}>
```
phantomjinx, line 244: In `InflightExchanges`, I used the react legacy table. If we are to go with `<TableComposable>` here then perhaps an issue to convert the `inflight` table to that format too?

packages/hawtio/src/plugins/camel/routes/Source.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  return found
+}
+export const Source: React.FunctionComponent = () => {
+  const { selectedNode, tree } = useContext(CamelContext)
+  const [xmlSource, setxmlSource] = useState<string>('')
+
+  useEffect(() => {
+    if (selectedNode) {
+      if (isRoutesFolder(selectedNode)) {
```
phantomjinx, line 26: Is this necessary given that the Source View should not even be displayed if the selectedNode is not a routes folder?

packages/hawtio/src/plugins/camel/routes/Source.tsx
```diff
@@ -0,0 +1,48 @@
...
+        const contextMbean = findParentContext(selectedNode, tree.getTree()[0])
+
+        if (contextMbean?.objectName) {
+          jolokiaService.execute(contextMbean.objectName, 'dumpRoutesAsXml()').then(xml => {
```
phantomjinx, line 30: Since the `routes-service` has already performed this [function](https://github.com/hawtio/hawtio-next/blob/main/packages/hawtio/src/plugins/camel/routes-service.ts#L104) when run from the `tree-processor`, might it be useful to simply cache the xml on the node and then get it rather than executing the same request a second time?

phantomjinx (changes requested):
A couple of things may need to be discussed and confirmed. Solving the linting errors will probably get to where we need to be before anythingelse.

packages/hawtio/src/plugins/camel/routes/Source.tsx
mmelko: yeah, that make sense.

packages/hawtio/src/plugins/camel/routes/Source.tsx
mmelko: that's true, It's just that selected node has `MBean | null` type so isRouteFolder method signature would need to be changed then.

packages/hawtio/src/plugins/camel/routes/Source.tsx
mmelko: I had it in the different file. Thanks for notify.

packages/hawtio/src/plugins/camel/routes/CamelRoutes.tsx
mmelko: oh you're right! It's a coincidence it worked for me because re-triggering the render. I will refactor this to use `for _ of ` loop instead.

packages/hawtio/src/plugins/camel/routes/CamelRoutes.tsx
mmelko: I noticed that plain array is used in Contexts component so I can refactor it to use that if you want.

packages/hawtio/package.json
```diff
@@ -47,9 +48,11 @@
     "jolokia.js": "1.7.2",
     "jquery": "^3.6.4",
     "js-logger": "^1.6.1",
+    "monaco-editor": "^0.36.1",
```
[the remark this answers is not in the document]
mmelko: so yes we do If we want to use monaco-editor which is transitive dependency of patternfly-code-editor. The truth is that it causing that errors in the tests so I can use code-block for now if you agree.

packages/hawtio/src/plugins/camel/routes/Source.tsx
phantomjinx: No. `isRouteFolder` is able to stay as that signature if this `useEffect` has an `if (!selectedNode) return null` clause at the beginning, eg. [here](https://github.com/hawtio/hawtio-next/blob/main/packages/hawtio/src/plugins/camel/exchanges/InflightExchanges.tsx#L15).

mmelko: Ah so I misread your comment before. The think is that Source component is displayed for both route folders and routes itself. That's why I have it there.

phantomjinx: Indeed. I see your point. What we could do is if `xml` is cached as a property on the `routes` node in the same manner as the `else` clause of this file (L34 below) then we will not have to differentiate between the types of node. Feel free to ping me if that makes no sense.

packages/hawtio/src/plugins/camel/routes-service.ts
tadayosi: I see. Then it depends on how cohesive the two services should be. If they are unrelated and share very little logic, they should be kept separated. Fow now, the name collision is fine so long as they are placed in different modules. If you can come up with better naming it'd be great to rename them though. Maybe the first service can be named something like `route-xml-service` as its sole resposibility is to process route XML.

packages/hawtio/src/plugins/camel/routes/CamelRoutes.tsx
tadayosi: We should eventually migrate all tables used in Hawtio next to the new `<TableComposable>`, so I think we can leave it as is for this page. We should replace the usages of legacy table with `TableComposable` in the other places including Camel Inflight and JMX attributes etc. step by step. But it's not high priority yet.

Let's choose `TableComposable` when creating a new view from now on.

packages/hawtio/src/plugins/camel/routes-service.ts
phantomjinx: `route-xml-service` would seem a sensible refactoring solution.

mmelko:
@tadayosi @phantomjinx I've incorporated most of the things:
I've replaced resolving promises with `await/async`
fixed added monaco-editor webpack plugin to properly work with XML
-added mocks for the monaco to pass the tests

My suggestion is to refactor routes-service later (when I will be working on visualisation for example)
Only thing missing is to cache XMLs. But you can review in the meantime.

mmelko:
@phantomjinx I also refactored the route service to cache the XML and limit calling. Please check

packages/hawtio/src/plugins/camel/routes/CamelRoutes.tsx
phantomjinx: Issue logged. https://github.com/hawtio/hawtio-next/issues/180

packages/hawtio/src/plugins/camel/routes/CamelRoutes.tsx
```diff
@@ -0,0 +1,294 @@
...
+} from '@patternfly/react-core'
+import { AsleepIcon, InfoCircleIcon, PlayIcon, Remove2Icon } from '@patternfly/react-icons'
+import { eventService } from '@hawtiosrc/core'
+import { workspace } from '@hawtiosrc/plugins/shared'
+
+export const CamelRoutes: React.FunctionComponent = () => {
+  const { selectedNode } = useContext(CamelContext)
+  const [routes, setRoutes] = useState<CamelRoute[]>([])
+  const [reload, setReload] = useState<boolean>(false)
+  // const [selected, setSelected] = useState<Map<string, boolean>>(new Map())
```
phantomjinx, line 28: Should this be removed?

packages/hawtio/src/plugins/camel/routes/Source.tsx
```diff
@@ -0,0 +1,34 @@
...
+  return found
+}
+export const Source: React.FunctionComponent = () => {
+  const { selectedNode, tree } = useContext(CamelContext)
+  const [xmlSource, setxmlSource] = useState<string>('')
+
+  useEffect(() => {
+    const xml = selectedNode?.getProperty('xml')
+    if (xml) {
+      setxmlSource(xml)
```
phantomjinx, line 25: Really small thing but should there be an error message applied to `setxmlSource` if xml is null. Obviously, the likelihood it is null will be very small.

packages/hawtio/src/plugins/camel/routes/Source.tsx
```diff
@@ -0,0 +1,34 @@
...
+    }
+  }, [tree, selectedNode])
```
phantomjinx, line 27: Does this need to depend on `tree`?

phantomjinx:
Some very minor things that could just be updated but otherwise LGTM!

joshiraez (approved):
lgtm

packages/hawtio/src/plugins/camel/routes/CamelRoutes.tsx
mmelko: whoops

packages/hawtio/src/plugins/camel/routes/Source.tsx
mmelko: thanks. I added an error pop up if xml will be null

packages/hawtio/src/__mocks__/codeEditorMock.js
tadayosi, line 1: If this becomes anything more than one liner, let's consider making it `.ts`. It's not a big deal, so we can do so later.

packages/hawtio/src/plugins/camel/routes/CamelRoutes.tsx
tadayosi: let's not do so for now.

packages/hawtio/src/plugins/camel/routes/Source.tsx
tadayosi: Is it still relevant?

packages/hawtio/src/plugins/camel/routes/Source.tsx
```diff
@@ -0,0 +1,40 @@
...
+export const Source: React.FunctionComponent = () => {
+  const { selectedNode } = useContext(CamelContext)
+  const [xmlSource, setxmlSource] = useState<string>('')
+
+  useEffect(() => {
+    const xml = selectedNode?.getProperty('xml')
+    if (xml) {
+      setxmlSource(xml)
+    } else {
+      eventService.notify({
```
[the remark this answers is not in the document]
tadayosi: In general, I'd like to keep the same level of frequency with the legacy Hawtio for using notifications to users, unless there are good reasons to add or remove notifications for usability.

packages/hawtio/src/plugins/camel/routes-service.test.ts
```diff
@@ -70,7 +75,7 @@ describe('routes-service', () => {
   test('processRouteXml wrong mbean', async () => {
```
tadayosi, line 75: So now it's a test for `getRoutesXml`, so please change the description as well.

tadayosi (commented):
A few more comments.

packages/hawtio/src/__mocks__/codeEditorMock.js
mmelko: just switching to ts and change in moduleNamemapper leads to failed tests again so lets keep it like that for now.

tadayosi (approved):
LGTM. Let's fix lint error and merge it.

Issue #154, opened 2023-03-16
Remove tilde from the name of MBeans node in Camel tree


tadayosi:
Right now, in the Camel tree the MBean node has tilde before its name, `~MBean`. I guess it's because of convenience of sorting for putting it to the bottom in order, but ideally we should remove tilde by improving the sort argorithm in the Camel plugin.

![Camel tree](https://user-images.githubusercontent.com/156692/225554821-c3dfe98c-f298-4eac-b9b4-604ec1b69d08.png)

tadayosi:
@phantomjinx Is my understanding correct? Is it easy to fix this?

phantomjinx:
> @phantomjinx Is my understanding correct? Is it easy to fix this?

Yes. That's right. It was a stop-gap so I'll look at removing it next.

PR #159, opened 2023-03-17
fix(#154): Removes the tilde as a prefix to camel -> MBeans

phantomjinx:
* tilde unnecessary for sorting as the node is added last to its parent and so is in the correct location despite its name


tadayosi (approved):

tadayosi:
@phantomjinx Is that all we need to fix the issue?

Issue #67, opened 2023-02-01
Connect - Import connections


tadayosi:
`Export connections` is already available, but `Import connections` is not yet implemented.

This feature should basically just import the JSON file exported using `Export connections` and then persist the config into local storage.

![Import connections](https://user-images.githubusercontent.com/156692/215955616-bc1ec4bd-4194-4a01-b0bd-c960078fbebd.png)

shubhamku044:
Can I work on this?

joshiraez:
Working on this

PR #160, opened 2023-03-17
wip(#67): Import connection modal core logic

joshiraez:
Business and component logic are ready. But I'm unable to get yarn start to run it so I suspect there is some kind of muted error somewhere even though it finishes building. I'll continue on it on monday.


tadayosi (approved):
thanks.

PR #164, opened 2023-03-19
fix(build): fix generate:camel-icons script and regen camel icons index

tadayosi:
When I run `generate:camel-icons` script it fails with the following error.
```
$ yarn workspace @hawtio/react generate:camel-icons
/home/tasato/projects/hawtio2/hawtio-next/packages/hawtio/scripts/generate-camel-svg-index.js:6
const svgdir = path.resolve(__dirname, ROOT, 'src', 'plugins', 'camel', 'icons', 'svg')
                    ^

TypeError: path.resolve is not a function
    at Object.<anonymous> (/home/tasato/projects/hawtio2/hawtio-next/packages/hawtio/scripts/generate-camel-svg-index.js:6:21)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
    at node:internal/main/run_main_module:17:47
```

Also, it seems the generated index hasn't been updated.


oscerd (approved):

tadayosi:
Thanks @oscerd :slightly_smiling_face:

joshiraez (approved):
svg/index.ts is basically sorted alphabetically now, right?

Also it's strange the change from import to require, as import is usually recommended. Glad that it just works

tadayosi:
> svg/index.ts is basically sorted alphabetically now, right?

Mostly, yes. It depends on how the script is implemented, but I think it's ok so long as it's consistent, i.e. if it generates the identical index file for the same set of icon files.

Yeah, I also wonder why `import` doesn't work. If you are interested, please investigate and fix it back to work with `import`. For now, it's working so I'll go ahead and merge it.

tadayosi:
The issue with imports filed at #188

Issue #41, opened 2023-01-30
Camel - Help & preferences


tadayosi:
#### Help

![Help](https://user-images.githubusercontent.com/156692/215407541-d64304e8-49df-4b71-8d2b-6501547a7067.png)

#### Preferences

![Preferences](https://user-images.githubusercontent.com/156692/215407663-e559c11d-cb8b-4733-adae-57fe911fd15d.png)

joshiraez:
Can I pick this one up as future after finishing my current 2 tasks?

tadayosi:
Thanks, you can work on this next.

joshiraez:
For the preferences page, old version didn't have a Reset button. Do we want one for this one too? I can see that in Connect Preferences (previously Jolokla preferences) there wasn't a reset button either and we implemented it for this version

lhein:
both PRs are merged. Can this issue be closed?

joshiraez:
The only thing left to do is the consumers side, whose components might have not yet be implemented. Ill check tomorrow and see which open issues for the camel plugin are related to the preferences and link them

tadayosi:
@joshiraez Can you file a new issue to track the task?  Since this one is already closed and a bit inconvenient to track todos with it.

joshiraez:
Sure

PR #191, opened 2023-03-22
feat(#41): Camel Help

joshiraez:
One part of #41 . Setting this up as it is straightforward enough and it can be merged already. I'll keep working on the preferences page.

### DIFF
--- a/packages/hawtio/src/plugins/camel/help.md
+++ b/packages/hawtio/src/plugins/camel/help.md
@@ -1,0 +1,15 @@
+### Camel
+
+Click [Camel](camel/contexts) in the top navigation bar to view all the running Camel Contexts in the current JVM. (The selection will not appear on the navigation bar if there is no Camel running).
+
+The Camel plugin allows you to view all the running Camel applications in the current JVM.
+You can among others see the following details:
+
+- Lists of all running Camel applications
+- Detailed information of each Camel Context such as Camel version number, runtime statics
+- Lists of all routes in each Camel applications and their runtime statistics
+- Manage the lifecycle of all Camel applications and their routes, so you can restart / stop / pause / resume, etc.
+- Graphical representation of the running routes along with real time metrics
+- Live tracing and debugging of running routes
+- Profile the running routes with real time runtime statics; detailed specified per processor
+- Browsing and sending messages to Camel endpoint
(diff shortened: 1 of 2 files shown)


joshiraez:
Mandatory pic so you don't have to build it

image

packages/hawtio/src/plugins/camel/help.md
```diff
@@ -0,0 +1,15 @@
+### Camel
```
tadayosi, line 1: The top section level for a help should be `##` (h2), because the topmost level h1 is reserved for the title *Help* in the page. Check the other helps for reference.

packages/hawtio/src/plugins/camel/index.ts
```diff
@@ -21,6 +21,6 @@ export const camel = () => {
   })
 
   treeProcessorRegistry.add('camel', camelTreeProcessor)
-  // helpRegistry.add('camel', 'Camel', help, 11)
+  helpRegistry.add('camel', 'Camel', help, 11)
```
tadayosi, line 24: The number passed as the 4th arg should be `13`, which affects order to be displayed in Help view. Currently, single digits are reserved for core helps (Home: 1, Preferences: 2), and the builtin plugins use numbers starting from 11 (Connect: 11, JMX: 12, ...).

The default number is `100`, and we let users to use numbers >= 100 for their own custom plugins. (Then, does it make more sense to change the default to `1000`?)

tadayosi (changes requested):
Overall, it looks great. There are two minor comments to be addressed before merging.

packages/hawtio/src/plugins/camel/index.ts
tadayosi: @phantomjinx @mmelko @jsolovjo ^^^ FYI, as probably I haven't explained that yet explicitly.

joshiraez: Oh, I see. Nice approach to that. I didn't know what the number should be and I thought it was already commented with the good value. I'll update it.

I don't think we need to increase it to 1000... Having 80+ plugins would be a nice achievement by then :D

tadayosi (approved):

packages/hawtio/src/plugins/camel/index.ts
tadayosi: Why I think maybe we should change default to `1000` is not because we hope to have 80+ more plugins :slightly_smiling_face: but because if it's `100` random custom plugins always take precedence over the custom ones with clear order definitions such as `101`, `102`, ... But probably it's ok to leave it as `100` and let users manage their own plugins order.

Issue #173, opened 2023-03-20
Input box get unfocused on type


joshiraez:
There is a known bug with react where if you were to use useState for retaining the state of an input field, the field will get unfocused on every keystroke because the state procs a re-rendering in React

Right now the only known place where this is happening is Connect Preferences that I know of, but I guess it'd happen to any other input fields we use.

https://stackoverflow.com/questions/42573017/in-react-es6-why-does-the-input-field-lose-focus-after-typing-a-character

There is a few solutions there that seems like quick fixes. I haven't had luck with them and we might need a component rewrite.

Another band-aid solution is to use the debounce fuction so people will be able to type their number whole. The input will be still unfocused but it wont be as bad UX

joshiraez:
So a interesting addition is that this PR doesn't has this bug even though it uses a similar code to the others. It would be a nice start to figure out why it happens #192

PR #192, opened 2023-03-22
feat(#41): Camel preferences

joshiraez:
image

This together with #191 completes #41 

Notes:
- Right now values are being saved to local storage and little else. Would like to get your input in if any components already exist to put the wiring on them.
- Reset button. Should I add it? Home preferences didn't have a Reset button before and it has it, same for Jolokla settings. I guess it's a yes but wanted to confirm.
- Hint icons. I saw we dropped this from other preferences pages. Should I file a separate issue to implement them on all preferences pages?
- Overuse of arrow functions on the component: trying to reduce the amount of duplicated code for type checking, I had to use arrow functions because I lose the "this" when passing them as parameters. Is not too much boilerplate... until prettier comes up and puts everything in different lines :(. If
  - I remember that the correct pattern to avoid that was to store this in a const variable in the class, but typescript seemed to not like it.
  
I think that's all I will otherwise put another comment tomorrow

### DIFF
--- a/packages/hawtio/src/plugins/camel/CamelPreferences.tsx
+++ b/packages/hawtio/src/plugins/camel/CamelPreferences.tsx
@@ -14,7 +14,7 @@ const CamelPreferencesForm: React.FunctionComponent = () => {
   const [camelPreferences, setCamelPreferences] = useState(camelPreferencesService.loadCamelPreferences())
 
   const updatePreferences = (value: boolean | number, key: keyof ICamelPreferences): void => {
-    const updatedPreferences = { ...camelPreferences, ...{ key: value } }
+    const updatedPreferences = { ...camelPreferences, ...{ [key]: value } }
 
     camelPreferencesService.saveCamelPreferences(updatedPreferences)
     setCamelPreferences(updatedPreferences)
@@ -31,9 +31,9 @@ const CamelPreferencesForm: React.FunctionComponent = () => {
     }
   }
 
-  const updateCheckboxValueFor = (key: keyof ICamelPreferences): ((value: boolean, _: any) => void) => {
+  const updateCheckboxValueFor = (key: keyof ICamelPreferences): ((value: boolean, _: React.FormEvent) => void) => {
     //Utility function generator to reduce boilerplate
-    return (value: boolean, _: any) => {
+    return (value: boolean, _: React.FormEvent) => {
       updatePreferences(value, key)
     }
   }
--- a/packages/hawtio/src/plugins/camel/CamelPreferences.tsx
+++ b/packages/hawtio/src/plugins/camel/CamelPreferences.tsx
@@ -1,0 +1,191 @@
+import { CardBody, Checkbox, Form, FormGroup, FormSection, TextInput } from '@patternfly/react-core'
+import React, { useState } from 'react'
+import { storageService } from './camel-preferences-storage-service'
+
+export const CamelPreferences: React.FunctionComponent = () => (
+  <CardBody>
+    <Form isHorizontal>
+      <CamelPreferencesForm />
+    </Form>
+  </CardBody>
+)
+
+const CamelPreferencesForm: React.FunctionComponent = () => {
+  const [isHideOptionDocumentation, setIsHideOptionDocumentation] = useState(
+    storageService.loadIsHideOptionDocumentation(),
+  )
+  const [isHideDefaultOptionValues, setIsHideDefaultOptionValues] = useState(
+    storageService.loadIsHideDefaultOptionValues(),
+  )
+  const [isHideUnusedOptionValues, setIsHideUnusedOptionValues] = useState(
+    storageService.loadIsHideUnusedOptionValues(),
+  )
+  const [isIncludeTraceDebugStreams, setIsIncludeTraceDebugStreams] = useState(
+    storageService.loadIsIncludeTraceDebugStreams(),
+  )
+  const [maximumTraceDebugBodyLength, setMaximumTraceDebugBodyLength] = useState(
+    storageService.loadMaximumTraceDebugBodyLength(),
+  )
+  const [maximumLabelWidth, setMaximumLabelWidth] = useState(storageService.loadMaximumLabelWidth())
+  const [isIgnoreIDForLabel, setIsIgnoreIDForLabel] = useState(storageService.loadIsIgnoreIDForLabel())
+  const [isShowInflightCounter, setIsShowInflightCounter] = useState(storageService.loadIsShowInflightCounter())
+  const [routeMetricMaximumSeconds, setRouteMetricMaximumSeconds] = useState(
+    storageService.loadRouteMetricMaximumSeconds(),
+  )
+
+  const onBooleanValueUpdated = (
+    value: boolean,
+    updateFunction: (value: boolean) => void,
+    saveToStorageFunction: (value: boolean) => void,
+  ): void => {
+    updateFunction(value)
+    saveToStorageFunction(value)
+  }
+
+  const onNumberValueUpdated = (
+    value: string,
+    updateFunction: (value: number) => void,
+    saveToStorageFunction: (value: number) => void,
+  ): void => {
+    const intValue = parseInt(value)
+
+    if (!intValue) return
+
+    updateFunction(intValue)
+    saveToStorageFunction(intValue)
+  }
+
+  const onIsHideOptionDocumentationChange = (value: boolean) =>
+    onBooleanValueUpdated(
+      value,
+      value => setIsHideOptionDocumentation(value),
+      value => storageService.setIsHideOptionDocumentation(value),
+    )
+
+  const onIsHideDefaultOptionValuesChange = (value: boolean) =>
+    onBooleanValueUpdated(
+      value,
+      value => setIsHideDefaultOptionValues(value),
+      value => storageService.setIsHideDefaultOptionValues(value),
+    )
+
+  const onIsHideUnusedOptionValuesChange = (value: boolean) =>
+    onBooleanValueUpdated(
+      value,
+      value => setIsHideUnusedOptionValues(value),
+      value => storageService.setIsHideUnusedOptionValues(value),
+    )
+
+  const onIsIncludeTraceDebugStreamsChange = (value: boolean) =>
+    onBooleanValueUpdated(
+      value,
+      value => setIsIncludeTraceDebugStreams(value),
+      value => storageService.setIsIncludeTraceDebugStreams(value),
+    )
+
+  const onMaximumTraceDebugBodyLengthChange = (value: string) =>
+    onNumberValueUpdated(
+      value,
+      value => setMaximumTraceDebugBodyLength(value),
+      value => storageService.setMaximumTraceDebugBodyLength(value),
+    )
+
+  const onMaximumLabelWidthChange = (value: string) =>
+    onNumberValueUpdated(
+      value,
+      value => setMaximumLabelWidth(value),
+      value => storageService.setMaximumLabelWidth(value),
+    )
+
+  const onIsIgnoreIDForLabelChange = (value: boolean) =>
+    onBooleanValueUpdated(
+      value,
+      value => setIsIgnoreIDForLabel(value),
+      value => storageService.setIsIgnoreIDForLabel(value),
+    )
+
+  const onIsShowInflightCounterChange = (value: boolean) =>
+    onBooleanValueUpdated(
+      value,
+      value => setIsShowInflightCounter(value),
+      value => storageService.setIsShowInflightCounter(value),
+    )
+
+  const onRouteMetricMaximumSecondsChange = (value: string) =>
+    onNumberValueUpdated(
+      value,
+      value => setRouteMetricMaximumSeconds(value),
+      value => storageService.setRouteMetricMaximumSeconds(value),
+    )
+
+  return (
+    <FormSection>
+      <FormGroup label='Hide option documentation' fieldId='camel-form-hide-option-documentation'>
+        <Checkbox
+          id='camel-form-hide-option-documentation'
+          isChecked={isHideOptionDocumentation}
+          onChange={(value, _) => onIsHideOptionDocumentationChange(value)}
+        />
+      </FormGroup>
+      <FormGroup label='Hide default options values' fieldId='camel-form-hide-default-option-value'>
+        <Checkbox
+          id='camel-form-hide-default-option-value'
+          isChecked={isHideDefaultOptionValues}
+          onChange={(value, _) => onIsHideDefaultOptionValuesChange(value)}
+        />
+      </FormGroup>
+      <FormGroup label='Hide unused options values' fieldId='camel-form-hide-unused-option-value'>
+        <Checkbox
+          id='camel-form-hide-unused-option-value'
+          isChecked={isHideUnusedOptionValues}
+          onChange={(value, _) => onIsHideUnusedOptionValuesChange(value)}
+        />
+      </FormGroup>
+      <FormGroup label='Include trace / debug streams' fieldId='camel-form-include-trace-debug-streams'>
+        <Checkbox
+          id='camel-form-include-trace-debug-streams'
+          isChecked={isIncludeTraceDebugStreams}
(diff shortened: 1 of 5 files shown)


packages/hawtio/src/plugins/camel/camel-preferences-storage-service.ts
tadayosi, line 1: I said we should normally try to follow the impl the old Hawtio does unless there are good reasons not to, but to me this one seems to be a good case that we can depart from the old one and make it more simple. Instead of putting each parameter to a local storage item, what about merging them into a single json object called `camel.preferences` and persisting it as a whole?  This way, probably the service can be simplified with less methods, one for save the object and another for load it, and maybe a few more helpers to do some specific jobs.

The React component part can be also simplified with this.

tadayosi:
And now I realise that this is only half of what #41 should implement. We also need to implement the consumer side of the preferences, since there should be consumers in the Camel plugin that uses these parameters to tune their logics.

So, at least, the `camelPreferencesService` should provide each method for a corresponding preferences parameter that is supposed to be used from some Camel view elsewhere. In the old Hawtio, these methods were implemented here:
https://github.com/hawtio/hawtio-integration/blob/main/plugins/camel/ts/camelHelpers.ts

@phantomjinx @mmelko Are there any places in the Camel views you've implemented where the preferences parameters should be referred to but haven't been yet since the prefs hadn't been implemented yet?

tadayosi:
> - Right now values are being saved to local storage and little else. Would like to get your input in if any components already exist to put the wiring on them.

Yes, that's the latter half of the task. But it can be broad and some may be even yet to be implemented, we can work on it in a separate pull req. In this pull req, I think it's ok so long as the `camelPreferencesService` provides the getter methods for all parameters.

> - Reset button. Should I add it? Home preferences didn't have a Reset button before and it has it, same for Jolokla settings. I guess it's a yes but wanted to confirm.

No, you don't need it. As I explained before, we only need two reset buttons in the entire Hawtio console. One is to reset every setting except connections, and the other is to reset connections specifically. That's why I said `localStorage.clear()` is enough for the former reset button. Probably you're still misunderstanding something.

> Home preferences didn't have a Reset button before and it has it, same for Jolokla settings.

I don't see what it means. The old Hawtio has been having the two buttons and nothing changes at this point. One thing to note is that the old `General` and `Reset` are refactored into single `Home` prefs, and `Jolokia` and `Connect` are refactored into `Connect`, because the old prefs were rather cluttered with prefs with sparse options. 

> - Hint icons. I saw we dropped this from other preferences pages. Should I file a separate issue to implement them on all preferences pages?

Good point. Yes, please file another issue and let's tackle it separately. I agree it's good idea to restore the hints.

> - Overuse of arrow functions on the component: trying to reduce the amount of duplicated code for type checking, I had to use arrow functions because I lose the "this" when passing them as parameters. Is not too much boilerplate... until prettier comes up and puts everything in different lines :(. If I remember that the correct pattern to avoid that was to store this in a const variable in the class, but typescript seemed to not like it.

Possibly my suggestion above to consolidate parameters into a single json could mitigate the overuse of arrows here?  I think if prettier makes it look boilerplate then it's already boilerplate... keeping them in single lines is just hiding the complexity, I guess.

packages/hawtio/src/plugins/camel/CamelPreferences.tsx
```diff
@@ -0,0 +1,191 @@
...
+      value => storageService.setRouteMetricMaximumSeconds(value),
+    )
+
+  return (
+    <FormSection>
+      <FormGroup label='Hide option documentation' fieldId='camel-form-hide-option-documentation'>
+        <Checkbox
+          id='camel-form-hide-option-documentation'
+          isChecked={isHideOptionDocumentation}
+          onChange={(value, _) => onIsHideOptionDocumentationChange(value)}
```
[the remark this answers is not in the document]
joshiraez: Yes, it has a type error. The onChange function for Checkbox has a type signature of (boolean, WindowEvent), so I need to discard the second parameter or TS will complain.

packages/hawtio/src/plugins/camel/camel-preferences-storage-service.ts
joshiraez: I was torn between both approachs and I think it's as sensible to store it as an object, and maybe easier to read? The boilerplate is going to be mostly the same, as most of the boilerplate is having one specific function and state for each form, although I can play around with it and just use keys as in the current design. I'll get with this

joshiraez:
> And now I realise that this is only half of what #41 should implement. We also need to implement the consumer side of the preferences, since there should be consumers in the Camel plugin that uses these parameters to tune their logics.
> 
> So, at least, the `camelPreferencesService` should provide each method for a corresponding preferences parameter that is supposed to be used from some Camel view elsewhere. In the old Hawtio, these methods were implemented here: https://github.com/hawtio/hawtio-integration/blob/main/plugins/camel/ts/camelHelpers.ts
> 
> @phantomjinx @mmelko Are there any places in the Camel views you've implemented where the preferences parameters should be referred to but haven't been yet since the prefs hadn't been implemented yet?

I asked @phantomjinx and he didn't realize if there were any component yet that consumed these settings, but I thought it was cool to ask in the PR as well to see if anyone had any more data. I'll leave the load functions for each preference as before but I'll modify the rest of "loads" to just use an object (I feel like in this case I can do a little more clever coding to ensure that we are always loading a whole object and making sure we don't overwrite any keys, should be simple enough). They should be readily available if something has to consume it :)

joshiraez:
> > * Right now values are being saved to local storage and little else. Would like to get your input in if any components already exist to put the wiring on them.
> 
> Yes, that's the latter half of the task. But it can be broad and some may be even yet to be implemented, we can work on it in a separate pull req. In this pull req, I think it's ok so long as the `camelPreferencesService` provides the getter methods for all parameters.

I'll leave all the "load" functions as they are interfaced for now so they are available to be consumed. I'll create a separate issue after this one is merged so we can keep track this is yet to be implemented in the components.

> > * Reset button. Should I add it? Home preferences didn't have a Reset button before and it has it, same for Jolokla settings. I guess it's a yes but wanted to confirm.
> 
> No, you don't need it. As I explained before, we only need two reset buttons in the entire Hawtio console. One is to reset every setting except connections, and the other is to reset connections specifically. That's why I said `localStorage.clear()` is enough for the former reset button. Probably you're still misunderstanding something.
> 
> > Home preferences didn't have a Reset button before and it has it, same for Jolokla settings.
> 
> I don't see what it means. The old Hawtio has been having the two buttons and nothing changes at this point. One thing to note is that the old `General` and `Reset` are refactored into single `Home` prefs, and `Jolokia` and `Connect` are refactored into `Connect`, because the old prefs were rather cluttered with prefs with sparse options.

oooooooooooooooooooooh, ok. Now I get it. Nothing then, you are totally right that we fuse those 2 plus connect uses their own behaviour with the dispatch event. Thanks for the explanation.

> > * Hint icons. I saw we dropped this from other preferences pages. Should I file a separate issue to implement them on all preferences pages?
> 
> Good point. Yes, please file another issue and let's tackle it separately. I agree it's good idea to restore the hints.

Done. #195 for traceability

> > * Overuse of arrow functions on the component: trying to reduce the amount of duplicated code for type checking, I had to use arrow functions because I lose the "this" when passing them as parameters. Is not too much boilerplate... until prettier comes up and puts everything in different lines :(. If I remember that the correct pattern to avoid that was to store this in a const variable in the class, but typescript seemed to not like it.
> 
> Possibly my suggestion above to consolidate parameters into a single json could mitigate the overuse of arrows here? I think if prettier makes it look boilerplate then it's already boilerplate... keeping them in single lines is just hiding the complexity, I guess.

It "should" in some cases. I'm going to be a bit more clever and just have a single function for the update of each input parameter instead of having them one by one. Most of the boilerplate comes from it actually, even though is configuration code.

Btw, the issue about using arrow functions on the onDidSomethingChange functions and the issue about using arrow functions on the Checkbox components are different.

The onDidSomethingChange functions will overwrite the this if I pass the function as argument :( (rather, the this is always the one on the environment the function is created. Using an arrow function ensures that the this is the one of the context where the arrow function was created) which is anoying, but is a JS thing.

The checkbox one is about checkbox onChange prop having a different type signature than other inputs.


Thank you for your input @tadayosi  I'll be refinining this during the following hours

joshiraez:
Also, something that I forgot to add yesterday pertaining to #173 . This PR doesn't has the bug where the input gets unfocused every time I type. I'll be focusing on finishing this PR by now but I'll keep an eye and try to extract why in this case it works and not in the other component (when they are essentially the same as far as I see)

packages/hawtio/src/plugins/camel/camel-preferences-storage-service.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  private loadFromStorage<T>(storageKey: string): T {
+    const localStorageValue = localStorage.getItem(storageKey)
+
+    if (!localStorageValue) return PREFERENCES_DEFAULT_VALUES[storageKey] as T
+    return JSON.parse(localStorageValue) as T
+  }
+}
+
+export const storageService = new CamelPreferencesStorageService()
```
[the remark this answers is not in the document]
joshiraez: Addressed all of the above at bd2b269

joshiraez:
@tadayosi Refactored the class to be stored in an object. I optimized to reduce boilerplate and with the `keyof`operator I was able to ensure type safety as well so I'm very happy.

There are some places that might find a little unorthodox but it should make the code more legible and avoid the overuse of arrow functions. Tell me your thoughts :)

packages/hawtio/src/plugins/camel/CamelPreferences.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const [camelPreferences, setCamelPreferences] = useState(camelPreferencesService.loadCamelPreferences())
+
+  const updatePreferences = (value: boolean | number, key: keyof ICamelPreferences): void => {
+    const updatedPreferences = { ...camelPreferences, ...{ [key]: value } }
+
+    camelPreferencesService.saveCamelPreferences(updatedPreferences)
+    setCamelPreferences(updatedPreferences)
+  }
+
+  const updateNumberValueFor = (key: keyof ICamelPreferences): ((value: string) => void) => {
```
joshiraez, line 23: This might seem unorthodox but allows to not have to put arrow functions below and just hide this bit of configuration code. I don't know if it might be too complex of a concept, I'd like your input on this.

packages/hawtio/src/plugins/camel/CamelPreferences.tsx
```diff
@@ -0,0 +1,111 @@
...
+    }
+  }
+
+  return (
+    <FormSection>
+      <FormGroup label='Hide option documentation' fieldId='camel-form-hide-option-documentation'>
+        <Checkbox
+          id='camel-form-hide-option-documentation'
+          isChecked={camelPreferences.isHideOptionDocumentation}
+          onChange={updateCheckboxValueFor('isHideOptionDocumentation')}
```
joshiraez, line 47: They keyof operator makes these "magic strings" be statically typed. If you were to put something like 'a' or 'notAKeyHahahahahaha' Ts will throw a compiler error so they are not really magic strings

packages/hawtio/src/plugins/camel/camel-preferences-service.ts
```diff
@@ -0,0 +1,95 @@
...
+}
+
+export const STORAGE_KEY_CAMEL_PREFERENCES = 'camel.preferences'
+
+class CamelPreferencesService implements ICamelStorage {
+  loadCamelPreferences(): ICamelPreferences {
+    return { ...CAMEL_PREFERENCES_DEFAULT_VALUES, ...this.loadFromStorage() }
+  }
+
+  saveCamelPreferences(newValues: Partial<ICamelPreferences>): void {
```
joshiraez, line 46: Deconstruction + Partial type makes for a very easy way to ensure that we don't erase any unset keys.

packages/hawtio/src/plugins/camel/CamelPreferences.tsx
```diff
@@ -0,0 +1,111 @@
...
+      <CamelPreferencesForm />
+    </Form>
+  </CardBody>
+)
+
+const CamelPreferencesForm: React.FunctionComponent = () => {
+  const [camelPreferences, setCamelPreferences] = useState(camelPreferencesService.loadCamelPreferences())
+
+  const updatePreferences = (value: boolean | number, key: keyof ICamelPreferences): void => {
+    const updatedPreferences = { ...camelPreferences, ...{ [key]: value } }
```
joshiraez, line 17: Preferring this method to camelPreferences[key] = value to ensure there is a new pointer for every update, so useState updates properly

packages/hawtio/src/plugins/camel/camel-preferences-service.ts
tadayosi, line 1: so far, you haven't wrote any tests yet. how about writing a test for the service with jest?  local storage can be use with jest tests.

packages/hawtio/src/plugins/camel/camel-preferences-service.ts
```diff
@@ -0,0 +1,95 @@
...
+    const preferencesToSave = { ...this.loadFromStorage(), ...newValues }
+
+    localStorage.setItem(STORAGE_KEY_CAMEL_PREFERENCES, JSON.stringify(preferencesToSave))
+  }
+
+  loadIsHideOptionDocumentation(): boolean {
```
tadayosi, line 52: I'm not sure if we really need these `load...` methods at this moment. To me it seems sufficient with having one `loadCamelOptions()` for consumers.

packages/hawtio/src/plugins/camel/index.ts
```diff
@@ -3,9 +3,9 @@ import { hawtio } from '@hawtiosrc/core'
 import { treeProcessorRegistry, workspace } from '@hawtiosrc/plugins/shared'
 import { jmxDomain, pluginPath } from './globals'
 import { camelTreeProcessor } from './tree-processor'
-// import { preferencesRegistry } from '@hawtiosrc/preferences/registry'
+import { preferencesRegistry } from '@hawtiosrc/preferences/registry'
 import { Camel } from './Camel'
-// import { CamelPreferences } from './CamelPreferences'
+import { CamelPreferences } from './CamelPreferences'
 // import help from './help.md'
 // import { jolokiaService } from './jolokia-service'
```
tadayosi, line 10: once it's done, please also clean up this unused comment line.

packages/hawtio/src/plugins/camel/camel-preferences-service.ts
```diff
@@ -0,0 +1,95 @@
+export interface ICamelStorage {
```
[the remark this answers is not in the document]
joshiraez: I agree, I'm changing it now

packages/hawtio/src/plugins/camel/camel-preferences-service.ts
```diff
@@ -0,0 +1,95 @@
...
+  loadMaximumTraceDebugBodyLength(): number
+  loadMaximumLabelWidth(): number
+  loadIsIgnoreIDForLabel(): boolean
+  loadIsShowInflightCounter(): boolean
+  loadRouteMetricMaximumSeconds(): number
+  loadCamelPreferences(): ICamelPreferences
+  saveCamelPreferences(newValues: Partial<ICamelPreferences>): void
+}
+
+export interface ICamelPreferences {
```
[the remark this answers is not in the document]
joshiraez: You have a point and I wholefully agree. I'll change to a type and rename it

packages/hawtio/src/plugins/camel/camel-preferences-service.ts
joshiraez: I'll be on it today :)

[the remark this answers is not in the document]
joshiraez: I'm glad you liked it :smile: Thanks!

joshiraez:
@tadayosi Everything has been addressed. Feel free to close the conversations when you check everything is ok :)

tadayosi (approved):

Issue #194, opened 2023-03-23
Implement the workspace monitoring of JMX tree and plugins


phantomjinx:
The workspace uses the hawtio:TreeWatcher and hawtio:Registry mbeans to monitor changes that may occur to the target app, eg. a request is sent to create an endpoint and the endpoint is actually created. The callbacks in the hawtio-next app need to be implemented to allow the camel and jmx trees to be refreshed in response to such changes.

phantomjinx:
See https://github.com/hawtio/hawtio-integration/blob/main/plugins/jmx/ts/workspace.ts#L154

PR #196, opened 2023-03-23
(#194): Implements workspace monitoring of tree server changes

phantomjinx:
* workspace.ts
  * Ports the functions for the Registry and TreeWatcher jolokia mbeans
  * The response for the TreeWatcher is to call workspace.refresh()
  * Allows new nodes created by async calls on target apps to be added to the JMX and Camel trees

* .../context.ts
  * Adds a react ref for holding onto the selected node between renderings which allows the existing selected node to be re-selected upon a change to the underlying nodes trees
  * Since both camel and JMX trees use the same instances of nodes (taken from workspace.getTree() then they can be directly selected in both trees
  * Ensure the trees expand to show the node selections if the one tree happens to have not ever selected the selection before, eg. select a context in Camel Plugin then click on JMX Plugin would expand and highlight the same context in the JMX Plugin tree

* [Camel|Jmx]TreeView
  * Adds a useEffect() dependent on a change to the tree to redisplay the treeview - important for workspace refreshing after a TreeWatcher report

* JmxTreeView
  * Enhances the tree expansion logic in the same way as the CamelTreeView to either include or not include the allExpanded prop. If not included then the node.defaultExpanded properties are observed instead

* node.ts
  * path() method for extracting the string path [] of the node -> root
  * forEach() method for calling a callback on a path of nodes

* tree.ts
  * Adds navigate and forEach methods to mirror those used in node.ts

### DIFF
--- a/packages/hawtio/src/plugins/camel/context.ts
+++ b/packages/hawtio/src/plugins/camel/context.ts
@@ -1,4 +1,4 @@
-import { createContext, useEffect, useState, useContext } from 'react'
+import { createContext, useEffect, useState, useContext, useRef } from 'react'
 import { TreeViewDataItem } from '@patternfly/react-core'
 import { PluginNodeSelectionContext } from '@hawtiosrc/plugins'
 import { workspace, MBeanNode, MBeanTree } from '@hawtiosrc/plugins/shared'
@@ -15,22 +15,42 @@ export function useCamelTree() {
   const { selectedNode, setSelectedNode } = useContext(PluginNodeSelectionContext)
   const navigate = useNavigate()
 
+  /*
+   * Need to preserve the selected node between re-renders since the
+   * populateTree function called via the refresh listener does not
+   * cache the value and stores it as null
+   */
+  const refSelectedNode = useRef<MBeanNode | null>()
+  refSelectedNode.current = selectedNode
+
   const populateTree = async () => {
     const wkspTree: MBeanTree = await workspace.getTree()
     const rootNode = wkspTree.findDescendant(node => node.name === jmxDomain)
     if (rootNode) {
+      /*
+       * Using the camel domain node from the original tree means it is the same
+       * node as that that appears in the workspace tree
+       */
       const subTree: MBeanTree = MBeanTree.createFromNodes(pluginName, [rootNode])
       setTree(subTree)
       if (rootNode && rootNode.children && rootNode.children.length > 0) {
-        const r: TreeViewDataItem = rootNode as TreeViewDataItem
-        const c: TreeViewDataItem = rootNode.children[0] as TreeViewDataItem
-        c.defaultExpanded = true
-        r.defaultExpanded = true
-        if (selectedNode) {
-          setSelectedNode(selectedNode)
+        const path: string[] = []
+        if (refSelectedNode.current) {
+          path.push(...refSelectedNode.current.path())
         } else {
-          setSelectedNode(rootNode.children[0])
+          path.push(...rootNode.getChildren()[0].path())
         }
+
+        // Expand the nodes to redisplay the path
+        rootNode.forEach(path, (node: MBeanNode) => {
+          const tvd = node as TreeViewDataItem
+          tvd.defaultExpanded = true
+        })
+
+        // Ensure the new version of the selected node is selected
+        const newSelected = rootNode.navigate(...path)
+        if (newSelected) setSelectedNode(newSelected)
+
         /* On population of tree, ensure the url path is returned to the base plugin path */
         navigate(pluginPath)
       }
@@ -52,7 +72,6 @@ export function useCamelTree() {
     }
 
     const listener = () => {
-      setSelectedNode(null)
       setLoaded(false)
       loadTree()
     }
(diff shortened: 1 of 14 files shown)


packages/hawtio/src/util/strings.ts
```diff
@@ -93,3 +93,8 @@ export function stringSorter(a: string, b: string): number {
   }
   return 0
 }
+
+export function parseBoolean(value: string) {
```
tadayosi, line 97: unless you declare return type, shouldn't it become any? then it could do more than expected, which should be avoided.

packages/hawtio/src/plugins/camel/context.ts
```diff
@@ -51,10 +71,9 @@ export function useCamelTree() {
       setLoaded(true)
     }
 
-    const listener = () => {
-      setSelectedNode(null)
+    const listener = async () => {
       setLoaded(false)
-      loadTree()
+      await loadTree()
```
tadayosi, line 76: `EventListener` is not `() => Promise<void>` so it doesn't take any effect to use async/await here. You can simply call `loadTree()` without async.

packages/hawtio/src/plugins/shared/workspace.ts
```diff
@@ -46,6 +54,68 @@ class Workspace {
     })
   }
 
+  private maybeUpdatePlugins(response: IResponse): void {
+    if (this.pluginUpdateCounter === null) {
+      this.pluginUpdateCounter = response.value as number | null
+      return
+    }
+    if (this.pluginUpdateCounter !== response.value) {
+      if (parseBoolean(localStorage['autoRefresh'])) {
```
tadayosi, line 63: we shouldn't directly touch localStorage here. like `updateRate` this parameter should be managed by `jolokiaService`. lets create save/load methods on jolokiaService.

like `updateRate` this parameters should be also able to change at the Jolokia preferences pane.

packages/hawtio/src/plugins/shared/workspace.ts
```diff
@@ -46,6 +54,68 @@ class Workspace {
...
+      this.treeWatcherCounter = counter as number | null
+      return
+    }
+    if (this.treeWatcherCounter !== counter) {
+      this.treeWatcherCounter = counter as number | null
+      this.refreshTree()
+    }
+  }
+
+  async maybeMonitorPlugins() {
```
tadayosi, line 81: This method name doesn't tell what it really does. What about splitting the method into two: `maybeMonitorPlugins()` and `maybeMonitorTree()`?

packages/hawtio/src/plugins/shared/workspace.ts
```diff
@@ -46,6 +54,68 @@ class Workspace {
...
+        this.pluginRegisterHandle = await jolokiaService.register(
+          {
+            type: 'read',
+            mbean: HAWTIO_TREE_WATCHER_MBEAN,
+            attribute: 'Counter',
+          },
+          (response: IResponse) => this.maybeReloadTree(response),
+        )
+      }
+    }
```
tadayosi, line 116: Why doesn't the tree watcher have the same unregistering branch with the plugin registry?  I see the old impl doesn't as well, but is there a good reason not to do so?

packages/hawtio/src/util/strings.ts
joshiraez: I don't know if I understood your message correctly. If you meant that by not putting `: boolean` there it might return anything (like an object) yes, you are right, and this should have return type of `boolean` on my opinion.

Setting types also allows TS to do type inference correctly. It might require a `as boolean` in the JSON.parse to conform to type checkin tho

packages/hawtio/src/plugins/shared/workspace.ts
```diff
@@ -13,6 +13,10 @@ export type MBeanCache = { [propertyList: string]: string }
 
 class Workspace {
   private tree: Promise<MBeanTree>
+  private pluginRegisterHandle: number | null = null
```
joshiraez, line 16: To avoid the overuse of number | null one can put `private pluginRegisterHandle? : number` and just check if it's set or not. Seeing as you are using null as a flag value, you can use undefined as well and use isSet() later

packages/hawtio/src/util/strings.ts
phantomjinx: Refactored and tests added

packages/hawtio/src/plugins/shared/workspace.ts
phantomjinx: Noted and refactored. Thanks.
Cannot use `isSet` from `util/types` due to an error in webpack as described [here](https://bobbyhadz.com/blog/module-not-found-error-cant-resolve-assert#module-not-found-error-cant-resolve-util). However, could not get the workaround config to solve it. Therefore, backed out for the moment

packages/hawtio/src/util/strings.ts
tadayosi: What I expected is defining return type explicitly like this:
```
export function parseBoolean(value: string): boolean {
```
The current impl is clear that it indeed returns boolean, but still the func signature relies on type inference. I expect that every exported function, const, type, class, etc. has clear type definitions unless it doesn't return anything (`void`). It's ok to omit them for private ones for simplicitly, because the scope is narrow so less chances of confusion.

phantomjinx: Occam's razor.
In this case, I updated the fn signature
Then thought I would no longer need it. 
Then realised I would and so `git checkout -f string.ts` to bring it back.
Then changed the body but not the signature.

Anyway, good catch and signature no updated.

tadayosi (approved):

PR #210, opened 2023-03-29
feat: Add base for the route diagram

mmelko:
This is a WIP for the route diagram based on reactflow+dagre.
What's missing: 
 - fixed styling, currently styles are all over the place, that needs to be seriously polished 
 - attributes for each node 
 
 current screenshot: 
Screenshot 2023-03-29 at 16 33 39


packages/hawtio/src/plugins/camel/routes/RouteDiagram.tsx
tadayosi, line 1: Actually, Routes and Route Diagram are two different things in terms of feature. I don't think they share a lot. What about creating another directory `route-diagram` and putting the related code under it instead of putting them under `routes`?

packages/hawtio/src/plugins/camel/routes/RouteDiagram.tsx
```diff
@@ -0,0 +1,127 @@
...
+import './routeDiagram.css'
+import { Flex, FlexItem, Page } from '@patternfly/react-core'
+
+const dagreGraph = new dagre.graphlib.Graph()
+dagreGraph.setDefaultEdgeLabel(() => ({}))
+
+const nodeWidth = 300
+const nodeHeight = 86
+
+export type ICamelNode = Node & {
```
tadayosi, line 27: You don't need `I` prefix. Non only because it's not an interface, but also because we generally wouldn't use `I` prefixes other than service interfaces.

packages/hawtio/src/plugins/camel/routes/RouteDiagram.tsx
```diff
@@ -0,0 +1,127 @@
...
+      x: nodeWithPosition.x - nodeWidth / 2,
+      y: nodeWithPosition.y - nodeHeight / 2,
+    }
+
+    return node
+  })
+
+  return { nodes, edges }
+}
+const CamelNode = ({ data }: NodeProps<CamelNodeData>) => {
```
tadayosi, line 61: For the top-level React components, I'd like to declare type `React.FunctionComponent` so that readers can easily find they are React components .

packages/hawtio/src/plugins/camel/routes/visualization-service.ts
```diff
@@ -0,0 +1,220 @@
...
+      ) {
+        siblingNodes = siblings
+      } else {
+        siblingNodes = [nodes.length - 1]
+      }
+    }
+    return siblingNodes
+  }
+}
+export const routeVizualisationService = new VisualizationService()
```
tadayosi, line 220: `routeVizualisationService` -> `routeVisualizationService` :-)
Or you can maybe simply name `visualizationService`?

packages/hawtio/src/plugins/camel/routes/RouteDiagram.tsx
```diff
@@ -0,0 +1,127 @@
...
+dagreGraph.setDefaultEdgeLabel(() => ({}))
+
+const nodeWidth = 300
+const nodeHeight = 86
+
+export type ICamelNode = Node & {
+  data: CamelNodeData
+}
+
+const getLayoutedElements = (nodes: ICamelNode[], edges: Edge[], direction = 'TB') => {
```
tadayosi, line 31: Maybe you can put those dagre specific code, i.e. this function and const `dagreGraph`, into another file or move them into `visualizationService` if that makes sense, in order to keep the React components simple.

packages/hawtio/src/plugins/camel/routes/visualization-service.ts
tadayosi, line 1: Is it hard to write tests for the service?

packages/hawtio/src/plugins/camel/CamelContent.tsx
```diff
@@ -78,6 +79,15 @@ export const CamelContent: React.FunctionComponent = () => {
         !ccs.isEndpointsFolder(node) &&
         (ccs.isRouteNode(node) || ccs.isRoutesFolder(node)),
     },
+    {
+      id: 'routeDiagram',
+      title: 'Route Diagram',
+      component: RouteDiagram,
+      isApplicable: (node: MBeanNode) =>
+        !ccs.isEndpointNode(node) &&
```
phantomjinx, line 87: Checking it is not an Endpoint Node or Folder is probably now unnecessary given they are mutually exclusive to Route Node and Folder.

phantomjinx (commented):
LGTM so far

packages/hawtio/src/plugins/camel/routes/visualization-service.ts
mmelko: I don't think so, I'm planning to add some tests for sure.

packages/hawtio/src/plugins/camel/routes/RouteDiagram.tsx
mmelko: Sure, NP .. I've put it in the same directory because it's all in the page under routes but I will move it.

mmelko:
@tadayosi I think I fixed all of the issues, also updated a style to look more representative (Still some issues though)

Screenshot 2023-03-31 at 16 12 10
I also added simple tests that validates parsing from the service. I think we can merge it in this state if you agree. 

Next I will implement route statistics (on hover and in the right top corner).

packages/hawtio/src/plugins/camel/route-diagram/RouteDiagram.tsx
```diff
@@ -0,0 +1,89 @@
...
+  useNodesState,
+} from 'reactflow'
+import { CamelNodeData, visualizationService } from '@hawtiosrc/plugins/camel/route-diagram/visualization-service'
+import 'reactflow/dist/style.css'
+import './routeDiagram.css'
+import { Page } from '@patternfly/react-core'
+
+// export type ICamelNode = Node & {
+//   data: CamelNodeData
+// }
```
tadayosi, line 22: Can we remove this comment?

tadayosi (approved):

joshiraez (approved):

packages/hawtio/src/plugins/camel/CamelContent.tsx
mmelko: I deleted that check.

phantomjinx (approved):

Issue #32, opened 2023-01-26
Core - Define plugin interface and mechanism


tadayosi:
The plugin interface and mechanism have already been implemented for the Connect and JMX plugins, but we need to further polish and stabilise them with new built-in plugins such as Camel, etc. and then expose the interface for external plugin developers, including Artemis plugin.

What we need to complete on the design and implementation of Hawtio.next plugin are:

- How to provide the JMX MBean tree to other plugins
- How to allow other plugins to post-process the JMX tree
- How to load external plugins dynamically at runtime

This also includes the following sub tasks:

- [x] https://github.com/hawtio/hawtio-next/issues/33
- ~https://github.com/hawtio/hawtio-next/issues/34~

tadayosi:
The plugin mechanism has been implemented using Webpack Module Federation. Also, examples are provided. Pushing the documentation task out of the scope of this ticket and close it. Documentation will be addressed separately later.

PR #213, opened 2023-03-30
feat(core,ui,app): implement plugin mechanism based on module federation

tadayosi:
Relates #32

This completes the implementation of the new plugin mechanism based on module federation. This works well in development mode, but I observe a plugin loading issue at the main Hawtio project. I'll keep it on hold until I'm sure it's resolvable at the main Hawtio project side before merging.


app/craco.config.js
```diff
@@ -1,11 +1,51 @@
...
+              requiredVersion: dependencies['react-dom'],
+            },
+            'react-router-dom': {
+              singleton: true,
+              requiredVersion: dependencies['react-router-dom'],
+            },
+            '@hawtio/react': {
+              singleton: true,
+              // Hardcoding needed because it cannot handle yarn 'workspace:*' version
+              requiredVersion: '^0.2.0-dev.4',
```
phantomjinx, line 33: Need to make a note of that somewhere to ensure we keep the versions in sync when we increment after release.

packages/hawtio/src/core/event-service.ts
```diff
@@ -49,30 +58,38 @@ class EventService implements IEventService {
...
+  refresh() {
     this.eventEmitter.emit(EVENT_REFRESH)
   }
 
-  onRefresh(listener: EventListener): void {
+  onRefresh(listener: EventListener) {
     this.eventEmitter.on(EVENT_REFRESH, listener)
   }
 
+  pluginsUpdated() {
```
phantomjinx, line 85: With the introduction of this event listener and response, it maybe useful to now update the `workspace` response to `pluginsUpdated` [here](https://github.com/hawtio/hawtio-next/blob/main/packages/hawtio/src/plugins/shared/workspace.ts#L66) as a `window.reload()` is going to make all other responses to `pluginsUpdated` pointless, I'd have thought?

phantomjinx (approved):
Looks good. Couple of follow-on considerations for us to think about.

app/craco.config.js
tadayosi: Actually, `^0.2.0-dev.4` means all 0.x versions higher than or equal to `0.2.0-dev.4` are compatible, so we won't need to update this so frequently. Adding to that, this is the configuration for the "app" package, which is only for development. So I don't think we need to be so careful about updating it.

On the other hand, updating https://github.com/hawtio/hawtio/blob/3.x/console/craco.config.js would be important as it's part of the actual distribution. But thankfully, the configuration there should look like this, so no need for updates:
```
            '@hawtio/react': {
              singleton: true,
              requiredVersion: dependencies['@hawtio/react'],
            },
```

joshiraez (approved):
Looks great

packages/hawtio/src/core/event-service.ts
tadayosi: Good catch. Yes, that's also what I had in mind a bit as well. We'd need to review consistency between the new plugin system and the watcher impl we have in the workspace.

In reality, probably that plugin watcher MBean is an old mechanism, where in the past we expected dynamic additions and deletions of plugins via JMX, especially on dynamic environments like OSGi/Karaf. But now, with Hawtio.next we currently only support Spring Boot and Quarkus as the runtimes and they are relatively static at runtime; all dependencies would be resolved before bootstrap. I can't think of use cases where plugins are added dynamically after application launches on Spring Boot and Quarkus. Correct me if I'm wrong.

This plugin implementation here only loads external plugins at the initial page loading time, and it might be enough for the runtimes Hawtio.next supports. If that's the case, we can consider just removing the plugin watcher from hawtio-next and the main hawtio project.

app/craco.config.js
tadayosi: Let me take it as resolved

packages/hawtio/src/core/event-service.ts
tadayosi: Filed follow up issue #214

Issue #125, opened 2023-03-01
JMX - Refactor the Attributes Component to include tabular form of multiple beans


phantomjinx:
The old Attributes Component includes the ability to display the children of a group category, eg. Contexts, as a table with the bean attributes as the columns and mbeans as the rows:

![jmx-contexts-screenshot-old](https://user-images.githubusercontent.com/1634180/222115965-baa5cf5b-6dbd-45f8-a56f-ac0e1f2bceed.png)

Old Attributes component for camel contexts ^

tadayosi:
@joshiraez You can work on this too.

joshiraez:
I'm picking this up then and using @phantomjinx approach in camel as a base. I'll do it after the current camel help page

joshiraez:
I spoke with Paul and I'll be picking this up to continue the work on the JMX plugin. From the mentioned Camel issue for the attributes view it seems they should be similar so I'll tr to take that as a base

joshiraez:
@tadayosi Just to be sure, this extends #10 right? The chart component should be used for this

Issue #190, opened 2023-03-22
Add E2E tests for `yarn start` at `app` package


tadayosi:
Sometimes dependabot upgrades break `yarn start` (e.g. #179), but currently there are no CI workflows to catch those errors in advance. We should add some basic E2E tests that ensures `yarn start` at `app` package doesn't break.

Is React Testing Library provide sufficient for providing such E2E tests?  Or should we consider introducing an E2E framework like [Cypress](https://www.cypress.io/)?

mmuzikar:
Done with #322

PR #215, opened 2023-04-03
Tabular attribute table

joshiraez:
Implements #125 

image

A few caveats after checking both versions.

* Added also the Name table that appears on the parent nodes when not all children nodes are mbeans
* There is some slight differences right now where some nodes would have the Object Name view instead of the normal one but they seem to follow a set of defined columns set for each component. It seemed way out of scope and for now I settled for this version filtering by conditions.


packages/hawtio/src/plugins/jmx/JmxContent.tsx
```diff
@@ -84,7 +87,17 @@ export const JmxContent: React.FunctionComponent = () => {
...
+          <AttributeTable />
+        )}
+        {navItems.length === 0 &&
+          !selectedNode.objectName &&
+          !mBeanCollectionApplicable(selectedNode) &&
+          !selectedNode.parent && <NodeNameTable />}
+        {navItems.length === 0 &&
+          !selectedNode.objectName &&
+          !mBeanCollectionApplicable(selectedNode) &&
+          selectedNode.parent && <JmxContentMBeans />}
```
tadayosi, line 100: The conditionals are getting complex. I think React fragment is not a good place to embed complex conditionals like this. Maybe it's better to put them outside of the fragment and make it clearer conditionals to choose a React element.

tadayosi, line 100: Also, I don't still understand why we need these three components `JmxContentMBeans`, `AttributeTable`, and `NodeNameTable`. Can't `JmxContentMBeans` be just implemented as `AttributeTable`?  What was the legacy Hawtio's behaviour when it had mbeans and folders at the same time?

packages/hawtio/src/plugins/shared/attributes/AttributeTable.tsx
```diff
@@ -0,0 +1,76 @@
...
+    return (
+      <Card>
+        <CardBody>
+          <Text component='p'>Reading attributes...</Text>
+        </CardBody>
+      </Card>
+    )
+  }
+
+  const tidyLabels = (str: string) =>
```
tadayosi, line 51: This looks ad hoc. In the legacy Hawtio, we created a dedicated util function to share the common logic. Also write tests for the function.
https://github.com/hawtio/hawtio-core/blob/main/src/app/utilities/baseHelpers.ts#L479-L494

packages/hawtio/src/plugins/jmx/JmxContent.tsx
joshiraez: In my opinion, doing so would just make the code far, far more unreadable. Look at the amalgamation that handles the rendering. It takes over 140 lines, mixes different use cases, and doesn't even have all the logic at all (as you need the defined COLUMN_DEFS to get some of the logic out, and a lot of those if-elses are just plain different use cases mixed together. In comparison, this only spans 12 and every use case is clearly defined in its own component.

This way, the filtering order is really clear in how a component is chosen, and every component has its own separate logic to handle the nodes.

```    function render(response: { request: any, value: any }): void {
      let data = response.value;
      let mbeanIndex = $scope.mbeanIndex;
      let mbean = response.request['mbean'];

      if (mbean) {
        // lets store the mbean in the row for later
        data["_id"] = mbean;
      }
      if (mbeanIndex) {
        if (mbean) {

          let idx = mbeanIndex[mbean];
          if (!angular.isDefined(idx)) {
            idx = $scope.mbeanRowCounter;
            mbeanIndex[mbean] = idx;
            $scope.mbeanRowCounter += 1;
          }
          if (idx === 0) {
            // this is to force the table to repaint
            $scope.selectedIndices = $scope.selectedItems.map((item) => $scope.gridData.indexOf(item));
            gridData = [];

            if (!$scope.columnDefs.length) {
              // lets update the column definitions based on any configured defaults

              let key = workspace.selectionConfigKey();
              let defaultDefs = _.clone(workspace.attributeColumnDefs[key]) || [];
              let defaultSize = defaultDefs.length;
              let map = {};
              angular.forEach(defaultDefs, (value, key) => {
                let field = value.field;
                if (field) {
                  map[field] = value
                }
              });

              let extraDefs = [];
              _.forEach(data, (value, key) => {
                if (includePropertyValue(key, value)) {
                  if (!map[key]) {
                    extraDefs.push({
                      field: key,
                      displayName: key === '_id' ? 'Object name' : Core.humanizeValue(key),
                      visible: defaultSize === 0
                    });
                  }
                }
              });

              // the additional columns (which are not pre-configured), should be sorted
              // so the column menu has a nice sorted list instead of random ordering
              extraDefs = extraDefs.sort((def, def2) => {
                // make sure _id is last
                if (_.startsWith(def.field, '_')) {
                  return 1;
                } else if (_.startsWith(def2.field, '_')) {
                  return -1;
                }
                return def.field.localeCompare(def2.field);
              });
              extraDefs.forEach(e => defaultDefs.push(e));

              if (extraDefs.length > 0) {
                $scope.hasExtraColumns = true;
              }

              $scope.columnDefs = defaultDefs;
            }
          }
          // mask attribute read error
          _.forEach(data, (value, key) => {
            if (includePropertyValue(key, value)) {
              data[key] = maskReadError(value);
            }
          });
          // assume 1 row of data per mbean
          gridData[idx] = data;

          let count = $scope.mbeanCount;
          if (!count || idx + 1 >= count) {
            // only cause a refresh on the last row
            let newSelections = $scope.selectedIndices.map((idx) => $scope.gridData[idx]).filter((row) => row);
            $scope.selectedItems.splice(0, $scope.selectedItems.length);
            $scope.selectedItems.push.apply($scope.selectedItems, newSelections);
            $scope.gridData = gridData;
            Core.$apply($scope);
          }
          // if the last row, then fire an event
        } else {
          log.info("No mbean name in request", JSON.stringify(response.request));
        }
      } else {
        $scope.columnDefs = PROPERTIES_COLUMN_DEFS;
        let showAllAttributes = true;
        if (_.isObject(data)) {
          let properties = [];
          _.forEach(data, (value, key) => {
            if (showAllAttributes || includePropertyValue(key, value)) {
              // always skip keys which start with _
              if (!_.startsWith(key, "_")) {
                // lets format the ObjectName nicely dealing with objects with
                // nested object names or arrays of object names
                if (key === "ObjectName") {
                  value = unwrapObjectName(value);
                }
                // lets unwrap any arrays of object names
                if (_.isArray(value)) {
                  value = value.map((v) => unwrapObjectName(v));
                }
                // the value must be string as the sorting/filtering of the table relies on that
                let type = lookupAttributeType(key);
                let data = {
                  key: key,
                  name: Core.humanizeValue(key),
                  value: maskReadError(Core.safeNullAsString(value, type))
                };

                generateSummaryAndDetail(key, data);
                properties.push(data);
              }
            }
          });
          if (!_.some(properties, (p) => {
            return p['key'] === 'ObjectName';
          })) {
            let objectName = {
              key: "ObjectName",
              name: "Object Name",
              value: mbean
            };
            generateSummaryAndDetail(objectName.key, objectName);
            properties.push(objectName);
          }
          properties = _.sortBy(properties, 'name');
          $scope.selectedItems = [data];
          data = properties;
        }
        $scope.gridData = data;
        Core.$apply($scope);
      }
    }
    ```

File: packages/hawtio/src/plugins/shared/attributes/AttributeTable.tsx (hunk 2 above)
Comment (joshiraez) [reply packages/hawtio/src/plugins/shared/attributes/AttributeTable.tsx:51]: A bit, I saw the old util function but because it was only used here it felt best to leave it here. I will move it to an util function and write tests.

File: packages/hawtio/src/plugins/jmx/JmxContent.tsx (hunk 1 above)
Comment (joshiraez) [reply packages/hawtio/src/plugins/jmx/JmxContent.tsx:100]: I think I can do that

File: packages/hawtio/src/plugins/jmx/JmxContent.tsx [hunk 3]
```diff
@@ -40,6 +42,7 @@ export const JmxContent: React.FunctionComponent = () => {
   }
 
   const mBeanApplicable = (node: MBeanNode) => node.objectName
+  const mBeanCollectionApplicable = (node: MBeanNode) => node?.children?.every(child => child.objectName)
```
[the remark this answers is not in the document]
joshiraez: Fixed

packages/hawtio/src/plugins/jmx/JmxContent.tsx
joshiraez: Fixed

tadayosi: I don't think you answered my question. I wasn't asking about how the legacy hawtio is implemented, but asking about what it behaves in the same cases. What does each of JmxContentMBeans, AttributeTable, and NodeNameTable do?  Why do we need JmxContentMBeans and AttributeTable at the same time?  What was the legacy Hawtio's behaviour when it had mbeans and folders at the same time?

joshiraez:
On top of that each table is a different view

Attributes show a list of attribute-valur pair with the mbean node
attributes

AttributesTable show a list of the attributes for every mbean in a row that
are children of the node.

NodeName is used when not all nodes below are mbeans. Seeing as this
behaviour only appeared in parent nodes i went with that logic as any one
that was not a parent node usually had the JMXContentMBean default one

JMXContentMBean is the default view, which only shows the objectNames

packages/hawtio/src/plugins/shared/attributes/AttributeTable.tsx
joshiraez: Yes, I did, and had to reimplement it because it failed in a few corner cases from my testing

We would need another one for snake-case too, but that should be more simple

packages/hawtio/src/plugins/jmx/JmxContent.tsx
joshiraez: Each node had configured a view in a per node basis from what I gathered (for example some folders like the topmost jmx one in the tree view had the tabular, while the others had the node names or just the default with objectName)

I tried to find a common logic that decided how each one was configured (specially because we are missing some nodes like the ones in camel) and from what I have been testing they match for almost all cases

joshiraez: On top of that each table is a different view

Attributes show a list of attribute-valur pair with the mbean node attributes

AttributesTable show a list of the attributes for every mbean in a row that are children of the node.

NodeName is used when not all nodes below are mbeans. Seeing as this behaviour only appeared in parent nodes i went with that logic as any one that was not a parent node usually had the JMXContentMBean default one. It shows the name of the children nodes

JMXContentMBean is the default view, which only shows the objectNames

I tried to be as close to what the other version was showing

packages/hawtio/src/plugins/jmx/JmxContent.tsx
```diff
@@ -21,6 +21,8 @@ import { Chart } from '@hawtiosrc/plugins/shared/chart'
 import { Operations } from '@hawtiosrc/plugins/shared/operations'
 import { Attributes } from '@hawtiosrc/plugins/shared/attributes'
 import { JmxContentMBeans, MBeanNode } from '@hawtiosrc/plugins/shared'
+import { AttributeTable } from '../shared/attributes/AttributeTable'
+import { NodeNameTable } from '../shared/attributes/NodeNameTable'
```
tadayosi, line 25: please avoid using relative paths for importing components outside of its own module. Use `@hawtiosrc` to refer to `shared` module components.

packages/hawtio/src/plugins/shared/util/helpers.ts
tadayosi, line 1: we put generic util helpers to top-level `./util`. you can define the function in `./util/strings.ts`. Also, for helper functions we use `function` declarations rather than `const` function objects.

packages/hawtio/src/plugins/shared/util/helpers.ts
```diff
@@ -0,0 +1,14 @@
...
+ * @returns The property with its proper spaces
+ */
+export const tidyLabels = (str: string) =>
+  str
+    .replace(/([a-z])([A-Z])/g, '$1 $2')
+    .replace(/\b([A-Z]+)([A-Z])([a-z])/, '$1 $2$3')
+    .replace('M Bean', 'MBean')
+    .replace(/^./, function (str) {
+      return str.toUpperCase()
+    })
```
tadayosi, line 14: this impl doesn't seem to cover every case that were covered in the old baseHelpers humanizeValue function. what do you do with the dash-separated cases like `this-is-an-example`?

packages/hawtio/src/plugins/shared/util/helpers.test.ts
tadayosi, line 1: The tests are ok but a bit verbose while covering few case. I think we can be more concise to cover more cases, like:
```
test('humanize strings', () => {
    expect(humanizeString('ObjectName').toEqual('Object Name')
    expect(humanizeString('XHTTPRequest').toEqual('XHTTP Request')
    ...
```

Also you should put it into the existing `./util/string.test.ts`.

packages/hawtio/src/plugins/shared/util/helpers.ts
tadayosi, line 14: also, we use `humanizeXxx` for performing the same thing in the old verison. I don't see reasons dare to use the other name here. Changing the name makes it a bit hard to map between the new and old impl in this case, IMO.

packages/hawtio/src/plugins/jmx/JmxContent.tsx
tadayosi: OK, so what I still don't see yet is the difference between `NodeNameTable` and `JMXContentMBeans`. Why do we need the two instead of one?  You said:

> JMXContentMBean is the default view, which only shows the objectNames

it suggests that the node is actually a parent, as otherwise it shouldn't have a set of objectNames. So `NodeNameTable` is for a parent but the default view is also for a parent, right?  I guess then we really don't yet clearly identify the true conditions to dispatch each view.

Apart from code, let's think it logically. I think there are four cases to capture in JMX view:

1. Single MBean
2. Multiple MBeans of the same kind
3. Multiple MBeans of different kinds
4. Mixture of MBeans and folder nodes

Are they MECE?  Then `NodeNameTable` corresponds to 4 and the default to 3, right?  If so, probably the conditions you defined haven't captured those switches clearly, because the difference between 3 and 4 is not whether it's parent or not.

joshiraez: Case 3 with NodeName

image

Case 4 with ObjectName which is the default view

image

Case 4 happens when is a folder that is not  a topmost parent if that makes thing more clear. Actually it happens if no other conditions are met according to the other version, as I pasted before.

I can also work on a simplified version of the old code if that's easier...? I mean, at this point I'm not sure what's the problem

tadayosi: Sorry if it confuses you, but in this case the behaviours of the legacy version might be a bit buggy, at least it's not consistent. I think we should think up a clearer specification now.

In case 4, why do we need different views?  I don't think it makes sense to distinguish a top-most parent from the other parents. For case 4, I think we should show object names for mbeans but just show the names for folders. So it's what `JMXContentMBeans` does, I think.

> Case 2 and 3 are kind of the same and use the same view (tabular attribute table)

Hmm, then how can the table shows MBeans that have different sets of attributes?  I thought it has a clear difference between the two.

So, to wrap up, now I think we only need three views:

1. `Attributes` for case 1
2. `AttributesTable` for case 2
3. `JMXContentMBeans` for case 3 & 4

WDYT?

joshiraez: >    Case 2 and 3 are kind of the same and use the same view (tabular attribute table)

> Hmm, then how can the table shows MBeans that have different sets of attributes? I thought it has a clear difference 
> between the two.

It defaults to the nodeNameTable if it doesn't match (or now the JMXContentMBeans table), or if there is no valid information.

Alright about the other thing. I only have to remove the new view and the conditional and it should be fine.

joshiraez: Ok so there is no current examples in current HawtIO to test case 3 (different kind of MBeans). I have checked all current nodes that are shown and every case always bring the same key structure so far. I'll put the code in place anyway to default to JMXContentMBeans when they don't match.

Also for the case where some MBeans don't bring info at all it also defaults to the JMXContentMBeans view
image
image

packages/hawtio/src/plugins/shared/util/helpers.ts
joshiraez: Agreed. Will change it

joshiraez: Added with the corresponding test

packages/hawtio/src/plugins/shared/util/helpers.test.ts
tadayosi: tbh I don't like it. it doesn't add any values other than we have some fancy test cases.

if you'd like to be creative in testing, there are other important issues that require it, such as:
- https://github.com/hawtio/hawtio-next/issues/190

packages/hawtio/src/plugins/shared/attributes/AttributeTable.tsx
```diff
@@ -0,0 +1,88 @@
...
+      Object.keys(attribute).every(label => Object.keys(attributes[0]).includes(label)),
+    )
+  }
+
+  if (
+    attributes.some(attribute => Object.keys(attribute).length === 0) ||
+    attributes.some(attribute => Object.values(attribute).length === 0) ||
+    !checkIfAllMBeansHaveSameAttributes(attributes)
+  ) {
+    return <JmxContentMBeans />
```
tadayosi, line 73: ok, so AttributeTable redispatches to JmxContentMBeans when the children are not of same kind. This flow looks a bit ugly to me but let's put up with it.

packages/hawtio/src/plugins/shared/attributes/AttributeTable.tsx
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    const firstAttributeElements = attributes[0].length
+
+    if (!attributes.every(attribute => attribute.length === firstAttributeElements)) {
+      return false
+    }
+
+    return attributes.every(attribute =>
+      Object.keys(attribute).every(label => Object.keys(attributes[0]).includes(label)),
```
tadayosi, line 64: while performance shouldn't be of much problem, you could do a bit better here. It's O(N^2) right now.

1. You can use `Set` and `Set.has` for the first attributes
2. Put the first attributes set outside of the every loop

Then it could become O(N) per attributes. If you can find find a better way to compare the two sets in TS, it could be more efficient.

packages/hawtio/src/plugins/shared/attributes/AttributeTable.tsx
joshiraez: As you need to fetch the data before making any decisions, this I found to be the better way right now to do this.

I'd like to do another refactor on to this that makes so the parent component is the one responsible for fetching the necessary data and then supplies it to the components below who only have to render, but I wanted to have something working first and then refactor later to a better design (nor that I didn't know if it'd be wanted so I decided to put that part in a separate PR later)

tadayosi: yeah, I understand that we need to fetch attributes set proactively to make decisions at the earlier stage, so I also think we can leave it as-is for now. Let's fix the other parts and try to merge it first.

packages/hawtio/src/plugins/jmx/JmxContent.tsx
```diff
@@ -39,10 +39,23 @@ export const JmxContent: React.FunctionComponent = () => {
...
+
+  const tableSelector: (node: MBeanNode) => React.FunctionComponent = (node: MBeanNode) => {
+    const tablePriorityList: { condition: (node: MBeanNode) => boolean; element: React.FunctionComponent }[] = [
+      { condition: mBeanApplicable, element: Attributes },
+      { condition: mBeanCollectionApplicable, element: AttributeTable },
+      { condition: DEFAULT, element: JmxContentMBeans },
+    ]
+
+    //Find first
+    return tablePriorityList.filter(entry => entry.condition(node))[0].element
```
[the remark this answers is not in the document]
tadayosi: You are doing the same thing with `array.filter`. you're not just covering the case when the return array is empty from the type point of view. instead of the current impl, the following should be more type safe:
```
return tablePriorityList.find(entry => entry.condition(node)) ?? JmxContentMBeans
```
plus, you'll no longer need `DEFAULT` const.

packages/hawtio/src/plugins/shared/attributes/AttributeTable.tsx
```diff
@@ -0,0 +1,90 @@
...
+    const firstMBeanAttributesElements = attributesList[0].length
+
+    if (!attributesList.every(mbeanAttributes => mbeanAttributes.length === firstMBeanAttributesElements)) {
+      return false
+    }
+
+    const labelSet: Set<string> = new Set()
+    Object.keys(attributesList[0]).forEach(label => labelSet.add(label))
+
+    return attributesList.every(attribute => Object.keys(attribute).every(label => labelSet.has(label)))
```
tadayosi, line 66: `attribute` -> `attributes`

packages/hawtio/src/util/strings.ts
```diff
@@ -99,3 +99,23 @@ export function parseBoolean(value: string): boolean {
 
   return /^true$/i.test(value) || parseInt(value) === 1
 }
+
+/**
+ * Will format a property to a standart human readable string with its spaces.
```
tadayosi, line 104: standart -> standard

packages/hawtio/src/plugins/jmx/JmxContent.tsx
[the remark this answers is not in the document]
joshiraez: I changed it to `return tablePriorityList.find(entry => entry.condition(node))?.element ?? JMXContentMBeans` as I still have to access the element to return it

I'd also still need the constant for this line for better readability https://github.com/hawtio/hawtio-next/pull/215/files#diff-64c355af4f9aa8d4446600bdc8f8295261dca01a008b626303d62019b5567c98R57 although I renamed it to ALWAYS

packages/hawtio/src/util/strings.test.ts
```diff
@@ -41,4 +41,13 @@ describe('strings', () => {
...
     expect(parseBoolean('0')).toBeFalsy()
   })
+
+  test('humanizeLabels()', () => {
+    expect(humanizeLabels('ObjectName')).toEqual('Object Name')
+    expect(humanizeLabels('XHTTPRequest')).toEqual('XHTTP Request')
+    expect(humanizeLabels('MBeanName')).toEqual('MBean Name')
+    expect(humanizeLabels('MBeanHTML')).toEqual('MBean HTML')
+
+    expect(humanizeLabels('object-name')).toEqual('Object Name')
```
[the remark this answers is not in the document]
joshiraez: I think I already did this, with some malformed labels and mixing them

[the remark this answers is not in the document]
tadayosi: what about empty string `''` or string with only spaces `'                  '`?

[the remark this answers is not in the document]
joshiraez: Added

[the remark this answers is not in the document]
joshiraez: I see. Any more tests you see would be a good idea to implement?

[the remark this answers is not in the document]
tadayosi: looks good to me now :+1:

tadayosi (approved):
thanks!

PR #218, opened 2023-04-04
chore(route-diagram): Remove console.log entries

mmelko:


joshiraez (approved):

tadayosi (approved):
if it's helpful you could keep them as either log.info() or log.debug with the dedicated component logger. see how other components handle logging.

packages/hawtio/src/plugins/camel/route-diagram/visualization-service.ts
```diff
@@ -234,7 +234,7 @@ class VisualizationService {
         siblingNodes = siblings
       } else {
         siblingNodes = [nodes.length - 1]
-        console.log(siblingNodes, [nodes.length - 1], nodes)
+        siblingNodes = [nodes.length - 1]
```
tadayosi, line 237: why do we need to repeat the same thing twice?

mmelko: huh 😕  .. we don't thanks!

PR #224, opened 2023-04-05
Endpoints View

phantomjinx:


packages/hawtio/src/plugins/camel/endpoints/endpoints-service.ts
```diff
@@ -0,0 +1,128 @@
...
+  if (!endPointPath) console.error('createEndpointFromData: endpoint path must be defined')
+
+  let uri = componentName + '://' + endPointPath
+
+  log.debug('Have endpoint data ' + JSON.stringify(parameters))
+
+  let params = ''
+  Object.entries(parameters).forEach(([key, value]) => {
+    const prefix = params ? '&' : ''
+    params += prefix + key + '=' + value
```
joshiraez, line 102: I think this can be put as 

& + Object.entries(parameters).map(entry => entry.join('='))

image

joshiraez (approved):
Except for the comment, LGTM. Great job.

packages/hawtio/src/plugins/camel/endpoints/endpoints-service.ts
joshiraez: So I believe you can do something like

```
const uri = componentName + '://' + endPointPath
              + params 
              ? '&' + Object.entries(parameters).map(entry => entry.join('=')).join('&')
              : ''

createEndpoint(node, uri)
```

packages/camel-model/src/index.ts
```diff
@@ -1,2 +1,9 @@
-import { _apacheCamelModelVersion, _apacheCamelModel } from './camelModel'
-export { _apacheCamelModelVersion as apacheCamelModelVersion, _apacheCamelModel as apacheCamelModel }
+import definitions from './definitions-camel-model.json'
+import rests from './rests-camel-model.json'
+import dataformats from './dataformats-camel-model.json'
+import languages from './languages-camel-model.json'
+import components from './components-camel-model.json'
+
+var apacheCamelModelVersion = '3.20.2'
```
tadayosi, line 7: so now is the camel model version not generated automatically anymore?  then we should set up a way to update this part from maven `version.org.apache.camel` property.

tadayosi: FYI, see `replace-version` script in packages/hawtio/package.json

packages/hawtio/src/plugins/camel/endpoints/endpoints-service.ts
```diff
@@ -0,0 +1,128 @@
...
+import { jolokiaService } from '@hawtiosrc/plugins/connect'
+import * as schema from '@hawtio/camel-model'
+import { MBeanNode, workspace } from '@hawtiosrc/plugins/shared'
+import * as ccs from '../camel-content-service'
+import { endpointsType, log } from '../globals'
+import { isObject } from '@hawtiosrc/util/objects'
+
+export class Endpoint {
+  constructor(public uri: string, public state: string, public mbean: string) {}
+}
```
tadayosi, line 11: if you don't plan to add behaviours to the class, let's make it type alias.

packages/camel-model/src/index.ts
phantomjinx: This file is generated automatically from the maven generator plugin. It is still named, by default `camelModel.js`, but we rename it to `index.ts` (effectively convert it to typescript). Therefore, the camel version is retrieved from the camel-model dependency used in the model generation.

packages/camel-model/pom.xml
```diff
@@ -65,14 +66,14 @@
                 if (!file.exists()) {
```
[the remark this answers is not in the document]
phantomjinx: Turns out we do need to retain it now that we are linting. The generator plugin generates the main src file as `camelModel.js` so pure javascript. We changes its name to `index.ts` to not only make it a module but also typescript. Therefore, to lint successfully it must conform to the linting rules, ie. no semi-colons and var->const.

So now added to linting, renamed to `index.ts` and uses groovy to make the javascript conform to typescript syntax.

tadayosi (approved):

PR #225, opened 2023-04-05
 feat(route-diagram): Add route statistics into the diagram

mmelko:
Added route diagram stats that are periodically refreshed:
Screenshot 2023-04-05 at 23 02 20

TBD:
 - add tests for the parsing the xml 
 - set the refresh period according hawtio settings 
 - add values for the edges


packages/hawtio/src/plugins/camel/route-diagram/routeDiagram.css
```diff
@@ -32,7 +32,12 @@
 .react-flow__node-camel .number {
   align-items: center;
 }
-.react-flow__node-camel .camelNodeLabel {
+.nodeTooltip {
```
tadayosi, line 35: Using camel case for css names is rare. What about using names with hyphens?  i.e. `node-tooltip`

packages/hawtio/src/plugins/camel/route-diagram/RouteDiagram.tsx
```diff
@@ -16,6 +16,8 @@ import { CamelNodeData, visualizationService } from '@hawtiosrc/plugins/camel/ro
 import 'reactflow/dist/style.css'
 import './routeDiagram.css'
 import { Page } from '@patternfly/react-core'
+import { routesService } from '@hawtiosrc/plugins/camel/routes-service'
```
tadayosi, line 19: it's in the same plugin module, so you can use relative path `../routes-service`.

packages/hawtio/src/plugins/camel/route-diagram/visualization-service.ts
```diff
@@ -85,22 +87,47 @@ class VisualizationService {
...
+      const routeId = route.id
       if (!selectedRouteId || !routeId || selectedRouteId === routeId) {
-        this.addRouteXmlChildren(route, nodes, links, '')
+        this.addRouteXmlChildren(route, nodes, links, routeId, '')
       }
     }
-    const camelNodes = nodes.map(node => ({ id: node.id, data: node, position: { x: 0, y: 0 }, type: 'camel' }))
+    //parse stats
+    const camelNodes = nodes.map(node => {
+      return {
```
tadayosi, line 101: no need for `return`

packages/hawtio/src/plugins/camel/routes-service.ts
```diff
@@ -20,6 +20,55 @@ export type CamelRoute = {
   MeanProcessingTime: number
 }
 
+export type ProcessorStats = {
+  id: string
+  index: number | string
+  state: string | null
```
tadayosi, line 26: can `string | null` be simply optional, i.e. `state?: string`?

packages/hawtio/src/plugins/camel/routes-service.ts
```diff
@@ -20,6 +20,55 @@ export type CamelRoute = {
...
+  externalRedeliveries: number | string
+  minProcessingTime: number | string
+  maxProcessingTime: number | string
+  deltaProcessingTime: number | string
+  meanProcessingTime: number | string
+  startTimestamp: number | string
+  resetTimestamp: number | string
+  firstExchangeCompleteTimestamp?: number | null
+  firstExchangeCompletedExchangeId?: number | null
+  lastExchangeCompleteTimestamp?: number | null
```
tadayosi, line 40: it doesn't seem to make sense to have it optional and type `number | null` at the same time. can't you simply `lastExchangeCompleteTimestamp?: number`?

same for all other variables.

packages/hawtio/src/plugins/camel/routes-service.ts
```diff
@@ -20,6 +20,55 @@ export type CamelRoute = {
...
+}
+
+export type RouteStats = {
+  id: string
+  state: string | null
+  sourceLocation: string | null
```
tadayosi, line 46: same here

packages/hawtio/src/plugins/camel/routes-service.ts
```diff
@@ -20,6 +20,55 @@ export type CamelRoute = {
...
+  resetTimestamp: number | string
+  firstExchangeCompletedTimestamp?: string | null
+  firstExchangeCompletedExchangeId?: string | null
+  firstExchangeFailureTimestamp?: string | null
+  firstExchangeFailureExchangeId?: string | null
+  lastExchangeCreatedTimestamp?: string | null
+  lastExchangeCompletedTimestamp?: string | null
+  lastExchangeCompletedExchangeId?: string | null
+  lastExchangeFailureTimestamp?: string | null
+  lastExchangeFailureExchangeId?: string | null
```
tadayosi, line 69: same here

packages/hawtio/src/plugins/camel/routes-service.ts
```diff
@@ -205,6 +256,102 @@ class RoutesService {
   async deleteRoute(objName: string) {
     await jolokiaService.execute(objName, 'remove()')
   }
+
+  async dumbRoutesStatsXML(routesNode: MBeanNode | null): Promise<string | null> {
+    if (!routesNode) {
+      throw new Error('No route/routes node provided')
+    }
```
tadayosi, line 263: if it doesn't accept `null` `routesNode`, you shouldn't make it `MBeanNode | null`. and you can remove the if guard.

packages/hawtio/src/plugins/camel/routes-service.ts
```diff
@@ -205,6 +256,102 @@ class RoutesService {
...
+      state: pDoc.getAttribute('state'),
+      index: parseInt(pDoc.getAttribute('index') ?? '0'),
+      lastExchangeCompleteTimestamp: parseInt(pDoc.getAttribute('lastExchangeCompleteTimestamp') ?? '0'),
+      maxProcessingTime: parseInt(pDoc.getAttribute('maxProcessingTime') ?? '0'),
+      meanProcessingTime: parseInt(pDoc.getAttribute('meanProcessingTime') ?? '0'),
+      minProcessingTime: parseInt(pDoc.getAttribute('minProcessingTime') ?? '0'),
+      redeliveries: parseInt(pDoc.getAttribute('redeliveries') ?? '0'),
+      resetTimestamp: parseInt(pDoc.getAttribute('resetTimestamp') ?? '0'),
+      sourceLineNumber: parseInt(pDoc.getAttribute('sourceLineNumber') ?? '0'),
+      startTimestamp: parseInt(pDoc.getAttribute('startTimestamp') ?? '0'),
```
tadayosi, line 312: oh I didn't know the notation `??`. it seems we should prefer this one over `||` in other places as well in most cases.

tadayosi (commented):
overall it looks great. there are only a few minor styling comments.

packages/hawtio/src/plugins/camel/routes-service.ts
```diff
@@ -205,6 +256,102 @@ class RoutesService {
   async deleteRoute(objName: string) {
     await jolokiaService.execute(objName, 'remove()')
   }
+
+  async dumbRoutesStatsXML(routesNode: MBeanNode | null): Promise<string | null> {
```
phantomjinx, line 260: Should this be `dump`??

packages/hawtio/src/plugins/camel/routes-service.ts
```diff
@@ -205,6 +256,102 @@ class RoutesService {
...
+  createRouteStats(pDoc: Element): RouteStats {
+    const procStats = pDoc.getElementsByTagName('processorStat')
+    const processorStats: ProcessorStats[] = []
+    for (const pStat of procStats) {
+      processorStats.push(this.createProcessorStats(pDoc.id, pStat))
+    }
+
+    return {
+      id: pDoc.id,
+      deltaProcessingTime: parseInt(pDoc.getAttribute('deltaProcessingTime') ?? '0'),
```
phantomjinx, line 325: Since these parseInt functions are all the same, can they be replaced with a single shorter function?

eg.
```
parseAttr(pDoc: any, name: string) {
  pDoc.getAttribute(name) ....
}
```

packages/hawtio/src/plugins/camel/routes-service.ts
```diff
@@ -205,6 +256,102 @@ class RoutesService {
...
+      meanProcessingTime: parseInt(pDoc.getAttribute('meanProcessingTime') ?? '0'),
+      minProcessingTime: parseInt(pDoc.getAttribute('minProcessingTime') ?? '0'),
+      processorStats: processorStats,
+      redeliveries: parseInt(pDoc.getAttribute('redeliveries') ?? '0'),
+      resetTimestamp: parseInt(pDoc.getAttribute('resetTimestamp') ?? '0'),
+      sourceLocation: pDoc.getAttribute('sourceLocation'),
+      startTimestamp: parseInt(pDoc.getAttribute('startTimestamp') ?? '0'),
+      state: pDoc.getAttribute('state'),
+      totalProcessingTime: parseInt(pDoc.getAttribute('totalProcessingTime') ?? '0'),
+      firstExchangeCompletedExchangeId: pDoc.getAttribute('firstExchangeCompletedExchangeId'),
```
phantomjinx, line 343: Likewise, it seems that maybe a loop over `pDoc` attributes would be less verbose?

packages/hawtio/src/plugins/camel/route-diagram/visualization-service.ts
joshiraez: If there is a TS error when trying to return an object, put it between parentheses

https://ultimatecourses.com/blog/return-object-arrow-function

packages/hawtio/src/plugins/camel/routes-service.ts
joshiraez: Neither did I, and after investigating a bit its really handy to avoid some edge cases with falsy values 😄

packages/hawtio/src/plugins/camel/routes-service.ts
joshiraez: A more cleaner way to avoid a lot of boilerplate, taking into account refactoring the parseInt part to its own function.

```
Const obj = Object.fromEntries([attributeName,
attributeName2,
...].map(attribute => getFromPdoc(attribute))

Obj.id =pdoc.id

Return obj
```

If there are any special case can be added to the object like with the id.


Anothwr way

```
Const attributesToRetrieve =[attr1, attr2, attr3...]

Return {
id: p.id
...Object.fromEntries(attributesToRetrieve.map(getFromPdoc))
```

packages/hawtio/src/plugins/camel/routes-service.ts
```diff
@@ -205,6 +256,102 @@ class RoutesService {
...
+
+  async dumbRoutesStatsXML(routesNode: MBeanNode | null): Promise<string | null> {
+    if (!routesNode) {
+      throw new Error('No route/routes node provided')
+    }
+
+    let xml = null
+    let mbeanName = routesNode.getProperty(contextNodeType)
+
+    try {
```
joshiraez, line 268: Put the logic inside named functions instead of using a comment imo

joshiraez:
Tidbits:

* Shouldnt this be a draft?
* Any way to just comment a PR without blocking it? Does any comment create a conversation thus blocking it?

mmelko:
@tadayosi @phantomjinx I did some changes here. I changed the types for the attributes of statistics to the strings. That makes it easier to work with. If we will do some math things we can convert later.

I also added the tests covering parsing. I still haven't added Inflight statistic to the connection between the nodes, however it's visible in the small table on hover so I guess it's not crucial here. It can be done later. 

I also did some preparation for showing all the statistics not only the one as in old hawtio but I have to think about it more how to display it properly. However I believe if you are ok with this it can be merged and rest can be tweaked with future PRs.

tadayosi (approved):
There are still some comments that are not yet addressed, but I think it's ok to merge.

packages/hawtio/src/plugins/camel/routes-service.ts
mmelko: @joshiraez Thanks for the suggestion, but I don't think I understand. The comment here is just for explanation why I'm checking mBeanName.

mmelko:
UPDATE: I added inflights exchanges number, just not tied to the connection edge but rather to the node.  Screenshot 2023-04-13 at 07 56 27

packages/hawtio/src/plugins/camel/routes-service.ts
mmelko: whoops.

packages/hawtio/src/plugins/camel/routes-service.ts
joshiraez: I feel like it's a bit hard to understand. I was thinking about putting them in their own functions to avoid the comment.

I have a small question though, you execute with an empty string for the mbeanName in case there is no object name for route node? Or you just catch the error?

In any case, personally I'd try to make clearer the logic for it as it's really hard to understand the if else. The operations are very similar and the intention is not clear. What do you think about something like this?

```
const routeNodeOperation = dumpRouteStatsAsXml
const routesFolderOperation = dumpRoutesStatsAsXml

const mbeanName = routesNode.getProperty(contextNodeType)

const operationForMBean = mbeanName ? routesFolderOperation : routeNodeOperation
const mbeanToQuery = mbeanName ? mbeanName :  routesNode.objectName

try {
  const xml = await jolokiaService.execute(mbeanToQuery ?? '', operationForMBean, [true, true])
} catch (error) {
  throw new Error('Failed to dump routes stats from mbean: ' + mbeanName + error)
}

if (!xml) {
    throw new Error('Failed to extract any xml from mbean: ' + mbeanName)
}

return xml as string
```

joshiraez: Using the ternary in mbeanToQuery because the ?? operator only works for void types. This way I preserve the original logic, you don't need to add a comment as it can be seen that it depends on if mbean is set or not, avoid the if-else and avoid the use of let. What do you think?

image

joshiraez: (If you don't like it we can leave it as is, it's just that I found it a bit confusing at first discerning the logic with the comment)

joshiraez (approved):

Issue #105, opened 2023-02-15
Should hawtio plugin retain use of the @hawtiosrc alias


phantomjinx:
Due to the introduction of a new package (camel-model), the `@hawtio` ident prioritizes the new purpose of being an organisation. In order to avoid confusion and failure, the ident referencing src paths is modified to `@Hawtiosrc` (see https://github.com/hawtio/hawtio-next/pull/93).

Is this the correct alias?
Should this alias even be retained or is there something better?

tadayosi:
When I introduced original `@hawtio` alias, it was designed so that each module can be independent of outer modules by not needing to refer to them using relative path `../` and make it self-contained with their own files under its own directory.

But with the code growing, it might be that with the usage of the local alias we have more to lose than to gain. Maybe we should remove the alias and just use relative path to import outer modules.

tadayosi:
At the same time, use of relative paths can be another source of annoyance.
https://stackoverflow.com/questions/43281741/how-to-use-paths-in-tsconfig-json

Possible candidates for better alias naming:
- `@hawtio-src`
- `@hawtio-root`
- `@package-root`
- `@src`
- `__hawtio__`
- `__src__`

tadayosi:
We keep using `@hawtiosrc`. Until we come up with better name, we can close it for now.

PR #232, opened 2023-04-11
feat(plugins): split @hawtio/react package into core and plugins

### LINKED ISSUE #229: Split `@hawtio/react` package into core and plugins
There is a request from Artemis dev team (the main consumer of the plugin API) to expose key objects such as `workspace` and `treeProcessorRegistry` from `@hawtio/react`'s `src/plugins`.

While we can just carelessly export everything in `src/plugins` from `src/index.ts` (like below) so that users can consume them from `@hawtio/react`, I don't think it's a good design because we cannot provide a clear separation between the core and plugins.
```js
export * from './plugins`
```

To achieve the goal of providing the resources from the plugins to users while keeping a clear separation between core and plugins, currently I think the only way is to split the `@hawtio/react` package into two: `@hawtio/react` (core) and `@hawtio/react-plugins` (plugins). That means the packages should look like this:

```
packages/
├── camel-model/
│   └── package.json
├── hawtio/
│   └── package.json
└── plugins/
    └── package.json
```

This is getting more to resemble the legacy JS components structure, but since we'll still keep monorepo it shouldn't be of much burden for releasing each package unlike it was with the legacy Hawtio.

What do you think? @phantomjinx @mmelko @joshiraez

tadayosi:
This is required to export plugins types to external plugin developers.

Splitting packages also demands exporting of util functions so that they can be shared with the plugins package. For now, util is exported with tsup multi entries, which means util cannot share any states with the core, that is, only side-effect free functions can be exported util.

Due to circular dependencies between core and plugins, it's hard to move everything in plugins into `@hawtio/react-plugins`. Thus, the common parts -- shared, connect, jmx, and rbac -- are still left in `@hawtio/react` for now. Ideally, we should move all of them to `@hawtio/react-plugins` as well after circular dependencies are resolved.

fix #229


tadayosi:
this also addresses #105 by changing `@hawtiosrc` to `__root__` for the alias of the src root path.

tadayosi:
@phantomjinx @mmelko @joshiraez Since this has a big impact on the entire project, please have a look and give me feedback before merging it.

tadayosi:
@phantomjinx Some Camel tests don't work well with the split package. It's because they depends on the internals of the workspace tree but the impl is in the separate core package, thus hard to substitute the internal parts with mock. As it seems not easy to fix them quickly, I ended up needing to skip the failing tests for now. We need to restore them later.
https://github.com/hawtio/hawtio-next/pull/232/commits/b1800e66600c7df36a82d0a141f8e59d5134502e

packages/hawtio/src/util/objects.ts
```diff
@@ -12,7 +12,7 @@ export function isEmpty(obj: object): boolean {
 }
 
 export function isString(value: unknown): value is string {
-  return typeof value === 'string' || value instanceof String
+  return value != null && typeof value.valueOf() === 'string'
```
joshiraez, line 15: Why not just typeof value?

joshiraez (approved):
Other than that and after looking at all the files it looks good. A lot of changes

packages/hawtio/src/util/objects.ts
tadayosi: because it not only handles `string` but also `String` objects.

tadayosi:
Marked it back to draft as we might not want to merge it yet.

tadayosi:
supeceded by #249

PR #236, opened 2023-04-13
fix(#125): Fixed component not polling for data to keep it updated

joshiraez:
Adds polling to the tabular attribute table to keep its data updated.

Planed to add it together with the Chart component after seeing that the other idea got scrapped, but I thought it was better to submit the fix now.


packages/hawtio/src/plugins/shared/attributes/AttributeTable.tsx
```diff
@@ -5,34 +5,81 @@ import { PluginNodeSelectionContext } from '@hawtiosrc/plugins/selectionNodeCont
...
-      setAttributesList([...childrenMbeansAttributes])
-      setIsReading(false)
-    }
-    readAttributes()
+    readAttributes(selectedNode)
+    setReadingJobs(selectedNode)
+
+    return () => attributeService.unregisterAll()
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
joshiraez, line 82: Needed if I don't want to put all the functions inside of the useEffect or have to use useCallback for every one (even though they don't have any dependencies with outside code)

tadayosi (approved):

PR #239, opened 2023-04-14
refactor(#226): Changed order of tabs

joshiraez:
Changed order of tabs following tadayosi's suggestion as I thought it was fine enough and it was very quick. Feel free to change or reorder it if you need it after more discussion

image

### DIFF
--- a/packages/hawtio/src/plugins/camel/CamelContent.tsx
+++ b/packages/hawtio/src/plugins/camel/CamelContent.tsx
@@ -58,6 +58,8 @@ export const CamelContent: React.FunctionComponent = () => {
     return ccs.hasMBean(node) && !ccs.isContextsFolder(node) && !ccs.isRoutesFolder(node) && !ccs.isRouteXmlNode(node)
   }
 
+  // The order of the items in the following list is the order in will the tabs will be visualized.
+  // For more info check: https://github.com/hawtio/hawtio-next/issues/237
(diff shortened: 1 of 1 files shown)


packages/hawtio/src/plugins/camel/CamelContent.tsx
```diff
@@ -59,6 +59,8 @@ export const CamelContent: React.FunctionComponent = () => {
   }
```
phantomjinx, line 60: Might just want a comment here documenting that this array is ordered in a particular way and document that order as a list. Otherwise, people like me will just come along and add another view in the middle without thinking about it :-)

phantomjinx (changes requested):
Small point of a comment needed, I think.

packages/hawtio/src/plugins/camel/CamelContent.tsx
joshiraez: Good idea, I'll add it

joshiraez: Er forgot about this and I have to go. I'll add it tomorrow

joshiraez: I did it up from GitHub UI

phantomjinx (approved):
LGTM

PR #240, opened 2023-04-14
refactor(#125): Small optimization and removed eslint warning

joshiraez:
This is a small optimization I found after reading about the lint warning. Supposedly, every function will get reevaluated in every component render (IE, after every fetch actually). There is some other ways to go about it but I found this to be the easiest and simplest one.

The only change was moving all the functions inside constants inside `useEffect` and using the supplied value of `setAttributeTable` to avoid having to put attributeList as a dependency too.

This way the warning is fixed and is a small optimization on the table component

There is a very good explanation in this URL if you want to know more about it https://codevoweb.com/react-hook-useeffect-has-a-missing-dependency-error/?utm_content=cmp-true


phantomjinx (approved):
LGTM

Issue #188, opened 2023-03-22
Investigate why generate-camel-svg-index.js doesn't work with `import`


tadayosi:
Currently, [packages/hawtio/scripts/generate-camel-svg-index.js](https://github.com/hawtio/hawtio-next/blob/main/packages/hawtio/scripts/generate-camel-svg-index.js) doesn't work if the require parts to imports #164.
```
const fs = require('fs')
const path = require('path')
```

Let's investigate and rewrite it back to imports like this:
```
import fs from 'fs'
import path from 'path'
```

joshiraez:
If it's not a priority I can take this out to investigate in blocked time

joshiraez:
I'll be assigning myself in the meanwhile

joshiraez:
I solved the issue and put some background about it in #241

joshiraez:
I'll be closing this issue as well as it's fine like it is right now.

PR #241, opened 2023-04-14
Fix import on yarn generate:camel-icons

joshiraez:
So I don't know since when #188 failed but the issue was that

Import (from ES6) is an async operation and returns a Promise
Require(from CJS) is a sync operation and returns the "module"

There are 2 ways to go about this:

* First one is to convert this file to a module, but that requires changing the type in package.json to a module, doing some small fixes and it works. That also allows to use the `import path from 'path'`way of importing.  Is the first commit in the branch if you want to check
  * Problem: As changing to type: module changes how files are treated, on yarn start there are some errors about missing modules.
* The current version is much less invasive and uses the import function. Both the command and yarn start work without problems.

Another option is to leave it as is, as there are more small differences between require and import (require executes the module, import doesn't, for example). If we want to go with using the import way, I feel adding a Promise resolve is not much of an issue.

What are your opinions?


packages/hawtio/scripts/generate-camel-svg-index.js
```diff
@@ -1,114 +1,113 @@
...
-      expContent.push(...contents[1])
-      expNames.push(...contents[2])
-    }
+const fs = import('fs')
+const path = import('path')
 
-    if (!ent.name.endsWith('.svg')) {
-      continue
-    }
+Promise.all([fs, path]).then(([fs, path]) => {
```
joshiraez, line 5: It seems like a lot of changes but I only moved the whole code inside the Promise resolve

tadayosi (commented):
The first one (converting to `module` type) seems to be the ideal approach. However, as you say, it should cause other problems elsewhere and we should consider what are the real implications of the change.

If converting to `module` isn't always a good thing, then the second best approach would be to leave it as-is (no changes). After all, the motivation of this issue is just removing the `eslint-disable` comment, which is kind of aesthetic issue. So long as we accept the comment, there should be no other imminent issues.

joshiraez:
Then I think it's best to leave it as is as the behaviour is mostly the same. I'll be closing this

PR #254, opened 2023-04-25
fix(test): Normalized tags for testing

joshiraez:
Quick fix to unblock https://github.com/hawtio/hawtio/issues/2826.

I was unable to find how it was generated in other version actually, but from inspecting the other version, it seems to just be concatenating the parents.

This is just a quick fix and will need to be refactored:
- The ideal is to maybe change the IDs generated by the MBeans to be like this, but I was unable to find this info for JMX, only for Camel (in the Camel Tree Processor).
- I don't know if those ids are used for something else
- If there is no search done I'm quite sure I'm mutating the MBeans themselves as well when doing this. While navigating a bit through the app, I didn't find any bugs though.
- As the tree comes from patternfly, there was no way to just change the component to have another logic in the tag, unfortunately

Screenshot from 2023-04-25 19-06-19

Sorry for the quick thingie but I saw it was blocking him and I'm on holiday from tomorrow onwards. But this should help in the testing process I think and make it consistent. We can further refine it after he has this merged and decide on another solution following this naming convention

### DIFF
--- a/packages/hawtio/src/plugins/camel/CamelTreeView.tsx
+++ b/packages/hawtio/src/plugins/camel/CamelTreeView.tsx
@@ -86,6 +86,29 @@ export const CamelTreeView: React.FunctionComponent = () => {
     }
   }
 
+  const actuallyRenameAccordingToParents = (mbean: MBeanNode) => {
+    // The names concats all parent names together. I was unable to find source on base version, but from debugging the app,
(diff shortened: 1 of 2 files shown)


tadayosi (approved):
lgtm. there are minor problems but I'll take care of them after merge.

PR #259, opened 2023-04-27
fix(jmx,camel): consistent tree node id generation for e2e testing

tadayosi:
Address hawtio/hawtio#2826

Now we hide HTML `id` generation behind `MBeanNode` constructor and have it autogenerated based on its name and parent, since the `id` is not some logical identifier for a node, but the HTML `id` that is rendered in HTML. We don't need to care about assigning `id` when creating a node. But their consistency is required for e2e testing downstream at the hawtio standalone project.


tadayosi:
FYI @phantomjinx

PR #269, opened 2023-05-03
The Debug Tab and View

phantomjinx:
~_Trace tab to follow if still required.~

Trace tab now included.

### DIFF
--- a/packages/hawtio/src/plugins/camel/CamelContent.tsx
+++ b/packages/hawtio/src/plugins/camel/CamelContent.tsx
@@ -49,7 +49,7 @@ export const CamelContent: React.FunctionComponent = () => {
   interface NavItem {
     id: string
     title: string
-    componentFn: () => JSX.Element
+    component: JSX.Element
     isApplicable(node: MBeanNode | null): boolean
   }
 
@@ -63,42 +63,40 @@ export const CamelContent: React.FunctionComponent = () => {
   // The order of the items in the following list is the order in will the tabs will be visualized.
   // For more info check: https://github.com/hawtio/hawtio-next/issues/237
   const allNavItems: NavItem[] = [
-    { id: 'attributes', title: 'Attributes', componentFn: () => <Attributes />, isApplicable: mBeanApplicable },
-    { id: 'operations', title: 'Operations', componentFn: () => <Operations />, isApplicable: mBeanApplicable },
+    { id: 'attributes', title: 'Attributes', component: <Attributes />, isApplicable: mBeanApplicable },
+    { id: 'operations', title: 'Operations', component: <Operations />, isApplicable: mBeanApplicable },
     {
       id: 'contexts',
       title: 'Contexts',
-      componentFn: () => <Contexts />,
+      component: <Contexts />,
       isApplicable: (node: MBeanNode) => ccs.isContextsFolder(node),
     },
     {
       id: 'routes',
       title: 'Routes',
-      componentFn: () => <CamelRoutes />,
+      component: <CamelRoutes />,
       isApplicable: (node: MBeanNode) => ccs.isRoutesFolder(node),
     },
     {
       id: 'endpoints',
       title: 'Endpoints',
-      componentFn: () => <Endpoints />,
+      component: <Endpoints />,
       isApplicable: (node: MBeanNode) => ccs.isEndpointsFolder(node),
     },
     {
       id: 'routeDiagram',
       title: 'Route Diagram',
-      componentFn: () => {
-        return (
-          <RouteDiagramContext.Provider value={ctx}>
-            <RouteDiagram />
-          </RouteDiagramContext.Provider>
-        )
-      },
+      component: (
+        <RouteDiagramContext.Provider value={ctx}>
+          <RouteDiagram />
+        </RouteDiagramContext.Provider>
+      ),
       isApplicable: (node: MBeanNode) => ccs.isRouteNode(node) || ccs.isRoutesFolder(node),
     },
     {
       id: 'source',
       title: 'Source',
-      componentFn: () => <Source />,
+      component: <Source />,
       isApplicable: (node: MBeanNode) =>
         !ccs.isEndpointNode(node) &&
         !ccs.isEndpointsFolder(node) &&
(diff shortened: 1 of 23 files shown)


packages/hawtio/src/plugins/camel/debug/debug-service.ts
```diff
@@ -0,0 +1,301 @@
...
+  async setDebugging(node: MBeanNode, flag: boolean): Promise<boolean> {
+    const db = this.getDebugBean(node)
+    if (!db) return false
+
+    const max = flag ? 1 : 0
+    camelPreferencesService.saveCamelPreferences({ maximumTraceDebugBodyLength: max })
+
+    const streams = camelPreferencesService.loadCamelPreferences().isIncludeTraceDebugStreams
+
+    await jolokiaService.writeAttribute(db.objectName as string, 'BodyMaxChars', max)
```
tadayosi, line 82: why do we need `as string`?  if it's just because of converting `string | undefined` to `string`, rather we should check the existence of `objectName` before the sequence of the method calls to make type check satisfied.

packages/hawtio/src/plugins/camel/debug/Debug.tsx
```diff
@@ -0,0 +1,782 @@
...
+import { isArray } from '@hawtiosrc/util/objects'
+import { RouteDiagram } from '../route-diagram/RouteDiagram'
+import './Debug.css'
+import { Annotation, RouteDiagramContext, useRouteDiagramContext } from '../route-diagram/route-diagram-context'
+import { CamelNodeData } from '../route-diagram/visualization-service'
+import { IResponse } from 'jolokia.js'
+import { log } from '../globals'
+import { parseXML } from '@hawtiosrc/util/xml'
+
+export const Debug: React.FunctionComponent = () => {
```
tadayosi, line 60: This component seems to have a bit too many states in a single component. Perhaps we can split it into smaller ones so each can handle a smaller set of the states.

packages/hawtio/src/plugins/camel/debug/Debug.tsx
```diff
@@ -0,0 +1,782 @@
...
+  const doubleClickNodeAction = useCallback((): ((nodeData: CamelNodeData) => void) => {
+    return async (nodeData: CamelNodeData) => {
+      if (nodeData.routeIdx === 0) {
+        ds.notifyError('Cannot breakpoint on the first node in the route')
+        return
+      }
+
+      const bkps = await ds.getBreakpoints(selectedNode as MBeanNode)
+      if (bkps.includes(nodeData.cid)) {
+        handleRemoveBreakpoint(selectedNode as MBeanNode, nodeData.cid)
```
tadayosi, line 194: existence of `selectedNode` should be checked beforehand. -> `as MBeanNode`

tadayosi (approved):

packages/hawtio/src/plugins/camel/CamelContent.tsx
```diff
@@ -47,8 +48,8 @@ export const CamelContent: React.FunctionComponent = () => {
   interface NavItem {
     id: string
     title: string
-    component: React.FunctionComponent
-    isApplicable(node: MBeanNode): boolean
+    componentFn: () => JSX.Element
```
joshiraez, line 51: Why change from Component to JSX.Element? Just curious, as both should work but I have usually seen the other approach. Did it make something easier?

joshiraez: If it's because of 
```(
          <RouteDiagramContext.Provider value={ctx}>
            <RouteDiagram />
          </RouteDiagramContext.Provider>
 )```
 
 I believe you could just paste the part between parentheses and it should work. JSX.Element and React.FunctionComponent are interchangeable here. I would have to look deeper into it anyway to see if there are more differences

**Comment by joshiraez** [comment]
LGTM so far but I want to take another look at the tests before approving (and feel free to resolve the comment whenever you want. Is not a change request but rather to know more). I'll take another look during the day and approve :)

**Comment by phantomjinx** [comment]
@joshiraez @tadayosi 

Feel free to please re-review and update your comments as you see fit.

File: packages/hawtio/src/plugins/camel/routes/CamelRoutes.tsx [hunk 5]
```diff
@@ -30,12 +30,19 @@ export const CamelRoutes: React.FunctionComponent = () => {
...
       setRoutes(routes)
+
+      timeoutHandle = setTimeout(getRouteAttributes, 10000)
     }
 
     getRouteAttributes()
+
+    return () => {
+      clearTimeout(timeoutHandle)
+    }
```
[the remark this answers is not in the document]
tadayosi: The definiteve source of Jolokia API is this doc: https://jolokia.org/reference/html/clients.html#js-poller

According to the doc, Jolokia seems to support registering a scheduler with a bulk request. And our jolokia.js already seems to support it as well. So only thing we should do is to enhance our jolokia service to make it support bulk requests.
https://github.com/hawtio/jolokia.js/blob/main/index.d.ts#L318

That said, as I already approved it, I don't think it's by any means a blocker. So feel free to mark it as resolved and work on the suggested improvement later.

packages/hawtio/src/plugins/camel/debug/Debug.tsx
tadayosi: What I suggested was to split the single `Debug` react component into smaller parts, like `Debug` -> `DebugToolbar`, `DebugPanel` (if that makes sense). When you actually do so, you'll be surprised to find how React can manage states in smaller chunks in a more maintainable way.

But it's not a big deal anyway, we can always polish code later.

packages/hawtio/src/plugins/camel/routes/CamelRoutes.tsx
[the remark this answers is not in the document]
tadayosi: Btw, RBAC plugin massively relies on Jolokia bulk requests. You might find some idea on how to use them with Camel routes from this.
https://github.com/hawtio/hawtio-next/blob/main/packages/hawtio/src/plugins/rbac/tree-processor.ts#L76

tadayosi:
My comments are general feedback, not to intend to block it. I think this pull req is great enough, so please feel free to go ahead and merge it.

tadayosi (approved):
My comments are general feedback, not to intend to block it. I think this pull req is great enough, so please feel free to go ahead and merge it.

Issue #10, opened 2022-12-23
JMX - Chart


tadayosi:
Let's port https://github.com/hawtio/hawtio-integration/tree/main/plugins/jmx/ts/charts

tadayosi:
cc @joshiraez

joshiraez:
I'll pick this up

PR #293, opened 2023-05-17
feat(#10): Chart JMX

joshiraez:
image

New watch selection modal
Screenshot from 2023-06-19 09-06-53

### DIFF
--- a/packages/hawtio/package.json
+++ b/packages/hawtio/package.json
@@ -34,6 +34,7 @@
   "dependencies": {
     "@hawtio/camel-model": "^3.20.3",
     "@module-federation/utilities": "^1.6.0",
+    "@patternfly/react-charts": "^6.94.19",
     "@patternfly/react-code-editor": "^4.82.113",
     "@patternfly/react-core": "^4.276.8",
     "@patternfly/react-table": "^4.113.0",
--- a/packages/hawtio/src/plugins/camel/CamelContent.tsx
+++ b/packages/hawtio/src/plugins/camel/CamelContent.tsx
@@ -1,5 +1,5 @@
 import { eventService } from '@hawtiosrc/core'
-import { Attributes, JmxContentMBeans, MBeanNode, Operations } from '@hawtiosrc/plugins/shared'
+import { Attributes, Chart, JmxContentMBeans, MBeanNode, Operations } from '@hawtiosrc/plugins/shared'
 import {
   EmptyState,
   EmptyStateIcon,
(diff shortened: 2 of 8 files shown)


tadayosi (changes requested):
With the current edit watches, it's not easy for users to select all and none at once. The closest to the old hawtio that PF 4 has seems to be Dual List Selector. Maybe you can try this for the modal.
https://www.patternfly.org/v4/components/dual-list-selector

Also, can you please squash the trial&error commits into a single one before submitting a pull req?  Those wip commits don't seem to add to our commit history, other than the logs of trials and errors. I wouldn't like to squash them on GitHub UI.

tadayosi:
I'd prefer to keep your trials&errors commit logs local, and do your best to make the commit log clean and meaningful to others whenever you submit it to public.

packages/hawtio/src/ui/page/HawtioPage.tsx
```diff
@@ -72,7 +72,9 @@ export const HawtioPage: React.FunctionComponent = () => {
         defaultManagedSidebarIsOpen={showVerticalNavByDefault}
       >
         {/* Provider for handling selected node shared between the plugins */}
-        <PluginNodeSelectionContext.Provider value={{ selectedNode, setSelectedNode }}>
+        <PluginNodeSelectionContext.Provider
+          value={{ selectedNode, setSelectedNode }}
+        >
           <Routes>
```
phantomjinx, line 78: Cannot see why we are exposing the `setSelectedNode` to the plugins given the `HawtioPage` should be controlling the selection to ensure its common to all plugins.

packages/hawtio/src/plugins/selectionNodeContext.ts
```diff
@@ -1,12 +1,27 @@
...
+  useEffect(() => {
+    ;(async () => {
+      setIsReadingAttributes(true)
+      setSelectedNodeAttributes(await selectionNodeDataService.getAttributesForNode(selectedNode))
+      setIsReadingAttributes(false)
+    })()
+  }, [selectedNode])
+
+  return { selectedNode, setSelectedNode, selectedNodeAttributes, isReadingAttributes }
 }
```
phantomjinx, line 25: Am not sure I am comfortable with attributes being collected here. The selectedNodeContext was at best a hack to get the same selection between the JMX and Camel trees. I don't think it advantageous to build on that hack.

If this is for the chart view only then it would be preferable to have <loading> tags then a useEffect in the Chart view to collect these attributes. This is the workflow that is used for many of the other views, eg. Debug, Trace, Routes, Endpoints.

phantomjinx (changes requested):

packages/hawtio/src/plugins/selectionNodeContext.ts
joshiraez: Ok, I removed the leftover code and left it as it was on main. That change was decided to not move forward and the feat doesn't need any of that. Thanks for spotting it :smile:

packages/hawtio/src/ui/page/HawtioPage.tsx
phantomjinx: Ah ... now found the reason for `setSelectedNode`. Turns out the view trees control it too so my comment was invalid.
Am still unsure what to do about the `PluginNodeSelectionContext.Provider` being exposed outside of the `plugins` but works at the moment until we find something better.

tadayosi:
@joshiraez Any update on this?  Are you still working on it?

joshiraez:
@tadayosi Yes, I was just finishing it. Sorry it took such a long time.

Now what I'm not sure is why the lint fix is failing even though it runs well in local. Spend a long time figuring some of the errors and I just reran it ...

lhein:
@phantomjinx anything else to review / change?

Issue #238, opened 2023-04-14
Camel - No need for the root node `org.apache.camel` in Camel tree view


tadayosi:
I don't think we need this `org.apache.camel` root node in the tree view and we can start from `Camel Contexts`. It's just redundant.

![image](https://user-images.githubusercontent.com/156692/231968039-f0954348-d0ce-43a1-bfb4-770436020be3.png)

tadayosi:
cc @phantomjinx @mmelko

joshiraez:
As far as @phantomjinx told me, from that root context there should be some more nodes for routes and other operations that are coming from other feats. I dont know if I misunderstood that.

On top of that, the other version also has this node and Im unsure if also it could complicate the logic in the tree view as JMX also has that node and it would be akward to just directly have Camel Context directly when the other nodes have their parent folders.

phantomjinx:
Actually, that should not be an issue. The [TreeProcessor](https://github.com/hawtio/hawtio-next/blob/main/packages/hawtio/src/plugins/camel/tree-processor.ts#L40) is responsible for resorting the JMX tree on load and sure enough the `domain` node is retained and so shown in the JMX tree.

However, the Camel tree is displayed by the [CamelTreeView](https://github.com/hawtio/hawtio-next/blob/main/packages/hawtio/src/plugins/camel/CamelTreeView.tsx) component and its display tree is already modified with respect to the possibility of being [filtered](https://github.com/hawtio/hawtio-next/blob/main/packages/hawtio/src/plugins/camel/CamelTreeView.tsx#L32). As such, adjusting that to `tree.getTree().getChildren()[0]` (obviously safety tested) should not be too difficult.

joshiraez:
You have a point. We already are filtering the tree without any problems so there should be no problems to remove that node at all

tadayosi:
Considering #126, we can even remove the root `Camel Contexts` node and directly start from an actual context MBean such as `SampleCamel`.

phantomjinx:
see https://github.com/hawtio/hawtio-next/pull/301

PR #301, opened 2023-05-25
fix(#238): Remove camel domain node from display

phantomjinx:
* Domain node remains available in the original tree yet it is filtered out of the tree displayed in the Camel Tree View

### DIFF
--- a/packages/hawtio/src/plugins/camel/CamelTreeView.tsx
+++ b/packages/hawtio/src/plugins/camel/CamelTreeView.tsx
@@ -1,9 +1,10 @@
 import { MBeanNode, MBeanTree, PluginTreeViewToolbar } from '@hawtiosrc/plugins/shared'
 import { TreeView, TreeViewDataItem } from '@patternfly/react-core'
-import React, { ChangeEvent, useContext, useEffect, useState } from 'react'
+import React, { ChangeEvent, useContext, useEffect, useState, useCallback } from 'react'
 import { useNavigate } from 'react-router-dom'
 import './CamelTreeView.css'
 import { CamelContext } from './context'
+import * as ccs from './camel-content-service'
 import { pluginPath } from './globals'
 
 /**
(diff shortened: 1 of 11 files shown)


phantomjinx:
In light of this comment (https://github.com/hawtio/hawtio-next/issues/238#issuecomment-1525002332), I also removed the "Camel Contexts" node in the Camel Tree View. However, before removing this (which is trivial), it would be necessary to migrate the functions of the Contexts View (below) over to the individual Context View itself.

contexts-view

packages/hawtio/src/plugins/camel/CamelTreeView.tsx
```diff
@@ -29,17 +30,28 @@ enum ExpansionValue {
...
 
+  const deriveTree = useCallback((): MBeanNode[] => {
+    const t = tree.getTree()
+    if (t.length === 0) return t
+
+    if (!ccs.isDomainNode(t[0])) return t
+
+    // Exclude the domain node from the tree display
+    return t[0].getChildren()
+  }, [tree])
```
tadayosi, line 43: Why do we need this derivation here instead of doing so inside the `useContext`?  Is there any reason we need to bring the skipped root node into the react component?

tadayosi (approved):
LGTM, especially I like the React component test!

tadayosi:
> In light of this comment ([#238 (comment)](https://github.com/hawtio/hawtio-next/issues/238#issuecomment-1525002332)), I also removed the "Camel Contexts" node in the Camel Tree View. However, before removing this (which is trivial), it would be necessary to migrate the functions of the Contexts View (below) over to the individual Context View itself.

What about making the "Camel Contexts" view and functionality as the default view when there's no selected node in the view?  Then we can simply move the functions in the Contexts view to the CamelContent.tsx.

Also, currently each Camel context node is represented as Cube icon, but it should be the Camel icon too, especially when we skip the "Camel Contexts" node.

phantomjinx:
> In light of this comment ([#238 (comment)](https://github.com/hawtio/hawtio-next/issues/238#issuecomment-1525002332)), I also removed the "Camel Contexts" node in the Camel Tree View. However, before removing this (which is trivial), it would be necessary to migrate the functions of the Contexts View (below) over to the individual Context View itself.
> 
> contexts-view

A rethink ked to the start/suspend toolbar being available on the context navigation view as well so `contexts` can be removed as well.

Issue #286, opened 2023-05-14
Camel plugin - group routes by "routeGroup"


tadayosi:
The original issue: https://github.com/hawtio/hawtio/issues/2648

This is a great feature to have in hawtio-next. Let's implement it for the next milestone.

phantomjinx:
PR -> https://github.com/hawtio/hawtio-next/pull/309

PR #309, opened 2023-05-31
feat(#286): support grouping of routes if group ids have been specified

phantomjinx:
* routes-service.ts
 * Add a 'group' property if the xml specifies the attribute
 * Makes transformXml asynchronous state explicit by using async & await

* tree-processor.ts
 * camelTreeProcessor function is already asynchronous so await conclusion of shredding xml to routes then call groupRoutes function to maybe group routes if they are xml group ids specified

* node.ts
 * Adds removeChild function to work with adopt. Ensures that a child being adopted by 1 parent is removed from the previous parent

* Adds tests for groups in tree processing


tadayosi (approved):
Looks great!  Can we have a screenshot of what this pull req would provide for grouped routes?

phantomjinx:
> Looks great! Can we have a screenshot of what this pull req would provide for grouped routes?

hawtio-grouped-routes

PR #313, opened 2023-06-01
fix(312): have unique ids for folders and items in JMXTreeView

### LINKED ISSUE #312: Tree view highlights both folders and items if they share the same name
If a folder and an item in the same hierarchy share name then both items get highlighted, and React complains about elements not having unique keys.
image
Same happens for XNIO-1:
image

mmuzikar:
Fixes #312


tadayosi (changes requested):
Thanks for the pull req, but I don't think it's the right way to fix the original issue.

I think we should rather fix the way the mbean tree is created so that the  we only have a single xnio-1 node in this case.

tadayosi (approved):
Changed my opinion. It looks good to me. Thanks.

tadayosi:
@mmuzikar You forgot to run `yarn format:fix`. Please update your pull req.

tadayosi:
@mmuzikar Tests are failing because they rely on the generated node IDs. Please update those tests and make them pass before merging.

tadayosi:
thanks!

PR #323, opened 2023-06-07
fix(#318): fix synchronisation with the populating the tree

### LINKED ISSUE #318: [Bug] No Camel Contexts found when a new Camel endpoint is added
When a new Camel endpoint is added, it redirects to a page with `No Camel Contexts found` message

Then clicking again on Camel -> endpoints I can see the newly added endpoint

https://github.com/hawtio/hawtio-next/assets/49488040/070860db-79ce-4518-b3ac-2489e0812ce3

**UPD:**
The same behaviour is when Start/Stop routes on Routes page and when Start/Stop a route from Operations tab - redirects to `No Camel Contexts found` page at `http://localhost:10001/actuator/hawtio/camel`

mmelko:
fixes #318.

Main cause was to introduce asynchronous element into `camelTreeProcessor`, so I updated the populating tree to be async.

### DIFF
--- a/packages/hawtio/src/plugins/shared/tree/tree.ts
+++ b/packages/hawtio/src/plugins/shared/tree/tree.ts
@@ -15,9 +15,9 @@ export class MBeanTree {
     return new MBeanTree(id)
   }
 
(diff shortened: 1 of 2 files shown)


tadayosi (approved):
awesome!

PR #328, opened 2023-06-07
fix: Add dummy success callbacks in dummy-jolokia methods

mmelko:
If nothing was found, Promise was never resolved which led to block of tree processing and hawtio coudln't run. @tadayosi does this work for you? I'm not sure if it's possible to use some 'onError' callback so I fixed it this way.

### DIFF
--- a/packages/hawtio/src/plugins/connect/jolokia-service.ts
+++ b/packages/hawtio/src/plugins/connect/jolokia-service.ts
@@ -31,7 +31,7 @@ import Jolokia, {
 } from 'jolokia.js'
 import 'jolokia.js/jolokia-simple'
 import $ from 'jquery'
-import { func, is, object } from 'superstruct'
+import { func, is, object, unknown } from 'superstruct'
 import { connectService, PARAM_KEY_CONNECTION } from './connect-service'
 import { log } from './globals'
 import { basicAuthHeaderValue } from '@hawtiosrc/util/http'
@@ -490,6 +490,7 @@ class DummyJolokia implements IJolokia {
   }
 
   getAttribute(mbean: string, attribute: string, path?: string | ISimpleOptions, opts?: ISimpleOptions) {
+    opts?.success?.({})
     return null
   }
   setAttribute(
@@ -500,6 +501,7 @@ class DummyJolokia implements IJolokia {
     opts?: ISimpleOptions,
   ) {
     /* no-op */
+    opts?.success?.(unknown)
   }
 
   execute(mbean: string, operation: string, ...args: unknown[]) {
@@ -507,6 +509,7 @@ class DummyJolokia implements IJolokia {
     return null
   }
   search(mBeanPattern: string, opts?: ISearchOptions) {
+    opts?.success?.([])
     return null
   }
   list(path: string, opts?: IListOptions) {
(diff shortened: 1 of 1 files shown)


tadayosi (changes requested):
Good catch!  As you guessed, there is a callback on error and we should use it instead.

Like this function, you should create `onSearchSuccessAndError` there and use it from the jolokia service:
https://github.com/hawtio/hawtio-next/blob/main/packages/hawtio/src/util/jolokia.ts#L30-L36

tadayosi:
> If nothing was found, Promise was never resolved which led to block of tree processing and hawtio coudln't run.

@mmelko Hmm, thinking it a bit further, I wonder if the usage of error callback really solves the issue. It's strange that no callback is invoked when Jolokia found nothing at search operation.

But still this might not be good coding, as using the return value from Jolokia should force it to use sync mode, thus blocking the process there.
```
const resp: string[] | null = jolokia.search(mbeanPattern)
```

mmelko:
@tadayosi maybe it's because it's returning null instead of empty object. So I will try with the error callback and see what's happening there.

mmelko:
After discussion with @tadayosi we found out that the real problem is how search function in dummy-jolokia implementation is done. So I changed it accordingly.

tadayosi (approved):
Thanks!

tadayosi:
@mmelko As we discussed, I think it's still important to implement the on-error callback for the jolokia service's search method so that it will be resolved in any case. Would you do that in the same pull req or in a separate one?

mmelko:
> @mmelko As we discussed, I think it's still important to implement the on-error callback for the jolokia service's search method (as you were originally trying to do) so that it will be resolved in any case. Would you do that in the same pull req or in a separate one?

done.

packages/hawtio/src/plugins/connect/jolokia-service.ts
```diff
@@ -500,20 +503,23 @@ class DummyJolokia implements IJolokia {
     opts?: ISimpleOptions,
   ) {
     /* no-op */
+    opts?.success?.(unknown)
```
tadayosi, line 506: it's not "no-op" anymore :-)

Also I don't know if it's the correct usage of superstruct `unknown` here. Can't you rather simply do `opts?.success?.({})`?

packages/hawtio/src/plugins/connect/jolokia-service.ts
```diff
@@ -406,10 +405,13 @@ class JolokiaService implements IJolokiaService {
...
   async search(mbeanPattern: string): Promise<string[]> {
     const jolokia = await this.jolokia
-    return new Promise(resolve => {
+    return new Promise((resolve, reject) => {
       jolokia.search(
         mbeanPattern,
-        onSearchSuccess((response: string[]) => resolve(response)),
+        onSimpleSuccessAndError(
+          response => resolve(response as string[]),
+          (response: IErrorResponse) => reject(response.stacktrace || response.error),
```
tadayosi, line 413: I would avoid using `reject` anywhere in the hawtio-next, as it would be interpreted as an error thrown which hawtio couldn't handle. I'd rather output `log.error(<some error message>)` and resolve `[] as string[]`.

tadayosi (changes requested):

tadayosi (approved):
looks great. thanks!

PR #331, opened 2023-06-08
Camel Plugin: Rest Services View

phantomjinx:


packages/hawtio/src/plugins/camel/rest-services/rest-services.service.ts
tadayosi, line 1: We don't use this Angular-style service name suffix `.service.ts`. We can instead name it `rest-services-service.ts`.

tadayosi (approved):
thanks!

mmelko:
looks good to me 👍

PR #349, opened 2023-06-14
feat(camel-plugin): Add 'Endpoint (in/out)' view into the camel plugin

### LINKED ISSUE #282: Camel - Endpoint Statistics view
`/camel/endpoints-statistics`

Screenshot - TBD. Let's investigate whether we need this view for Camel 3+ or not.

mmelko:
Add component test to cover rendering EndpointStats component.
closes #282 
Screenshot 2023-06-14 at 15 18 31
Filtering:
Screenshot 2023-06-14 at 15 18 48

### DIFF
--- a/packages/hawtio/src/plugins/camel/endpoints/endpoints-service.ts
+++ b/packages/hawtio/src/plugins/camel/endpoints/endpoints-service.ts
@@ -3,6 +3,7 @@ import { jolokiaService } from '@hawtiosrc/plugins/connect'
 import * as schema from '@hawtio/camel-model'
 import { MBeanNode, workspace } from '@hawtiosrc/plugins/shared'
 import * as ccs from '../camel-content-service'
+import { getDefaultRuntimeEndpointRegistry } from '../camel-content-service'
 import { contextNodeType, endpointsType, log } from '../globals'
 import { isObject } from '@hawtiosrc/util/objects'
 import { parseXML } from '@hawtiosrc/util/xml'
@@ -11,6 +12,17 @@ export type Endpoint = {
   uri: string
   state: string
   mbean: string
+}
+
+export type EndpointStatistics = {
+  hits: number
+  routeId: string
+  static: boolean
+  index: number
+  dynamic: boolean
+  url: string
+  direction: string
+  [key: string]: string | boolean | number
 }
 
 export type MessageData = {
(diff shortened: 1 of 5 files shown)


packages/hawtio/src/plugins/camel/endpoints/endpoints-service.ts
```diff
@@ -280,3 +292,13 @@ function parseMessagesFromXml(pDoc: XMLDocument): MessageData[] {
 
   return messagesData
 }
+
+export async function getEndpointStatistics(node: MBeanNode) {
+  let stats: EndpointStatistics[] = []
+  const registry = getDefaultRuntimeEndpointRegistry(node)
+  if (registry) {
+    const res = await jolokiaService.execute(registry.objectName ?? '', 'endpointStatistics()')
```
tadayosi, line 300: if `objectName` is absent it doesn't make sense to invoke the method. Rather shouldn't we guard it based on its presence and if it's absent something else should be done, like throwing or logging error?

tadayosi (approved):
Overall, looks great. There's one minor comment.

packages/hawtio/src/plugins/camel/endpoints/endpoints-service.ts
mmelko: updated

packages/hawtio/src/plugins/camel/endpoints/endpoints-service.ts
```diff
@@ -280,3 +292,17 @@ function parseMessagesFromXml(pDoc: XMLDocument): MessageData[] {
...
+export async function getEndpointStatistics(node: MBeanNode) {
+  let stats: EndpointStatistics[] = []
+  const registry = getDefaultRuntimeEndpointRegistry(node)
+  if (registry) {
+    if (registry.objectName) {
+      const res = await jolokiaService.execute(registry.objectName ?? '', 'endpointStatistics()')
+      stats = Object.values(res as { [key: string]: EndpointStatistics })
+    } else {
+      log.error('Error with the retrieving registry')
+    }
```
tadayosi, line 305: so, now the code just ignores if `registry` is absent. Is it what we expect?

Also, now you already guard with `if (regitry.objectName)` so `registry.objectName ?? ''` doesn't make sense as you can assume it's present.

tadayosi (approved):

PR #361, opened 2023-06-19
fix(connect): check negative numbers in input fields for Connect Preferences

### LINKED ISSUE #337: Need to check for negative numbers in input fields
It turned out that it is possible to input negative numeric values in Preferences and moreover, it gets applied and the Hawtio crashes. Checked it with 3.x branch of hawtio standalone.

Screenshot 2023-06-12 at 14 13 25

It's understandable that a regular user won't type negative numeric values. However, a typo and/or mis-click may happen which will cause the Hawtio crash.

To fix it you can open a dev tools panel in a browser, open Application tab and clean Local Storage

tadayosi:
Fix #337

@jsolovjo In case you are interested in how to fix it.

### DIFF
--- a/packages/hawtio/src/plugins/connect/ConnectPreferences.tsx
+++ b/packages/hawtio/src/plugins/connect/ConnectPreferences.tsx
@@ -5,6 +5,7 @@ import {
   Checkbox,
   Form,
   FormGroup,
+  FormGroupProps,
   FormSection,
   Modal,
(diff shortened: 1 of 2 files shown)


jsolovjo (approved):
Thank you for the fix and letting me know. Sure, I am interested how it got fixed and I am a real beginner in TypeScript, so thank you :)

Issue #352, opened 2023-06-15
Should we add sorting by name, value, attributes, etc. to tables how it was in older Hawtio?


jsolovjo:
In new Hawtio there is only one table which has sorting feature
<img width="1710" alt="Screenshot 2023-06-15 at 11 04 26" src="https://github.com/hawtio/hawtio-next/assets/49488040/6f28992a-e156-44a9-8de5-59c6e44b3953">

Other tables like below do not have such a feature
<img width="1712" alt="Screenshot 2023-06-15 at 11 04 52" src="https://github.com/hawtio/hawtio-next/assets/49488040/1826fcbc-cf7a-4b35-a673-3c35f0333fec">

tadayosi:
Yeah, it sounds like a good idea. Let's follow the PF 4 desing guidelines to add sorting features to tables where they are relevant.
https://www.patternfly.org/v4/components/table/design-guidelines

mmelko:
@jsolovjo your example(attributes in particular endpoint) isn't sortable even in legacy hawtio. So far I've found these tables that are sortable in legacy and not in next:
context tabs:
- Endpoints (in/out)
- TypeConverters
Routes tab:
 - routes
 
so I'd start with these at first

jsolovjo:
> @jsolovjo your example(attributes in particular endpoint) isn't sortable even in legacy hawtio. So far I've found these tables that are sortable in legacy and not in next: context tabs:
> 
> * Endpoints (in/out)
> * TypeConverters
>   Routes tab:
> * routes
> 
> so I'd start with these at first

@mmelko true, my bad - I provided poor example. Some tables have sorting feature and some don't have it in older Hawtio and I didn't check it properly when choosing the example, sorry

PR #362, opened 2023-06-19
feat(endpoints-stats): make table sortable

mmelko:
Makes table sortable as it was in the legacy hawtio.
Related to #352


tadayosi (approved):
Wow, simple fix!

Issue #363, opened 2023-06-19
Operations - Expanded view of operation is reset every once in a while


joshiraez:
Noticed this while testing the other tickets after rebasing. The operations view gets reset every few in a while.  Has a similar solution to #273 so I'll be including the fix there
![Peek 2023-06-19 16-51](https://github.com/hawtio/hawtio-next/assets/17336236/e5fdc06f-7aab-4086-80a6-b101e5dc8241)

tadayosi:
Duplicate of #346. This should be already fixed with #360.

Issue #278, opened 2023-05-08
JMX/Camel - Implement copy method name & Jolokia URL actions


tadayosi:
We still haven't implemented the following actions in the JMX Operations view:
- Copy method name
- Copy Jolokia URL

![image](https://user-images.githubusercontent.com/156692/236746228-9bf495d8-501a-4de7-b0e2-6bd6ea04fa9a.png)

Related to but not a duplicate of #273.

tadayosi:
cc @joshiraez

joshiraez:
I'll be picking this up after finishing the chart component

tadayosi:
We'd like to have it in 0.3.0 so I'll take it. @joshiraez Please focus on #195 instead next, if you don't have other important tasks after Chart.

joshiraez:
Hey Tada, I'm halfway here and I realized is for 0.3.0 and we need it for
the end of June.

Do you mind letting me finish it please?

Issue #346, opened 2023-06-13
An automatic request distracts a Camel operation expanded view


jsolovjo:
The recording how the issue occurs

https://github.com/hawtio/hawtio-next/assets/49488040/b74ef472-6923-43ad-8d72-a665be78d6e2

The auto-request which causes the issue with Operation expanded view.
```
[...]

PR #364, opened 2023-06-19
feat(#273,#363,#278): Implements additional copy buttons, bugfixes

### LINKED ISSUE #273: Bug: Button for a drop-down menu is not working in Camel Operations
The button (3 dots) doesn't react on clicking. I assume the click on it should show a drop-down menu.

Screenshot 2023-05-04 at 13 53 54

joshiraez:
Implements #278 
Fixes #273 #363 

While fixing this I noticed I think there is some kind of bug in the contexts. Pages are being refreshed endlessly even when their state is not changed. Should I file a bug for this? I'm not sure if it's expected because the tree is constantly being refreshed.

### DIFF
--- a/packages/hawtio/src/plugins/shared/operations/OperationForm.tsx
+++ b/packages/hawtio/src/plugins/shared/operations/OperationForm.tsx
@@ -1,5 +1,7 @@
 import {
   ActionGroup,
+  Alert,
+  AlertGroup,
   Button,
   Checkbox,
   ClipboardCopy,
@@ -21,7 +23,7 @@ import {
   TextInput,
   Title,
 } from '@patternfly/react-core'
-import React, { useContext, useState } from 'react'
+import React, { useContext, useMemo, useState } from 'react'
 import { Operation } from './operation'
 import { operationService } from './operation-service'
 import { PluginNodeSelectionContext } from '@hawtiosrc/plugins/selectionNodeContext'
@@ -170,12 +172,7 @@ export const OperationForm: React.FunctionComponent<OperationFormProps> = props 
   const { name, operation } = props
   const [isExpanded, setIsExpanded] = useState(false)
   const [isDropdownOpen, setIsDropdownOpen] = useState(false)
-
-  if (!selectedNode || !selectedNode.objectName || !selectedNode.mbean) {
-    return null
-  }
-
-  const { objectName } = selectedNode
+  const [currentAlerts, setCurrentAlerts] = useState<React.ReactNode[]>([])
 
   const handleToggle = () => {
     setIsExpanded(!isExpanded)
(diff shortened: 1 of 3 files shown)


tadayosi (changes requested):
#363 should be already fixed with #360. Please revert the part which is supposed to fix #363.

IMO, `useMemo` or `useCallback` should be used only after we try our best to modularise React components with minimum states. The first approach should be always splitting components into parts so that where rerendering happens is narrowed.

packages/hawtio/src/plugins/shared/operations/OperationForm.tsx
```diff
@@ -185,14 +182,57 @@ export const OperationForm: React.FunctionComponent<OperationFormProps> = props
     setIsDropdownOpen(!isDropdownOpen)
   }
 
+  const copyToClipboard = (text: string) => {
+    navigator.clipboard.writeText(text)
+    setCurrentAlerts(prev => [
+      ...prev,
+      <Alert key={`{alert-${prev.length + 1}`} isLiveRegion timeout={2000} title='Copied to clipboard' />,
```
tadayosi, line 189: Default alert timeout we use throughout hawtio-next is 8s (= 8000). Unless it was the behaviour from the old Hawtio, we should apply the same timeout duration here as well.

joshiraez:
Oh great that it was already fixed. I also see that https://github.com/hawtio/hawtio-next/issues/346 you already noticed the frequent refreshes.

Ok, I'll refactor it to a different component instead of using useMemo. Let me a moment

packages/hawtio/src/plugins/shared/operations/OperationForm.tsx
joshiraez: Alright. I checked the old-hawtio and 8 seconds seemed too long but I don't think there is any strong reason to not go with the default. It's patternfly's default as well so it'll make the code cleaner just removing that prop :smile:

joshiraez:
The diff it's hard to understand. The fix has been to just move the OperationActions component outside of the component where it was, group it together with its own state and the new implemented functions in #278

packages/hawtio/src/plugins/shared/operations/OperationForm.tsx
```diff
@@ -165,54 +167,47 @@ const OperationExecuteForm: React.FunctionComponent<{
...
+  }
 
-  const OperationActions = () => (
+  return (
     <DataListAction
       id={`operation-actions-${name}`}
       aria-label={`operation actions ${name}`}
       aria-labelledby={`${name} operation-actions-${name}`}
     >
+      <AlertGroup isToast isLiveRegion>
```
joshiraez, line 208: I just saw that we have an HawtIONotification element in the main page for this kind of toast alerts. Should I refactor this to use that better?

tadayosi (approved):

PR #370, opened 2023-06-20
fix(#369):Fixes watches to modify attributes to watch correctly

### LINKED ISSUE #369: Chart - Edit watches not removing charts
Putting it here for traceability.

Peek 2023-06-20 10-59

I think the problem lies in one of the later fixes I did to ensure that children keys were not being duplicated but it seems I didn't test it thoroughly. I'll submit a fix shortly

joshiraez:
Fixes #369 

Peek 2023-06-20 11-40

### DIFF
--- a/packages/hawtio/src/plugins/shared/chart/WatchableAttributesForm.tsx
+++ b/packages/hawtio/src/plugins/shared/chart/WatchableAttributesForm.tsx
@@ -18,7 +18,7 @@ export const WatchableAttributesForm = ({
   // Because the key needs to be unique across all trees (or else there is a bug where the check selection applies to
   // any element in any tree with the same key), I had to change the keys and no longer can use that prop for storing the key
   // to the original attributes to watch.
-  // There is a way around this building a dictionary storing attribute names - keys, but it seemed to overengineered
+  // There is a way around this building a dictionary storing attribute names - keys, but it seemed too overengineered
   // when the text is currently the same attribute name as the key in attributes to watch
 
   return (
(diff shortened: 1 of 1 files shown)


packages/hawtio/src/plugins/shared/chart/WatchableAttributesForm.tsx
```diff
@@ -67,17 +73,19 @@ export const WatchableAttributesForm = ({
...
-              newWatches[newUnwatchedAttribute.id] = {
-                ...(newWatches[newUnwatchedAttribute.id] || {}),
-                ...Object.fromEntries(newUnwatchedAttribute.children?.map(({ id }) => [id, false]) || []),
+              newWatches[newUnwatchedAttribute.text] = {
+                ...(newWatches[newUnwatchedAttribute.text] || {}),
+                ...Object.fromEntries(newUnwatchedAttribute.children?.map(({ text }) => [text, false]) || []),
               }
             })
 
+            console.log(newWatches)
```
tadayosi, line 87: don't leave `console.log`. if you want to leave some debug logging, use `log.debug` from `globals.ts`

tadayosi (changes requested):

packages/hawtio/src/plugins/shared/chart/WatchableAttributesForm.tsx
joshiraez: So sorry. Removing this right away.

I'm going to build a custom lint rule at this rate so it tells me of any console.logs. It happens too much

Issue #262, opened 2023-04-28
Camel tree nodes overlay on another section when a smaller screen resolution is used


tadayosi:
https://github.com/hawtio/hawtio/issues/2824

When a smaller screen resolution is used, Camel tree items/nodes overlay on another UI section

![image](https://user-images.githubusercontent.com/49488040/234301055-e7ae95cd-e94e-480f-8b79-a47f13e55db8.png)

tadayosi:
Actually, word wrapping works with the Split.js. Overlays happen when it's a long single word (no spaces in between).

What we need to do would be to shorten the words until it fits within the split frame when otherwise overlay might happen, e.g. `a-mbean-with-long-name` -> `a-mbean-with-...`

joshiraez:
Closing this as it was merged :)

PR #372, opened 2023-06-21
feat(#262): Prevent overlay on smaller screens

joshiraez:
Implements #262 . Implemented in Camel and JMX view

Before
overflow-1

After
overflow-2

Also working in JMX tree
image

Although one side effect is that on very small screens the text is outright hidden,
image

but the alternative is to have it all outside which looks bad. 

image

I feel like in that case the user should just adjust the slider to increase the space.


tadayosi (approved):
That's awesome!!!  Thanks for the fix!

PR #373, opened 2023-06-21
feat(Routes): Add sorting functionality to Routes table

mmelko:
This PR adds sorting to the Route table and also add a new component tests that are covering routes rendering and sorting


packages/hawtio/src/util/utils.ts
tadayosi, line 1: Let's not name util module generic like `utils`. We already categorise utils with different names under `/util/`.

And we already have a sorter for strings. I wonder if we can reuse or extend it in some way.
https://github.com/hawtio/hawtio-next/blob/main/packages/hawtio/src/util/strings.ts#L87

tadayosi (approved):
thanks!

PR #374, opened 2023-06-21
fix(#268): Removes overflow auto and sets to inherit so selects in card can go over the div boundary

### LINKED ISSUE #268: Enhancement: Improve scrolling when adding a new endpoint from Data
There is inconvenient scrolling when adding a new endpoint from Data and selecting a component.

Perhaps, the UI section should be bigger and/or scrolling should be implement differently.

Check the screen recordings. 

https://user-images.githubusercontent.com/49488040/235914102-bbcbe226-ca2d-4b88-8810-309c6444458d.mov

joshiraez:
Fixes #268 

image

@phantomjinx Do you know why you hard-set the value to auto on the css? I checked all the camel pages to see if the change breaks something but it'd be great to know if you remember what was it so I can double check.

In any case, using inherit should be extra safe instead of removing the rule altogether

### DIFF
--- a/packages/hawtio/src/plugins/camel/CamelContent.css
+++ b/packages/hawtio/src/plugins/camel/CamelContent.css
@@ -1,5 +1,9 @@
 #camel-content-main > article {
-  overflow: inherit;
+  overflow-y: inherit;
+}
+
+#camel-content-main .pf-c-page__main {
+  overflow-x: auto;
 }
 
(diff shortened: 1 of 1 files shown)


joshiraez:
Ok I just talked with @phantomjinx and there was one thing I broke with the change. Horizontal overflows on smaller screens don't work (because the Page component itself has it set to hidden). I added that as well.

tadayosi (approved):

PR #376, opened 2023-06-21
fix(proxy): aligns the dev server proxy url with the real backend

phantomjinx:
* The proxy url in hawtio-backend includes the context in the form: host:port/hawtio/proxy/

* The hawtio-backend-middleware provides an express server that makes the backend available via the '/proxy' url

* Rather than trying to redo the hawtio-backend-middleware, aligns the dev server with the hawtio-backend by setting up a 'proxy' to rewrite 'hawtio/proxy' and point to '/proxy'

* connect-service.ts
 * Adds the base path to the proxy url creation

### DIFF
--- a/packages/hawtio/src/plugins/connect/connect-service.ts
+++ b/packages/hawtio/src/plugins/connect/connect-service.ts
@@ -163,12 +163,12 @@ class ConnectService implements IConnectService {
       return connection.jolokiaUrl
     }
 
-    // TODO: Better handling of doc base and proxy URL construction
     const url = joinPaths(
+      hawtio.getBasePath() ?? '',
       '/proxy',
-      connection.scheme || 'http',
-      connection.host || 'localhost',
-      String(connection.port || 80),
+      connection.scheme ?? 'http',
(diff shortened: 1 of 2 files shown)


packages/hawtio/src/plugins/connect/connect-service.ts
```diff
@@ -165,6 +165,7 @@ class ConnectService implements IConnectService {
 
     // TODO: Better handling of doc base and proxy URL construction
```
tadayosi, line 166: I think we can remove this line now, as we handle the base path correctly now.

packages/hawtio/src/plugins/connect/connect-service.ts
```diff
@@ -165,6 +165,7 @@ class ConnectService implements IConnectService {
...
     const url = joinPaths(
+      hawtio.getBasePath() || '',
```
tadayosi, line 168: Let's use `??` instead of `||`. Please also update the following lines as well.

tadayosi (approved):
A few minor comments but looks good overall.

Issue #258, opened 2023-04-27
Add test for checking html component tree has consistent ids


tadayosi:
For downstream E2E testing, it is important that the generated HTML components have some meaningful ids hawtio/hawtio#2826. So, ideally we should have some test that checks whether we've assigned some ids in a consistent way within hawtio-next project. I wonder if we can write such a test using React Testing Library.

tadayosi:
cc @joshiraez

joshiraez:
I can take care of this. Assigning myself.

I was wondering what is the behaviour we should have after #238 is added. I will do some tests for the current behaviour but what happens with the removed nodes? Should we remove the nodes directly from the tree and ignore them while in id composition? (The tests should be fairly easy still to change, just wondering)

joshiraez:
I'm now on this and after that i'll start right away with the 0.4.0 tickets (most likely the Logs plugin)

PR #378, opened 2023-06-22
test(#258),fix(#377): Tree ids test and fixes

### LINKED ISSUE #377: [Bug] XNIO-2 node under XNIO folder doesn't have correct IDs
image

As far as I have seen it is the only outlier. My best guess is that somehow the XNIO-2 folder it's created, then it adopts its children and because the adopt() function doesn't regenerate the ids for the tree, it doesn't update.

I'm currently checking how the JMX tree is created in order to find why the ids didn't update for that particular case.

joshiraez:
Tests #258 
Fixes #377 

I saw there were some assertions done by @mmuzikar 3 weeks ago but there was not a fully dedicated tests for the IDs.
On the process I discovered #377 . I created an issue to keep track of it and did two fixes to ensure that the outlier behaviour doesn't repeat. Although I'm not exactly sure if that would still make it exhaustive for all tree constructions, but now that we have a test we can leave it as is and revisit it later if the test starts discovering any issues again.

### DIFF
--- a/packages/hawtio/src/plugins/shared/tree/node.ts
+++ b/packages/hawtio/src/plugins/shared/tree/node.ts
@@ -73,6 +73,7 @@ export class MBeanNode implements TreeViewDataItem {
       this.icon = Icons.mbean
     }
 
+    if (this.parent) this.parent.id = this.parent.generateId(true)
     this.id = this.generateId(folder)
   }
 
--- a/packages/hawtio/src/plugins/shared/tree/node.ts
+++ b/packages/hawtio/src/plugins/shared/tree/node.ts
@@ -387,6 +387,9 @@ export class MBeanNode implements TreeViewDataItem {
 
     child.parent = this
     this.children.push(child)
+
+    //Refresh ids on adoption
+    this.initId(true)
   }
 
   /**
--- a/packages/hawtio/src/plugins/camel/routes-service.test.ts
+++ b/packages/hawtio/src/plugins/camel/routes-service.test.ts
@@ -91,10 +91,10 @@ describe('routes-service', () => {
     }
 
     const expected: MBeanAttr[] = [
-      { id: 'from', name: 'from' },
-      { id: 'setBody', name: 'setBody' },
-      { id: 'to', name: 'to' },
-      { id: 'to', name: 'to' },
+      { id: 'sample-camel-1-folder-routes-2-folder-simple-folder-from', name: 'from' },
+      { id: 'sample-camel-1-folder-routes-2-folder-simple-folder-setBody', name: 'setBody' },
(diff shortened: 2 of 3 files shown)


packages/hawtio/src/plugins/shared/tree/node.ts
```diff
@@ -73,6 +73,7 @@ export class MBeanNode implements TreeViewDataItem {
       this.icon = Icons.mbean
     }
 
+    if (this.parent) this.parent.id = this.parent.generateId(true)
```
tadayosi, line 76: why do we need to update the parent id?  I think id should be updated in its own and children shouldn't do that.

I think there's more subtle issue in this bug. At some point in the past, we decided to allow two nodes (substantial mbean and mere folder) with the same path up to themselves (e.g. `org.xnio:type=Xnio,provider="nio",worker="XNIO-1"`). So we decided to add suffix `-folder` to id so that the mbean node and folder node are handled separately. But looking at `org.xnio:type=Xnio,provider="nio",worker="XNIO-2"`, it seems it's rather behaving as mbean and folder in a single node, which is not something we expect. We'd need to figure out why and fix it so that it can behave the same way as XNIO-1.
image

tadayosi (changes requested):

packages/hawtio/src/plugins/shared/tree/node.ts
tadayosi: This one: #313

tadayosi:
#377 seems to be a bit complicated to fix quickly, while the part of your pull req that fixes #258 looks fine. What about splitting the pull req and only addressing #258 first?

packages/hawtio/src/plugins/shared/tree/node.ts
joshiraez: Ok. From the code I thought a node was a folder if it has children, as there are several lines that to know if something is a folder it just checks if it has children.
MMMMM, I think what it'd be best would be to do a refactor of the tree class. As I have seen also there are several ways to add nodes to a folder (camel tree uses adopt. Other places I think just create MBean then assign the children. From what I have seen, there is no way to create a folder mbean with (...children) as its parameter which would have the desired behaviour you want.

joshiraez:
Yes, I was going to mention that. #377 will need to at least have a complete revisit to the tree. Let me split them up, leave a guard on the test for the outlier case (and the mock case because right now it was failing too as I had to reassing ids, check  so it can pass, check https://github.com/hawtio/hawtio-next/pull/378/commits/262d21adfc8a31b98a473e615b1c6a4cc9ca9ec7) and after I fix #377 I'll revisit the test and modify them as necessary

I'll submit a different PR

PR #381, opened 2023-06-23
feat(ForwardMessage): Add typeahead suggestions to forward message input

mmelko:
Added suggestions to forward message text input as in legacy hawtio: 
Screenshot 2023-06-23 at 15 05 18
Screenshot 2023-06-23 at 15 05 42


joshiraez:
Looks great!

tadayosi (approved):
Looks great! So, does that mean the schema information for the suggestions is already provided from the camel-model and we didn't need to enhance anything around the camel catalog?

mmelko:
@tadayosi  this is different. Here  I'm fetching list of endpoints from the context. So for message-headers it I still has to investigate how to do it properly. I did this at first at least to figure out the React stuff.

Thanks for the positive feedback :)

PR #388, opened 2023-06-28
Feat: Add  suggestion's for a header name input

mmelko:
Screenshot 2023-06-28 at 08 59 35
also: 
- manually created exchange headers definitions into the `camel-model` 
- reused input with suggestions from the `BrowseMessages` component

### DIFF
--- a/packages/hawtio/src/plugins/camel/endpoints/BrowseMessages.tsx
+++ b/packages/hawtio/src/plugins/camel/endpoints/BrowseMessages.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useEffect, useRef, useState } from 'react'
+import React, { useContext, useEffect, useState } from 'react'
 import { CamelContext } from '@hawtiosrc/plugins/camel/context'
 import {
   Bullseye,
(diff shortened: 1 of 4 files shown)


packages/camel-model/src/exchange-headers-camel-model.json
tadayosi, line 1: Maybe it's better to put the file under the hawtio package until the camel-model package is enabled to autogenerate it, because otherwise we'd need to publish another camel-model version `3.20.6-1` just for the change.

If it's used only from the endpoint view, you can put it under `plugins/camel/endpoint` directly.

packages/hawtio/src/plugins/camel/endpoints/BrowseMessages.tsx
```diff
@@ -40,13 +40,11 @@ import {
...
 import { Position } from 'reactflow'
 import { MBeanNode } from '@hawtiosrc/plugins'
-
-const ForwardMessagesComponent: React.FunctionComponent<{
-  onForwardMessages: (uri: string, message?: MessageData) => void
-  currentMessage?: MessageData
-  endpoints: string[]
-}> = ({ onForwardMessages, currentMessage, endpoints }) => {
-  const [uri, setUri] = useState('')
+export const InputWithSuggestionsComponent: React.FunctionComponent<{
```
tadayosi, line 43: If it's used from at least two places, why not extract it into a distinct file `InputWithSuggestions.tsx`?  And I think the name suffixes `...Component` is generally redundant. We don't use those suffixes in other places.

mmelko:
@tadayosi thanks for your feedback. I updated the PR according your suggestions

Issue #368, opened 2023-06-20
Camel - "Warning: Encountered two children with the same key" at routes node with multiple endpoints of the same kind


tadayosi:
When I look at the Camel plugin and expand the `routes` folder, I can see there are two `to` endpoint nodes under `cron` in my example. Opening the browser console, I see the following error:

![image](https://github.com/hawtio/hawtio-next/assets/156692/67f59518-4fb3-40b9-9524-dec8355e243c)

> Warning: Encountered two children with the same key, `org.apache.camel-folder-CamelContexts-folder-SampleCamel-folder-routes-folder-cron-folder-to`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.

```
    at ul
    at TreeViewList (http://localhost:3000/hawtio/static/js/vendors-node_modules_patternfly_react-charts_dist_esm_index_js-node_modules_patternfly_react--46edeb.chunk.js:45861:7)
    at TreeView (http://localhost:3000/hawtio/static/js/vendors-node_modules_patternfly_react-charts_dist_esm_index_js-node_modules_patternfly_react--46edeb.chunk.js:45757:7)
    at li
    at TreeViewListItemBase (http://localhost:3000/hawtio/static/js/vendors-node_modules_patternfly_react-charts_dist_esm_index_js-node_modules_patternfly_react--46edeb.chunk.js:45901:5)
    at ul
    at TreeViewList (http://localhost:3000/hawtio/static/js/vendors-node_modules_patternfly_react-charts_dist_esm_index_js-node_modules_patternfly_react--46edeb.chunk.js:45861:7)
    at TreeView (http://localhost:3000/hawtio/static/js/vendors-node_modules_patternfly_react-charts_dist_esm_index_js-node_modules_patternfly_react--46edeb.chunk.js:45757:7)
    at li
    at TreeViewListItemBase (http://localhost:3000/hawtio/static/js/vendors-node_modules_patternfly_react-charts_dist_esm_index_js-node_modules_patternfly_react--46edeb.chunk.js:45901:5)
    at ul
    at TreeViewList (http://localhost:3000/hawtio/static/js/vendors-node_modules_patternfly_react-charts_dist_esm_index_js-node_modules_patternfly_react--46edeb.chunk.js:45861:7)
[...]

PR #392, opened 2023-06-29
Properties view and sorts possible confusion of identifiers of XML nodes

phantomjinx:


tadayosi (approved):
thx!

phantomjinx:
Screenshot_20230629_100947

Addresses #368 by taking advantage of the xml id and uri attributes for ensuring identifiiers are unique and displays this info in the tree

phantomjinx:
Screenshot_20230629_101029

Properties View as implemented

packages/hawtio/src/plugins/camel/routes-service.test.ts
```diff
@@ -87,14 +88,19 @@ describe('routes-service', () => {
...
+      localName: string
       children?: MBeanAttr[]
     }
 
     const expected: MBeanAttr[] = [
-      { id: 'from', name: 'from' },
-      { id: 'setBody', name: 'setBody' },
-      { id: 'to', name: 'to' },
-      { id: 'to', name: 'to' },
+      {
```
joshiraez, line 96: There is a few things wrong with this, although it works for now for what you need:

    This test right now is falsily passing as there is a bug with the IDs generations. You can check the created tree and see that the ids of these MBeans should have something concatenated to them as those are children MBeans, so those id's are false. On top of that, we are no longer using the to3/to4 camel style IDs that we used before. That was fixed on https://github.com/hawtio/hawtio-next/pull/254, as it made really hard for testing.

You can check https://github.com/hawtio/hawtio-next/commit/b8c86143270f9d18761464fc75b526a8e8d0b8fc how the ids should be actually

    Does this only affect "from" nodes? as https://github.com/hawtio/hawtio-next/issues/368 bug happens with the to nodes, that we know will have the same id still if we went with the current generation.

    I'd refrain heavily of using the ids to put some inner developing data inside them, which is only used on one element on top of all from the test, making having for the same attribute two different use cases, which is a code smell. Will make it extremely hard to use it as a css attribute to link to, or for our testers to be able to test.

Instead, why not leave the id as it was (instead of putitng it in localName, as that's not rendered in the HTML), and put the current id string in a private property? From what I understood, you needed it to access the xml properties, so you should be able to access it just fine using a private variable.

In top of that, I'd add a random number as a quick fix for the time being if the problem lies in selecting the correct node when there are multiple similar nodes present. But from the test, the problematic nodes are the to ones, and those remain unchanged?

joshiraez: Never mind the random number. If we go back to putting a number there it diverges from what we did for juri and it becomes arbitrary based on the number of elements. I guess we will need to see if Juri likes that.

But I'd still heavily refraing from having the id be like it is. Id should stay as `from` and `quartz:simple?trigger.repeatInterval={{quartz.repeatInterval}}:from'` should be on private bar, ids shouldn't be used for internal development information as they are used by testers and styling

joshiraez (approved):

PR #401, opened 2023-07-03
Use esbuild plugin to speedup app build

mmuzikar:
Same stuff as in https://github.com/hawtio/hawtio/pull/2873.

`yarn build:all` from ~ 3 min to under 1 minute :tada:


tadayosi (approved):
awesome, thanks!!!

PR #414, opened 2023-07-07
refactor(camel): refactor Camel RouteDiagrams view

tadayosi:
* Removed `selectedNode` from route diagram context; we have been trying to merge the react contexts for each submodule, but I think it's perfectly fine to use multiple react contexts at the same time. React contexts are just a convenient way to use states without passing them as props for cascading sub components. So, ideally we should have normally three contexts for each sub component: the Hawtio context (if any), the root plugin context (camel), and the context specific to a sub component (e.g. route diagram context). Node selection should be provided by the root plugin context.
* Declare return types for the visualization service
* Some other code changes to align with the styles of the other components

Initially I tried to remove the route diagram context and move them to component props, as it seemed to bring tight coupling between RouteDiagram and Debug, Tracer. But now I understand it seems rather necessary (at least for now), so I'll leave it as is.


phantomjinx (approved):
LGTM

PR #416, opened 2023-07-07
refactor(camel): rename ccs -> camelService to align with naming of services

tadayosi:
This service is so generic that it shouldn't be named like serving only CamelContent component.

Also, let's avoid using abbreviation for the alias of such key service module.


phantomjinx (approved):
LGTM

PR #417, opened 2023-07-07
chore: bump prettier to 3.0.0

phantomjinx:
* Adds extra check to format:check so these are fixed as well.

* Supercedes https://github.com/hawtio/hawtio-next/pull/410

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "eslint-config-react-app": "^7.0.1",
     "eslint-plugin-import": "^2.26.0",
     "eslint-plugin-react-hooks": "^4.6.0",
-    "prettier": "2.8.8"
+    "prettier": "3.0.0"
   },
   "config": {
     "commitizen": {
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -43,7 +43,7 @@
     "eslint-config-react-app": "^7.0.1",
     "eslint-plugin-import": "^2.26.0",
     "eslint-plugin-react-hooks": "^4.6.0",
-    "prettier": "2.8.8"
+    "prettier": "^3.0.0"
```
tadayosi, line 46: I think it's better to point to the exact version for prettier (that's what the official doc suggests as well), as automatic upgrades of prettier might cause unexpected lint errors at pull reqs. GitHub bot can automate upgrades of it for us.

tadayosi (changes requested):
Thanks!  There is only one minor comment I'd like to have it addressed.

tadayosi (approved):

PR #463, opened 2023-07-25
chore: exports the auth module for use with external clients

phantomjinx:
This is required for hawtio-online since we need to connect the osAuthService to the userService


tadayosi (approved):
:+1:

PR #480, opened 2023-07-31
fix(build): fix tsconfig.json target warning from esbuild

### LINKED ISSUE #474: This warning seems to be a common occurence during building....
```
Warning: ▲ [WARNING] "tsconfig.json" does not affect esbuild's own target setting [tsconfig.json]
[hawtio] 
[hawtio]     tsconfig.json:7:4:
[hawtio]       7 │     "target": "es2015",
[hawtio]         ╵     ~~~~~~~~
[hawtio] 
[hawtio]   This is because esbuild supports reading from multiple "tsconfig.json" files within a single build, and using different language targets for different files in the same build wouldn't be correct. If you want to set esbuild's language target, you should use esbuild's own global "target" setting such as with "target: 'es2015'".
[hawtio] 
```

Is this something of concern or should it just be ignored?

tadayosi:
Fix #474

@phantomjinx It appears this is enough to fix the warning. Please check.

### DIFF
--- a/packages/hawtio/tsconfig.json
+++ b/packages/hawtio/tsconfig.json
@@ -4,7 +4,6 @@
     "paths": {
       "@hawtiosrc/*": ["src/*"]
     },
-    "target": "es2015",
     "lib": ["dom", "dom.iterable", "esnext"],
     "allowJs": true,
     "skipLibCheck": true,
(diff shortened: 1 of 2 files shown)


phantomjinx (approved):
LGTM. Excellent.

I will also migrate this fix into the packages in hawtio-online.

PR #484, opened 2023-08-03
fix(ui): HAWNG-11 custom brand logo must not be greater than height=50px

tadayosi:
Right now, when a custom brand logo with a big size is provided via `hawtconfig.json`, it's rendered full size in the header and consumes a lot of space on the console.


tadayosi:
E2E Test is flaky, let's ignore the checks.

PR #537, opened 2023-09-06
fix: Ability to add custom toolbar items to main page nav bar

phantomjinx:
This is an attempt to address the dynamic navbar that displays in hawtio-online. The toolbar dropdown is dynamic in that it changes link target, dependent on which plugin route is active. Therefore, this solution attempts to mirror this functionality.

To test, execute `yarn start` and click on `Example 3` plugin to see the newly added custom toolbar items for that plugin.

---

* core.ts
  * Adds optional toolbarItems property to plugin interface

* HawtioHeader.tsx
  * Picks up the plugins from the context and find the plugin that is currently being displayed. If this plugin has a toolbar item property then pick up the components into a custom toolbar group

### DIFF
--- a/packages/hawtio/src/core/core.ts
+++ b/packages/hawtio/src/core/core.ts
@@ -13,6 +13,9 @@ export interface Plugin {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   component: React.ComponentType<any>
 
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  headerItems?: React.ComponentType<any>[]
+
   /**
    * Returns if this plugin should be activated.
    * This method needs to return a promise as the process of resolving if a plugin
--- a/packages/hawtio/src/ui/page/HawtioHeader.tsx
+++ b/packages/hawtio/src/ui/page/HawtioHeader.tsx
@@ -1,5 +1,5 @@
 import { PUBLIC_USER, userService } from '@hawtiosrc/auth'
-import { DEFAULT_APP_NAME, useHawtconfig } from '@hawtiosrc/core'
+import { DEFAULT_APP_NAME, useHawtconfig, Plugin } from '@hawtiosrc/core'
 import { hawtioLogo, userAvatar } from '@hawtiosrc/img'
 import { preferencesService } from '@hawtiosrc/preferences/preferences-service'
 import { HawtioAbout } from '@hawtiosrc/ui/about'
@@ -22,7 +22,7 @@ import {
 } from '@patternfly/react-core'
 import { BarsIcon, HelpIcon } from '@patternfly/react-icons'
 import React, { useContext, useState } from 'react'
-import { Link } from 'react-router-dom'
+import { Link, useLocation } from 'react-router-dom'
 import './HawtioHeader.css'
 import { PageContext } from './context'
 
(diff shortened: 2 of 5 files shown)


packages/hawtio/src/ui/page/HawtioHeader.tsx
```diff
@@ -104,9 +105,27 @@ const HawtioHeaderToolbar: React.FunctionComponent = () => {
...
 
+  /*
+   * Determine which plugin is currently displaying
+   * based on the path of the current location
+   */
+  const pluginFromLocation = (): Plugin | null => {
+    const path = location.pathname
+    if (!path) return null
+
+    return plugins.find(plugin => plugin.path === path) || null
```
tadayosi, line 116: use `??` instead of `||`

packages/hawtio/src/ui/page/HawtioHeader.tsx
```diff
@@ -104,9 +105,27 @@ const HawtioHeaderToolbar: React.FunctionComponent = () => {
     userItems.pop()
   }
 
+  /*
+   * Determine which plugin is currently displaying
+   * based on the path of the current location
+   */
+  const pluginFromLocation = (): Plugin | null => {
+    const path = location.pathname
+    if (!path) return null
```
tadayosi, line 114: when can it be `!path`?  since `pathname` is of type `string`, this check seems useless.

packages/hawtio/src/ui/page/HawtioHeader.tsx
```diff
@@ -104,9 +105,27 @@ const HawtioHeaderToolbar: React.FunctionComponent = () => {
...
+  }
+
+  const plugin = pluginFromLocation()
+
   return (
     <Toolbar id='hawtio-header-toolbar'>
       <ToolbarContent>
+        <ToolbarGroup>
+          {plugin?.toolbarItems?.map((comp, index) => (
+            <ToolbarItem key={`custom-toolbar-item-${index}`}>{React.createElement(comp)}</ToolbarItem>
```
tadayosi, line 126: I would name the key as `hawtio-header-toolbar-plugin-item-${index}`. I want the ids to be named hierarichally starting from the name of the top-level component (i.e. `hawtio-header` in this case) so that the ids are named consistently throughout the entire console.

packages/hawtio/src/ui/page/HawtioHeader.tsx
tadayosi, line 116: does this test still work if the location is under a subpath such as `/jmx/attributes`?  shouldn't it use `startsWith`?

packages/hawtio/src/core/core.ts
```diff
@@ -13,6 +13,9 @@ export interface Plugin {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   component: React.ComponentType<any>
 
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  toolbarItems?: React.ComponentType<any>[]
```
tadayosi, line 17: I would name it `headerItems` instead of `toolbarItems`, as there can be other toolbars in one page such as table toolbars.

tadayosi:
Also, please update the description of the Example 3 page body to explain what this example demonstrates (i.e. extending the header with custom actions).

image

tadayosi:
Also, this should be `feat:` instead of `fix:` as it brings a new feature.

PR #545, opened 2023-09-11
feat(quartz): initial Quartz plugin

### LINKED ISSUE #171: Quartz plugin
Port [Quatz](https://github.com/hawtio/hawtio-integration/tree/main/plugins/quartz) plugin from legacy Hawtio, which provides you with the following views for Quartz schedulers within an application.

- Scheduler
- Triggers
- Jobs

### Scheduler

image

### Triggers

image

### Jobs

image

tadayosi:
Fix #171

image
image
image


mmelko:
I noticed that in the old hawtio quarts scheduler is under the `Quartz Schedulers` folder. If we don't want to have that wouldn't it make more sense just to have some sort of list of schedulers ? I can imagine that could be problematic due to shared tree though.

tadayosi:
Thanks @mmelko. Yes, I omitted the `Quartz Schedulers` folder in the same way as Camel plugin does for `Camel Contexts`. It just consumed vertical space unnecessarily.

As you pointed out, now the schedulers tree is just a tree with a single level, so effectively it's just a list. I thought the same thing and tried a simple PF4 list for the left side of the view:
http://v4-archive.patternfly.org/v4/components/list#with-icons
but the looks were a bit different from the other views with a left side such as JMX and Camel. That's why I choose TreeView for the left side for now.

If we can come up with a better UI using a different component, we can replace the tree view.

PR #546, opened 2023-09-11
fix: Handles operation arguments imported as strings

phantomjinx:
* Some Operation Arguments retrieved by jolokia are surrounded with quotes hence are handled internally as strings.

* Using the current mechanism only populates the arg.name parameter, leaving arg.type null, resulting in some ugly errors when OperationArgument is constructed.

* By checking whether the arg is an Object and if not parsing it mitigates the problem of the quotes

### DIFF
--- a/packages/hawtio/src/plugins/shared/operations/operation.ts
+++ b/packages/hawtio/src/plugins/shared/operations/operation.ts
@@ -1,5 +1,7 @@
+import { isObject } from '@hawtiosrc/util/objects'
 import { stringSorter, trimEnd } from '@hawtiosrc/util/strings'
 import { IJmxOperation, IJmxOperations } from 'jolokia.js'
+import { log } from '../globals'
 
 /**
  * Factory function for Operation objects.
@@ -23,13 +25,21 @@ function addOperation(
   name: string,
   op: IJmxOperation,
 ): void {
-  const operation = new Operation(
-    name,
-    op.args.map(arg => new OperationArgument(arg.name, arg.type, arg.desc)),
-    op.desc,
-    op.ret,
-    op.canInvoke,
-  )
+  const args: OperationArgument[] = []
+  for (const arg of op.args) {
+    let argObj = arg
+    if (!isObject(arg)) {
+      try {
+        argObj = JSON.parse(arg as string)
+      } catch (error) {
+        log.error(`Cannot parse argument ${arg} in operation ${op.desc}`, error)
+        continue
+      }
+    }
+    args.push(new OperationArgument(argObj.name, argObj.type, argObj.desc))
+  }
+
+  const operation = new Operation(name, args, op.desc, op.ret, op.canInvoke)
   operations.push(operation)
   operationMap[operation.name] = operation
 }


packages/hawtio/src/plugins/shared/operations/operation.ts
```diff
@@ -23,13 +25,21 @@ function addOperation(
...
-    name,
-    op.args.map(arg => new OperationArgument(arg.name, arg.type, arg.desc)),
-    op.desc,
-    op.ret,
-    op.canInvoke,
-  )
+  const args: OperationArgument[] = []
+  for (const arg of op.args) {
+    let argObj = arg
+    if (!isObject(arg)) {
```
tadayosi, line 31: It's still not clear to me why this check is needed. The arg type is defined like this:
https://github.com/hawtio/jolokia.js/blob/dc6bb0c8c8c1f1904b58630c8a1aab9e69ac2f36/index.d.ts#L65-L69
Maybe you can provide us with an example of Jolokia requests that are irregular to the defined type, then we can modify the type definition accordingly. Or it could be a bug in Jolokia that should rather be raised to Jolokia project.

packages/hawtio/src/plugins/shared/operations/operation.ts
```diff
@@ -23,13 +25,21 @@ function addOperation(
...
+      try {
+        argObj = JSON.parse(arg as string)
+      } catch (error) {
+        log.error(`Cannot parse argument ${arg} in operation ${op.desc}`, error)
```
tadayosi, line 35: Oftentimes `op.desc` doesn't provide enough information, so it's better to use `name` here.

tadayosi (changes requested):
Can you provide us with an example of Jolokia requests that you want to address with the fix?  I wonder how it happens in the first place.

tadayosi:
Also, the code hasn't been changed so much from v2. I wonder why it wasn't a case with the old Hawtio v2.  (Or it was?)
https://github.com/hawtio/hawtio-integration/blob/c7d814d3cdd087384a8b3d09cfd6514da9052b81/plugins/jmx/ts/operations/operations.service.ts#L58-L66
```ts
    private addOperation(operations: Operation[], operationMap: { [name: string]: Operation },
      opName: string, op: { args: any[], desc: string }): void {
      let operation = new Operation(
        opName,
        op.args.map((arg) => _.assign(new OperationArgument(), arg)),
        op.desc);
      operations.push(operation);
      operationMap[operation.name] = operation;
    }
```

phantomjinx:
Screenshot_20230912_102622

I added some console log messages to the existing source to diagnose the problem - see screenshot.

1. Click on the route node in the camel plugin tree - display the Attributes view;
2. Click on the Operations view and the result is an exception being thrown;

The reason for the exception is that `arg.type` is null because `arg.name` assumes the whole value when the arg is a string.

The console log confirms displays (see 1) that the arg array normally contains properties objects. However, in the case of 'Dumps the route as XML' operation, the arg value is defined as a string (see 2). It seems to be the only arg that is in this form so it may be an error in the dependencies.

If it is a bug further down the stack, then it is probably worth investigating and fixing. However, I think it cautious to check the arguments rather than assume. That was my thinking here.

phantomjinx:
https://github.com/hawtio/hawtio-next/assets/1634180/75eab8be-075e-4eef-910a-a8b86d7befca

Thought a quick video might clarify the process.

tadayosi:
@phantomjinx Can you reproduce the issue with the old Hawtio as well?

tadayosi:
@phantomjinx I cannot reproduce the issue at my end, with either the current `3.x` branch of hawtio standalone project or the `main` branch of hawtio-next.

> I added some console log messages to the existing source to diagnose the problem - see screenshot.

Actually, you don't need to insert `console.log` to just inspect the received mbean tree. Just turn on `hawtio-shared-workspace` logger to DEBUG level and you'll get the mbean tree loaded in the browser console.

In my case, both `dumpStatsAsXml` and `dumpRouteAsXml` operations get the args correctly as objects. I'd suspect the issue is specific to your Camel application setup. We should check which versions of Camel and Jolokia are needed to replicate it.

image

And talking about the fix, if this issue indeed happens in some case and needs to be fixed, I think we should rather fix it earlier than when an operation object is instantiated, somewhere like here:
https://github.com/hawtio/hawtio-next/blob/03ea6c97c7e0d32da09507486f88765a49c63eab/packages/hawtio/src/plugins/shared/workspace.ts#L70-L85

phantomjinx:
> @phantomjinx Can you reproduce the issue with the old Hawtio as well?

Curiously no. The Operation args seem to come through correctly.
Screenshot_20230913_104659

phantomjinx:
The test application, forked from the main quick-starts, is located here -> https://github.com/phantomjinx/spring-boot-camel/tree/fuse-7.x.redhat

tadayosi:
> The test application, forked from the main quick-starts, is located here -> https://github.com/phantomjinx/spring-boot-camel/tree/fuse-7.x.redhat

@phantomjinx I don't reproduce the issue even with the quickstart you use. I guess you're deploying it to OpenShift and trying to connect to it from hawtio-online, aren't you?  Then it should be the issue with the way hawtio-online processes the returned mbeans at its in-the-middle nginx server. See: https://github.com/hawtio/hawtio-online/blob/2.x/docker/nginx.js

If that's the case, let's close this one and instead file this issue to hawtio-online and fix it there.

phantomjinx:
Closing as discussed. Has not proven to be the case elsewhere and if it occurs again on `hawtio-online` will investigate with the target being how the latter provides the network data rather than hawtio-next.

PR #554, opened 2023-09-13
feat: Further enhance the plugin header items extension point

phantomjinx:
* core.ts
  * Some plugins require their header items to remain displayed on the
    header navbar, even though the plugin is not displayed, eg. when
    click 'connect' in hawtio-online with the discover plugin, a new window
    displays hawtio-app showing the connection JMX data. This new window
    still benefits from having the discover plugin dropdown list present so
    that other pods or console can be accessed without returing to the
    discover plugin.
  * Adds a 'universal' property to Plugin structure so that an individual
    plugin can state whether its header items should remain on the header
    toolbar even if the rest of the plugin is not the focused-on display

* HawtioHeader.tsx
  * Find all header items that are either from the focused plugin or
    are marked as 'universal' to build the customized toolbar

### DIFF
--- a/packages/hawtio/src/core/core.ts
+++ b/packages/hawtio/src/core/core.ts
@@ -3,6 +3,36 @@ import $ from 'jquery'
 import { eventService } from './event-service'
 import { log } from './globals'
 
+/*
+ * Components to be added to the header navbar
+ * Can define either a single component type or
+ * a component with a universal property.
+ *
+ * By default, components will only be displayed
+ * if the plugin UI is also visible. However, setting
+ * universal to 'true' will ensure the component
+ * remains displayed regardless of which plugin is
+ * given focus.
+ */
+export interface UniversalHeaderItem {
+  // The components that should be populated as
+  // dropdown items on the header bar
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  component: React.ComponentType<any>
+
+  // Should components remain visible on header even when
+  // the plugin is not being displayed.
+  universal: boolean
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type HeaderItem = React.ComponentType<any> | UniversalHeaderItem
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function isUniversalHeaderItem(obj: any): obj is UniversalHeaderItem {
+  return typeof obj.universal === 'boolean'
+}
+
 /**
  * Internal representation of a Hawtio plugin.
  */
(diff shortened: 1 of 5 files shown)


phantomjinx:
Tests seem to be all failing :-|

packages/hawtio/src/core/core.ts
```diff
@@ -3,6 +3,17 @@ import $ from 'jquery'
...
+export interface HeaderItems {
+  // The components that should be populated as
+  // dropdown items on the header bar
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  components: React.ComponentType<any>[]
+
+  // Should components remain visible on header even when
+  // the plugin is not being displayed.
+  universal: boolean
+}
```
tadayosi, line 15: We can instead define `HeaderItem` type as follows:
```
export type HeaderItem = React.ComponentType<any> | {
  component: React.ComponentType<any>
  universal: boolean
}
```
This way you can define a mix of local and universal header items within a plugin. The current fix only allows either all local or all universal.

packages/hawtio/src/ui/page/HawtioHeader.tsx
```diff
@@ -109,20 +109,31 @@ const HawtioHeaderToolbar: React.FunctionComponent = () => {
...
+      if (path.startsWith(plugin.path)) return true
+
+      if (!plugin.headerItems) return false
+
+      return plugin.headerItems.universal === true
+    })
   }
 
-  const plugin = pluginFromLocation()
+  const filteredPlugins = filterPlugins()
```
tadayosi, line 123: I think we don't need to filter plugins here. Instead we can just directly collect only relevant header item components againt all the plugins here, like:
```
const headerItems: React.ComponentType<any>[] = collectHeaderItems()
```

packages/hawtio/src/ui/page/HawtioHeader.tsx
```diff
@@ -109,20 +109,31 @@ const HawtioHeaderToolbar: React.FunctionComponent = () => {
...
   return (
     <Toolbar id='hawtio-header-toolbar'>
       <ToolbarContent>
         <ToolbarGroup>
-          {plugin?.headerItems?.map((comp, index) => (
-            <ToolbarItem key={`hawtio-header-toolbar-plugin-item-${index}`}>{React.createElement(comp)}</ToolbarItem>
-          ))}
+          {filteredPlugins.map(
+            plugin =>
+              plugin.headerItems?.components.map((comp, index) => (
```
tadayosi, line 131: Then you can directly map against the `headerItems` components here instead of the lengthy `filterPlugins.map(plugin => plugin.headerItems?.components)` chain.

tadayosi:
> Tests seem to be all failing :-|

It's because of https://github.com/hawtio/hawtio/pull/2937. Can you talk to @nstefeko @mmuzikar @jsolovjo to resolve the test failures as well?

jsolovjo:
> > Tests seem to be all failing :-|
> 
> It's because of [hawtio/hawtio#2937](https://github.com/hawtio/hawtio/pull/2937). Can you talk to @nstefeko @mmuzikar @jsolovjo to resolve the test failures as well?

Hello,
Nina is OOO today, I can check the test failures and try to fix it
Thanks for letting me know

PR #556, opened 2023-09-14
feat(runtime): Initial Runtime plugin

mmelko:
Adds Runtime plugin: with Properties, Metrics and Threads tabs:
image

image

image
image

image

### DIFF
--- a/packages/hawtio/src/plugins/runtime/Metrics.tsx
+++ b/packages/hawtio/src/plugins/runtime/Metrics.tsx
@@ -1,23 +1,31 @@
 import { Card, CardBody, CardHeader, Grid, GridItem, Title } from '@patternfly/react-core'
 import React, { useEffect, useState } from 'react'
-import { loadMetrics, REFRESH_INTERVAL } from '@hawtiosrc/plugins/runtime/runtime-service'
-import { Metric } from '@hawtiosrc/plugins/runtime/types'
+import { runtimeService } from './runtime-service'
+import { Metric } from './types'
 import { ChartBullet } from '@patternfly/react-charts'
 
 export const Metrics: React.FunctionComponent = () => {
-  const [metrics, setMetrics] = useState<Metric[]>([])
+  const [metrics, setMetrics] = useState<Record<string, Metric>>({})
 
   useEffect(() => {
-    let timeoutHandle: NodeJS.Timeout
-    const readMetrics = async () => {
-      const metrics = await loadMetrics()
-      setMetrics(metrics)
+    const registerMetricsRequests = () => {
+      let metricsRecord: Record<string, Metric> = {}
+      runtimeService.registerMetrics(metric => {
+        metricsRecord = { ...metricsRecord, [metric.name]: metric }
+        setMetrics(metricsRecord)
+      })
+    }
 
-      timeoutHandle = setTimeout(readMetrics, REFRESH_INTERVAL)
+    const readMetrics = async () => {
+      const metricsList = await runtimeService.loadMetrics()
+      let metricsRecord: Record<string, Metric> = {}
+      metricsList.forEach(metric => (metricsRecord = { ...metricsRecord, [metric.name]: metric }))
+      setMetrics(metricsRecord)
     }
 
     readMetrics()
-    return () => timeoutHandle && clearTimeout(timeoutHandle)
+    registerMetricsRequests()
+    return () => runtimeService.unregisterAll()
   }, [])
 
   return (
@@ -28,7 +36,7 @@ export const Metrics: React.FunctionComponent = () => {
             <Title headingLevel='h2'>System</Title>
           </CardHeader>
           <CardBody>
-            {metrics
+            {Object.values(metrics)
               .filter(m => m.type === 'System')
               .map((metric, index) => {
                 return (
@@ -63,15 +71,15 @@ export const Metrics: React.FunctionComponent = () => {
           </CardHeader>
 
           <CardBody>
-            {metrics
+            {Object.values(metrics)
               .filter(m => m.type === 'JVM')
               .map((metric, index) => {
                 return (
                   <div key={index}>
-                    {metric.name} :{' '}
+                    {metric.name} :
                     <span>
-                      {metric.value} {metric.unit ?? ''}{' '}
-                      {metric.available && 'of' + metric.available + ' ' + metric.unit}
+                      {metric.value} {metric.unit ?? ''}
+                      {metric.available && 'of' + metric.available + ' ' + (metric.unit ?? '')}
                     </span>
                   </div>
                 )
--- a/packages/hawtio/src/plugins/runtime/Metrics.tsx
+++ b/packages/hawtio/src/plugins/runtime/Metrics.tsx
@@ -1,6 +1,6 @@
 import { Card, CardBody, CardHeader, Grid, GridItem, Title } from '@patternfly/react-core'
 import React, { useEffect, useState } from 'react'
-import { getMetrics } from '@hawtiosrc/plugins/runtime/runtime-service'
+import { getMetrics, REFRESH_INTERVAL } from '@hawtiosrc/plugins/runtime/runtime-service'
 import { Metric } from '@hawtiosrc/plugins/runtime/types'
 import { ChartBullet } from '@patternfly/react-charts'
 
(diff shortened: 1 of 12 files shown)


packages/hawtio/src/plugins/runtime/SysProps.tsx
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    //filter with the rest of the filters
+    filters.forEach(value => {
+      const attr = value.split(':')[0] ?? ''
+      const searchTerm = value.split(':')[1] ?? ''
+      filtered = filtered.filter(prop =>
+        (attr === 'name' ? prop.key : prop.value).toLowerCase().includes(searchTerm.toLowerCase()),
+      )
+    })
```
tadayosi, line 104: The filtering logic itself is something we can move to runtime-service.ts, so that the component logic can be simplified.

packages/hawtio/src/plugins/runtime/SysProps.tsx
```diff
@@ -0,0 +1,235 @@
...
+        const bValue = sortIndex === 1 ? b.value : b.key
+        return objectSorter(aValue, bValue, sortDirection === 'desc')
+      })
+    }
+
+    return sortedProps
+  }
+  return (
+    <Card isFullHeight>
+      {/*<CardTitle>Browse Messages</CardTitle>*/}
```
tadayosi, line 154: Need this?

packages/hawtio/src/plugins/runtime/SysProps.tsx
```diff
@@ -0,0 +1,235 @@
...
+
+            <ToolbarItem variant='pagination'>
+              <PropsPagination />
+            </ToolbarItem>
+          </ToolbarContent>
+        </Toolbar>
+
+        {sortProperties().length > 0 && (
+          <FormGroup>
+            <TableComposable aria-label='Message Table' variant='compact' height='80vh'>
```
tadayosi, line 200: you can make the table striped.

packages/hawtio/src/plugins/runtime/Threads.tsx
```diff
@@ -0,0 +1,339 @@
...
+            </ToolbarGroup>
+
+            <ToolbarItem variant='pagination'>
+              <PropsPagination />
+            </ToolbarItem>
+          </ToolbarContent>
+        </Toolbar>
+        {sortThreads().length > 0 && (
+          <FormGroup>
+            <TableComposable aria-label='Message Table' variant='compact' height='80vh'>
```
tadayosi, line 293: you can make the table striped.

packages/hawtio/src/plugins/runtime/index.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const runtime: HawtioPlugin = () => {
+  hawtio.addPlugin({
+    id: pluginId,
+    title: 'Runtime',
+    path: pluginPath,
+    component: Runtime,
+    isActive: async () => workspace.hasMBeans(),
+  })
+  helpRegistry.add(pluginId, 'Runtime', help, 15)
```
tadayosi, line 16: Quartz already uses 15. It should use 16.

packages/hawtio/src/plugins/runtime/runtime-service.ts
```diff
@@ -0,0 +1,134 @@
...
+function convertMsToDaysHours(ms: number): string {
+  const seconds = Math.floor(ms / 1000)
+  const minutes = Math.floor(seconds / 60)
+  const hours = Math.floor(minutes / 60)
+  const days = Math.floor(hours / 24)
+
+  return `${days} days, ${hours % 24} hours`
+}
+
+export function getSystemProperties(): Promise<SystemProperty[]> {
```
tadayosi, line 14: I'd use `load` instead of `get` when a function reads data from an external service like jolokia.

packages/hawtio/src/plugins/runtime/Metrics.tsx
```diff
@@ -0,0 +1,87 @@
...
+export const Metrics: React.FunctionComponent = () => {
+  const [metrics, setMetrics] = useState<Metric[]>([])
+
+  useEffect(() => {
+    let timeoutHandle: NodeJS.Timeout
+    const readMetrics = async () => {
+      const metrics = await getMetrics()
+      setMetrics(metrics)
+
+      timeoutHandle = setTimeout(readMetrics, REFRESH_INTERVAL)
```
tadayosi, line 16: I realised that using Jolokia scheduler instead of `setTimeout()` is a bit more efficient, especially when we need to fire multiple requests on jolokia like here. It's because with Jolokia scheduler only one scheduling loop is used for all the registered requests in the console. So if each view makes sure to use Jolokia scheduler for constant reloading of MBeans states, Hawtio only needs to fire a single scheduled (batch) request per 5 seconds (default). Instead, if you use `setTimeout` the requests are fired separately, thus imposing additional network loads.

See how I implemented a custom mbean load scheduler for Quartz Jobs view:
https://github.com/hawtio/hawtio-next/blob/61b3d7f9cf14312487fcb8141610e3c2c0cd7f0c/packages/hawtio/src/plugins/quartz/jobs/Jobs.tsx#L47-L66

tadayosi (commented):
thanks. a few comments.

packages/hawtio/src/plugins/runtime/Metrics.tsx
```diff
@@ -0,0 +1,87 @@
...
+                  </div>
+                )
+              })}
+          </CardBody>
+        </Card>
+      </GridItem>
+      <GridItem>
+        <Card>
+          <CardHeader>
+            {' '}
```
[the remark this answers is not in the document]
mmelko: prettier sometimes decides to format some whitespace like that .. I missed those. Thanks!

packages/hawtio/src/plugins/runtime/SysProps.tsx
mmelko: hmm I was rather thinking to create a new generic table component with sorting and filtering at some point. Because I have sort of similar things at 3 places.

tadayosi: Sounds good. Then maybe we can still put the filtering logic to a generic table-service.ts.

But this is not a big problem so we can leave it as is for now if opinions diverge.

tadayosi (commented):
There are still some comments that are yet to be addressed.

packages/hawtio/src/plugins/runtime/Threads.tsx
```diff
@@ -0,0 +1,333 @@
...
+  const [currentThread, setCurrentThread] = useState<Thread>()
+  const [threadConnectionMonitoring, setThreadConnectionMonitoring] = useState(false)
+
+  useEffect(() => {
+    const readThreads = async () => {
+      const threads = await runtimeService.loadThreads()
+      setThreads(threads)
+      setFilteredThreads(threads)
+      setThreadConnectionMonitoring(await runtimeService.isThreadContentionMonitoringEnabled())
+      await runtimeService.registerLoadThreadsRequest(threads => setThreads(threads))
```
tadayosi, line 88: Correct me if I'm wrong, but in this case when the scheduler updates the threads, aren't the filtered threads updated with the existing filters as well?  So, I guess the updated set of threads doesn't get updated until the user hits the search button, does it?

packages/hawtio/src/plugins/runtime/Metrics.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { runtimeService } from './runtime-service'
+import { Metric } from './types'
+import { ChartBullet } from '@patternfly/react-charts'
+
+export const Metrics: React.FunctionComponent = () => {
+  const [metrics, setMetrics] = useState<Map<string, Metric>>(new Map())
+
+  useEffect(() => {
+    const registerMetricsRequests = () => {
+      const map = new Map<string, Metric>()
```
tadayosi, line 12: We don't use `Map` in general. You can use `Record` instead unless there's a specific reason to use `Map`.

packages/hawtio/src/plugins/runtime/Metrics.tsx
```diff
@@ -0,0 +1,93 @@
...
+    const readMetrics = async () => {
+      const metricsList = await runtimeService.loadMetrics()
+      const metricsMap: Map<string, Metric> = new Map()
+      metricsList.map(m => metricsMap.set(m.name, m))
+      setMetrics(metricsMap)
+    }
+
+    readMetrics()
+    registerMetricsRequests()
+    //   return () => timeoutHandle && clearTimeout(timeoutHandle)
```
tadayosi, line 28: can you remove the comment?

packages/hawtio/src/plugins/runtime/Threads.tsx
mmelko: nice catch!  thanks!

PR #569, opened 2023-09-25
fix(shared): upgrade jolokia.js to 2.0.0-dev.9

tadayosi:
The new official jolokia.js should be backward compatible with Jolokia 1.x, so it should be fine to upgrade the js client library in advance for the coming Jolokia 2.0 release.

The new jolokia.js should also support both Jolokia 1.x and 2.x agents with a single distribution.

Based on jolokia/jolokia#574


tadayosi:
@phantomjinx @mmelko I asked your reviews just because it's an important change to know for the Hawtio developers.

packages/hawtio/src/plugins/shared/jolokia-service.ts
```diff
@@ -518,57 +523,73 @@ class JolokiaService implements IJolokiaService {
...
-export type AttributeValues = { [name: string]: unknown }
+export type AttributeValues = Record<string, unknown>
 
+/* eslint-disable @typescript-eslint/no-unused-vars */
 /**
  * Dummy Jolokia implementation that does nothing.
  */
-/* eslint-disable @typescript-eslint/no-unused-vars */
-class DummyJolokia implements IJolokia {
+class DummyJolokia implements Jolokia {
```
phantomjinx, line 534: I have had to copy this into hawtio-online in order to always return an implementation of jolokia, regardless of error. Can we export it, please?

Of course, it might be that a review of `ManagedPod` might render this mute.

https://github.com/hawtio/hawtio-online/blob/2.x/packages/management-api/src/managed-pod.ts#L21
https://github.com/hawtio/hawtio-online/blob/2.x/packages/management-api/src/managed-pod.ts#L110

packages/hawtio/package.json
```diff
@@ -45,13 +45,14 @@
...
     "@types/jest": "^29.5.5",
+    "@types/jquery": "^3.5.19",
     "@types/node": "^18.17.18",
     "@types/react": "^18.2.21",
     "@types/react-dom": "^18.2.7",
     "@types/react-router-dom": "^5.3.3",
     "dagre": "^0.8.5",
     "eventemitter3": "^5.0.1",
-    "jolokia.js": "^1.7.3-2",
+    "jolokia.js": "^2.0.0-dev.9",
```
phantomjinx, line 55: Is a dev release a viable version we should be depending on? Presumably, this soak-tests it before doing a full release of both `jolokia` and `hawtio-next` ?

phantomjinx (approved):
LGTM

packages/hawtio/src/plugins/shared/jolokia-service.ts
tadayosi: @phantomjinx It's an important point but I don't think is related to this pull req. It's a bigger topic than briefly duscussing here, so please file a new issue to discuss it further.

Here are my points:
* As we discussed, hawtio-online should rather use `jolokiaService` instead of directly depending on `jolokia.js`. This way, I don't think we need to export this `DummyJolokia` class.
* What's important to discuss here is really that currently `jolokiaService` is provided as a singleton service, whereas hawtio-online should expect to use multiple Jolokia instances with each distinct connection, because hawtio-online needs to switch to different Jolokia endpoints provided by pods. We'd rather need to consider how to change the `jolokiaService` so that it can serve hawtio-online as well.

Let's discuss it further at a new issue.

packages/hawtio/package.json
tadayosi: Good point. What's changed really between `1.7.3` and `2.0.0-dev.x` is the type definitions only and the main `jolokia.js` and `jolokia-simple.js` haven't been changed. So I can safely say we can depend on it even for the current Jolokia 1.7.x agent. After all, who wrote the type definitions is me and I know it works ;-)

packages/hawtio/src/plugins/shared/jolokia-service.ts
tadayosi: @phantomjinx Went ahead and filed #573 for you.

Issue #576, opened 2023-09-27
Bump @types/react


tadayosi:
Dependabot bump of the dependency failed because of the breaking changes #574. Manual bump of the version is needed.

@phantomjinx @mmelko @lhein Please anyone take a look.

tadayosi:
@phantomjinx Thanks!

PR #585, opened 2023-10-02
chore(deps): bump @types/react from 18.2.21 to 18.2.23 #576

phantomjinx:
Appears to be compiling correctly.

ref: https://github.com/hawtio/hawtio-next/pull/574

#576


tadayosi (approved):
Thanks!

PR #607, opened 2023-10-12
Changes to <HawtioLogin> to allow for plugin to provide customisations

phantomjinx:
While most changes concern updating the UI of the `<HawtioLogin/>` to support a customised child login form, a necessary change is also included for the `JolokiaService`. Specifically, the use-case concerns what happens when logging-out occurs.

When the logging-out button is clicked, the app routes back to `/login` but *was* stuck on displaying the waiting symbol.
Analysis showed that because `<HawtioLogin>` is now using the `usePlugins` hook, the component waits on the `pluginsLoaded` flag. This was never returning `true` indicating that the plugins were never completing resolution. Digging in further indicated that despite not being logged-in the `JMX` plugin was, as a result of its `isActive()` plugin function, still trying to list nodes from the `jolokiaService` yet this process was hanging.
This has been resolved by, since the service was already waiting for the `userService.isLogin()` to return, taking the result of `isLogin()` and if `false` then returning a `null` jolokia url. This causes a dummy tree to be created and the `JMX` plugin to resolve satisfactorily.

### DIFF
--- a/packages/hawtio/src/plugins/shared/jolokia-service.ts
+++ b/packages/hawtio/src/plugins/shared/jolokia-service.ts
@@ -130,7 +130,11 @@ class JolokiaService implements IJolokiaService {
 
   private async initJolokiaUrl(): Promise<string | null> {
     // Wait for resolving user as it may attach credentials to http request headers
-    await userService.isLogin()
+    const loggedIn = await userService.isLogin()
+    if (!loggedIn) {
+      log.debug("Not logged in so no jolokia url should be possible")
+      return null
+    }
 
     // Check remote connection
     const conn = connectService.getCurrentConnectionName()
--- a/packages/hawtio/src/plugins/shared/jolokia-service.ts
+++ b/packages/hawtio/src/plugins/shared/jolokia-service.ts
@@ -100,19 +100,30 @@ export interface IJolokiaService {
 }
 
 class JolokiaService implements IJolokiaService {
-  private jolokiaUrl: Promise<string | null>
-  private jolokia: Promise<Jolokia>
+  private jolokiaUrl: Promise<string | null> | null = null
+  private jolokia: Promise<Jolokia> | null = null
   private config: JolokiaConfig = {
     method: JolokiaListMethod.DEFAULT,
     mbean: OPTIMISED_JOLOKIA_LIST_MBEAN,
   }
 
-  constructor() {
-    this.jolokiaUrl = this.initJolokiaUrl()
-    this.jolokia = this.createJolokia()
+  protected async init(): Promise<Jolokia> {
+    // Wait for resolving user as it may attach credentials to http request headers
+    const loggedIn = await userService.isLogin()
+    if (!loggedIn) {
+      throw new Error('Jolokia service not available as user is not logged-in')
+    }
 
-    // Start Jolokia
-    this.jolokia.then(jolokia => {
+    if (!this.jolokiaUrl) {
+      this.jolokiaUrl = this.initJolokiaUrl()
+      await this.jolokiaUrl
+    }
+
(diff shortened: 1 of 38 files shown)


phantomjinx:
Is a dependency of https://github.com/hawtio/hawtio-online/pull/155

packages/hawtio/src/plugins/shared/jolokia-service.ts
```diff
@@ -130,7 +130,11 @@ class JolokiaService implements IJolokiaService {
 
   private async initJolokiaUrl(): Promise<string | null> {
     // Wait for resolving user as it may attach credentials to http request headers
-    await userService.isLogin()
+    const loggedIn = await userService.isLogin()
+    if (!loggedIn) {
+      log.debug("Not logged in so no jolokia url should be possible")
+      return null
+    }
```
tadayosi, line 137: Why is this change needed? This should slightly change the flow, where the connection name is no longer checked when it's not logged in. So I'd like to understand why it's really necessary and that it's really a safe change.

tadayosi (approved):
thanks! there is only one comment I'd like to be cleared.

tadayosi:
Please fix lint error too.

packages/hawtio/src/plugins/shared/jolokia-service.ts
tadayosi: so it's needed as otherwise JMX view hangs. but I haven't seen such issues before. what's the specific condition that may cause the issue?

I wouldn't like jumping into solution without clearly identifying a real issue.

tadayosi: OK, now I understand what's the issue here. So we changed the plugins so that they can also affect the login phase, now the plugins also need to be aware of the login status. Before that, we could assume that login is done before resolving plugins. That's why this change is needed.

Then, I think there are other options to fix it.
1. Make every `isActive` function check the login status and only activate plugins that are relevant depending on the status.
2. Change `HawtioCore.resolvePlugins()` to be aware of the login status and 1) if it's `isLogin` it should be activated only if the user is not logged in yet & it's `isActive`, 2) if it's not `isLogin` it should be activated only if the user is logged in & it's `isActive`. 
https://github.com/hawtio/hawtio-next/blob/621f1a3eae543767526cc1731e87abbfb969f545/packages/hawtio/src/core/core.ts#L224-L228

I think of these, 2) is the preferable option.

It is because there has been an implicit assumption that every plugin is activated under user's login. The original fix might break the assumption, and I'm not sure if it's a good thing.

tadayosi: Additionally, we can keep the `loggedIn` check in the original change but throw an exception when it's not logged in, because the jolokia service assumes it's always after successful login.

phantomjinx: Cool. I'll look into that. Glad we can come up with a solution. Was not wholly comfortable with the change I made hence the long explanations so good to have another set of eyes on the problem. Thx!

phantomjinx: Has required changes to `core.ts` to only activate plugins if user is logged-in, with the exception of plugins that are flagged as `login` plugins.

Has required changes to the following services to have them respect the userService.isLogin() state. This has meant that they can no longer execute jolokia-related functions in their constructors but instead execute an init() function prior to execution of their public functions:
- jolokia-service.ts
- workspace.ts
- rbac-service.ts

tadayosi:
There are a few issues that need to be addressed in the latest fix. But instead of running another cycle of commit & review, I can take care of them after merging it. Feel free to turn it to "ready for merge" and merge it.

PR #616, opened 2023-10-18
fix(shared): jolokia instance is created every time getJolokia() is invoked

tadayosi:
Basically, lazy-init methods that return Promise should not use async/await as otherwise we would be screwed up with subtle concurrency issues.


tadayosi:
cc @phantomjinx

Issue #625, opened 2023-10-19
Add noconsole rule to eslint to stop casual use of console.[log|error] messages


phantomjinx:
While console.[log|error] is *really* useful for debugging, the messages should not be carried forward into committed code. Use [eslint-noconsole](https://eslint.org/docs/latest/rules/no-console#rule-details) rule to detect and flag.

PR #626, opened 2023-10-19
fix: Adds noconsole rule to guard against casual console.log msgs #625

phantomjinx:
* Uses eslint-disable-next-line to keep those instances where it makes sense to use console.[log|error]

issue #625

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -13,6 +13,7 @@ module.exports = {
   parser: '@typescript-eslint/parser',
   plugins: ['@typescript-eslint', 'import'],
   rules: {
+    'no-console': 'error',
     semi: ['error', 'never'],
     '@typescript-eslint/explicit-member-accessibility': [
       'warn',
--- a/packages/hawtio/scripts/generate-camel-svg-index.js
+++ b/packages/hawtio/scripts/generate-camel-svg-index.js
@@ -1,4 +1,5 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
+/* eslint-disable no-console */
 /* jshint node: true */
 const fs = require('fs')
 const path = require('path')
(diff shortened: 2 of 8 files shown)


.eslintrc.js
```diff
@@ -38,6 +38,7 @@ module.exports = {
         },
       },
     ],
+    'no-console': 'error',
```
tadayosi, line 41: you can put it higher than other testing related rules (and even react specific one), because it concerns the main code.

packages/hawtio/scripts/generate-camel-svg-index.js
```diff
@@ -100,6 +100,7 @@ async function writeToIndex(svgpath, idx) {
 }
 
 if (!fs.existsSync(svgdir)) {
+  // eslint-disable-next-line no-console
```
tadayosi, line 103: it makes more sense to disable the lint for entire file, because it's a script.

packages/hawtio/scripts/generate-camel-svg-index.js
```diff
@@ -109,4 +110,5 @@ if (fs.existsSync(index)) {
   fs.unlinkSync(index)
 }
 
+// eslint-disable-next-line no-console
```
tadayosi, line 113: same here.

packages/hawtio/src/core/logging.ts
```diff
@@ -79,13 +79,15 @@ class LocalStorageHawtioLogger implements HawtioLogger {
       const logLevel = this.loadLogLevel()
       jsLogger.setLevel(logLevel)
     } catch (e) {
+      // eslint-disable-next-line no-console
```
tadayosi, line 82: likewise, this is where the logging is defined so it rather makes sense to disable it for the entire file.

tadayosi (commented):
Thanks!  There are a few comments but otherwise it looks great.

tadayosi:
@phantomjinx What's the current status of this?

tadayosi (approved):

PR #654, opened 2023-11-09
fix(camel,jmx): fix Camel plugin should precede JMX

### LINKED ISSUE #653: Camel plugin should precede JMX
Right now, JMX is at the top position after connecting to the JVM, but JMX is so generic that we should put Camel on top of JMX.

image

However, it's still debatable that we should always put JMX at the bottom.

tadayosi:
Fix #653


tadayosi:
cc @lhein

PR #671, opened 2023-11-16
fix(jmx,camel): fix Operations tab crashes when used with Camel JBang

### LINKED ISSUE #670: Camel - Operations tab crashes when used with Camel JBang
Reproducer step:
```
camel init test.java
camel run test.java &
camel hawtio test --version=4.0-M3
```
then open Operations tab on any of Camel MBean nodes.

It appears the root cause is somehow related to #546, as the root cause seems that some of the Camel operations' args return as string instead of object.

image

tadayosi:
Fix #670


tadayosi:
FYI @phantomjinx

PR #680, opened 2023-11-22
feat(springboot-plugin): Add trace view 

mmelko:
This PR adds a trace view into the spring-boot plugin: 
Screenshot 2023-11-22 at 21 31 26
Screenshot 2023-11-22 at 21 31 33
Also spring-boot service is refactored to the class, for preserving infromation about spring-boot version used.

### DIFF
--- a/packages/hawtio/src/plugins/springboot/Health.tsx
+++ b/packages/hawtio/src/plugins/springboot/Health.tsx
@@ -11,8 +11,8 @@ import {
   PageSection,
   Title,
 } from '@patternfly/react-core'
-import { springbootService } from '@hawtiosrc/plugins/springboot/springboot-service'
-import { HealthComponentDetail, HealthData } from '@hawtiosrc/plugins/springboot/types'
+import { springbootService } from './springboot-service'
+import { HealthComponentDetail, HealthData } from './types'
 import { TableComposable, Tbody, Td, Tr } from '@patternfly/react-table'
 import { humanizeLabels } from '@hawtiosrc/util/strings'
 import { ChartDonutUtilization } from '@patternfly/react-charts'
--- a/packages/hawtio/src/plugins/springboot/Health.tsx
+++ b/packages/hawtio/src/plugins/springboot/Health.tsx
@@ -11,7 +11,7 @@ import {
   PageSection,
   Title,
 } from '@patternfly/react-core'
-import { loadHealth } from '@hawtiosrc/plugins/springboot/springboot-service'
+import { springbootService } from '@hawtiosrc/plugins/springboot/springboot-service'
 import { HealthComponentDetail, HealthData } from '@hawtiosrc/plugins/springboot/types'
 import { TableComposable, Tbody, Td, Tr } from '@patternfly/react-table'
 import { humanizeLabels } from '@hawtiosrc/util/strings'
@@ -97,7 +97,7 @@ export const Health: React.FunctionComponent = () => {
   const [healthData, setHealthData] = useState<HealthData>()
 
   useEffect(() => {
-    loadHealth().then(healthData => {
+    springbootService.loadHealth().then(healthData => {
       setHealthData(healthData)
     })
   }, [])
@@ -124,7 +124,7 @@ export const Health: React.FunctionComponent = () => {
               key={component.name}
               mdRowSpan={componentSpanRecord[component.name]![1] as gridSpans}
             >
-              <Card>
+              <Card isFullHeight>
                 <CardHeader>
                   <Title headingLevel='h3'>{humanizeLabels(component.name!)}</Title>
                 </CardHeader>
--- a/packages/hawtio/src/plugins/springboot/Info.tsx
+++ b/packages/hawtio/src/plugins/springboot/Info.tsx
@@ -1,6 +1,6 @@
 import React, { useEffect, useState } from 'react'
 import { FormGroup, PageSection } from '@patternfly/react-core'
-import { springbootService } from '@hawtiosrc/plugins/springboot/springboot-service'
+import { springbootService } from './springboot-service'
 import { TableComposable, Tbody, Td, Th, Thead, Tr } from '@patternfly/react-table'
 
 export const Info: React.FunctionComponent = () => {
--- a/packages/hawtio/src/plugins/springboot/Info.tsx
+++ b/packages/hawtio/src/plugins/springboot/Info.tsx
@@ -1,13 +1,13 @@
 import React, { useEffect, useState } from 'react'
 import { FormGroup, PageSection } from '@patternfly/react-core'
-import { getInfo } from '@hawtiosrc/plugins/springboot/springboot-service'
+import { springbootService } from '@hawtiosrc/plugins/springboot/springboot-service'
 import { TableComposable, Tbody, Td, Th, Thead, Tr } from '@patternfly/react-table'
 
 export const Info: React.FunctionComponent = () => {
   const [systemProperties, setSystemProperties] = useState<{ key: string; value: string }[]>([])
 
   useEffect(() => {
-    getInfo().then(res => {
+    springbootService.getInfo().then(res => {
       setSystemProperties(res)
     })
   }, [])
--- a/packages/hawtio/src/plugins/springboot/Loggers.tsx
+++ b/packages/hawtio/src/plugins/springboot/Loggers.tsx
@@ -19,7 +19,7 @@ import {
   ToolbarGroup,
   ToolbarItem,
 } from '@patternfly/react-core'
-import { configureLogLevel, getLoggerConfiguration } from '@hawtiosrc/plugins/springboot/springboot-service'
+import { springbootService } from '@hawtiosrc/plugins/springboot/springboot-service'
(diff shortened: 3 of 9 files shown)


packages/hawtio/src/plugins/springboot/Health.tsx
```diff
@@ -11,7 +11,7 @@ import {
   PageSection,
   Title,
 } from '@patternfly/react-core'
-import { loadHealth } from '@hawtiosrc/plugins/springboot/springboot-service'
+import { springbootService } from '@hawtiosrc/plugins/springboot/springboot-service'
```
tadayosi, line 14: should be relative path.

packages/hawtio/src/plugins/springboot/Info.tsx
```diff
@@ -1,13 +1,13 @@
 import React, { useEffect, useState } from 'react'
 import { FormGroup, PageSection } from '@patternfly/react-core'
-import { getInfo } from '@hawtiosrc/plugins/springboot/springboot-service'
+import { springbootService } from '@hawtiosrc/plugins/springboot/springboot-service'
```
tadayosi, line 3: should be relative path.

packages/hawtio/src/plugins/springboot/Loggers.tsx
```diff
@@ -19,7 +19,7 @@ import {
   ToolbarGroup,
   ToolbarItem,
 } from '@patternfly/react-core'
-import { configureLogLevel, getLoggerConfiguration } from '@hawtiosrc/plugins/springboot/springboot-service'
+import { springbootService } from '@hawtiosrc/plugins/springboot/springboot-service'
```
tadayosi, line 22: should be relative path.

packages/hawtio/src/plugins/springboot/SpringBoot.tsx
```diff
@@ -5,8 +5,8 @@ import { Navigate, NavLink, Route, Routes, useLocation } from 'react-router-dom'
 import { Health } from '@hawtiosrc/plugins/springboot/Health'
 import { Info } from '@hawtiosrc/plugins/springboot/Info'
 import { Loggers } from '@hawtiosrc/plugins/springboot/Loggers'
-import { Trace } from '@hawtiosrc/plugins/springboot/Trace'
-import { hasEndpoint } from '@hawtiosrc/plugins/springboot/springboot-service'
+import { TraceView } from '@hawtiosrc/plugins/springboot/TraceView'
+import { springbootService } from '@hawtiosrc/plugins/springboot/springboot-service'
```
tadayosi, line 9: the two imports should be relative path.

packages/hawtio/src/plugins/springboot/TraceView.tsx
```diff
@@ -0,0 +1,328 @@
...
+  ToolbarGroup,
+  ToolbarItem,
+} from '@patternfly/react-core'
+import { TableComposable, Tbody, Td, Th, Thead, Tr } from '@patternfly/react-table'
+import { CheckCircleIcon, ExclamationCircleIcon, SearchIcon } from '@patternfly/react-icons'
+import { Trace } from './types'
+import { springbootService } from './springboot-service'
+
+const HttpStatusIcon: React.FunctionComponent<{ code: number }> = ({ code }) => {
+  if (code < 400) return <CheckCircleIcon color={'#3E8635'} />
```
tadayosi, line 32: why need `{'...'}` instead of just `'...'`?

packages/hawtio/src/plugins/springboot/TraceView.tsx
```diff
@@ -0,0 +1,328 @@
...
+  else return <ExclamationCircleIcon color={'#C9190B'} />
```
tadayosi, line 33: same here

packages/hawtio/src/plugins/springboot/index.ts
```diff
@@ -4,7 +4,7 @@ import { helpRegistry } from '@hawtiosrc/help'
 import { SpringBoot } from '@hawtiosrc/plugins/springboot/SpringBoot'
```
tadayosi, line 4: should be relative path.

packages/hawtio/src/plugins/springboot/springboot-service.ts
```diff
@@ -1,90 +1,146 @@
 import { jolokiaService, workspace } from '@hawtiosrc/plugins'
-import { HealthComponent, HealthData, JolokiaHealthData, Logger, LoggerConfiguration } from './types'
+import { HealthComponent, HealthData, JmXTrace, JolokiaHealthData, Logger, LoggerConfiguration, Trace } from './types'
 
-export function isActive(): Promise<boolean> {
-  return workspace.treeContainsDomainAndProperties('org.springframework.boot')
-}
+class SpringbootService {
```
tadayosi, line 4: `SpringBootService`?

packages/hawtio/src/plugins/springboot/springboot-service.ts
```diff
@@ -1,90 +1,146 @@
...
+  isSb3 = true
```
tadayosi, line 5: if it's public, let's not use abbreviation. instead, name `isSpringBoot3`

packages/hawtio/src/plugins/springboot/types.ts
```diff
@@ -37,3 +37,65 @@ export type Logger = {
...
+export class Trace {
+  timestamp: string
+  method: string = ''
+  path: string = ''
+  httpStatusCode: number = -1
+  timeTaken: string = '-1'
+  info: string = ''
+
+  constructor(traceData: unknown) {
+    const trace = traceData as JmXTrace
```
tadayosi, line 50: receiving it as `unknown` and casting it to `JmxTrace` here is not a good api. rather, let's cast it before passing it to this constructer and explicitly declare what type it accepts.

packages/hawtio/src/plugins/springboot/types.ts
```diff
@@ -37,3 +37,65 @@ export type Logger = {
...
+      }
+
+      if (trace.response && trace.response.status) {
+        this.httpStatusCode = parseInt(trace.response.status)
+      }
+    }
+  }
+}
+
+export type JmXTrace = {
```
tadayosi, line 81: `JmXTrace` -> `JmxTrace` (small T)

tadayosi (approved):
Thanks. Minor comments but otherwise it looks good.

packages/hawtio/src/plugins/springboot/TraceView.tsx
```diff
@@ -0,0 +1,328 @@
...
+          case 'HTTP Status':
+            filtered = filtered.filter(trace => trace.httpStatusCode.toString().includes(searchTerm))
+            break
+          case 'Path':
+            filtered = filtered.filter(trace => trace.path.includes(searchTerm))
+            break
+          case 'Time Taken':
+            filtered = filtered.filter(trace => trace.timeTaken.includes(searchTerm))
+            break
+        }
```
tadayosi, line 117: no default?

packages/hawtio/src/plugins/springboot/TraceView.tsx
```diff
@@ -0,0 +1,328 @@
...
+  const [perPage, setPerPage] = useState(20)
+  const [searchTerm, setSearchTerm] = useState('')
+  const [filters, setFilters] = useState<string[]>([])
+  const [currentTraceFilter, setCurrentTraceFilter] = useState('Timestamp')
+  const [isFilterDropdownOpen, setIsFilterDropdownOpen] = useState(false)
+  const [isHttpMethodFilterDropdownOpen, setIsHttpMethodFilterDropdownOpen] = useState(false)
+  const [isTraceDetailsOpen, setIsTraceDetailsOpen] = useState(false)
+  const [traceDetails, setTraceDetails] = useState<string>('')
+
+  const handleSearch = useCallback(
```
tadayosi, line 88: why do we need `useCallback`?

mmelko: So when I used handleSearch in the useEffect it required to be as a dependency. But when it's as dependency handleSearch refference is constantly changing and it ends up in the render roop. By using useCallback handleSearch is recreated only when dependencies changes. But I'm thinking about just removing it and moving the logic directly to the useEffect because I'm not using it anywhere else.

packages/hawtio/src/plugins/springboot/TraceView.tsx
mmelko: Probably it's default in my IDE. btw I've noticed if it's in `{}` I have also displayed used color directly in IDE: 

Screenshot 2023-11-23 at 10 53 17

packages/hawtio/src/plugins/springboot/TraceView.tsx
mmelko: I added 'Path' filter as default

packages/hawtio/src/plugins/springboot/TraceView.tsx
tadayosi: Maybe it's better to use PF CSS variables then, but it's ok for now.

tadayosi:
Let's merge. Thanks! @mmelko

PR #681, opened 2023-11-23
fix(stringSorter): Compare lowercase strings to prevent incorrect string ordering.

mmelko:
fix [HAWNG-7](https://issues.redhat.com/browse/HAWNG-7)

### DIFF
--- a/packages/hawtio/src/util/strings.ts
+++ b/packages/hawtio/src/util/strings.ts
@@ -78,13 +78,7 @@ export function trimQuotes(text: string): string {
 }
 
 export function stringSorter(a: string, b: string, sortDesc?: boolean): number {
-  let res = 0
-  if (a < b) {
-    res = -1
-  }
-  if (a > b) {
-    res = 1
-  }
+  let res = a.localeCompare(b)
   if (sortDesc) {
     res *= -1
   }
--- a/packages/hawtio/src/util/strings.test.ts
+++ b/packages/hawtio/src/util/strings.test.ts
@@ -1,4 +1,4 @@
-import { humanizeLabels, matchWithWildcard, parseBoolean, toString, trimQuotes } from './strings'
+import { humanizeLabels, matchWithWildcard, parseBoolean, stringSorter, toString, trimQuotes } from './strings'
 
 describe('strings', () => {
   test('toString', () => {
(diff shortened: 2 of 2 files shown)


oscerd (approved):

tadayosi (commented):
I'm not yet convinced that it's the right fix.

* This function is used in many places. Have you checked that this change doesn't cause any ill behaviours in other places?  I'm concerned about broad effect on this change.
* Even looking at HAWNG-7 I'm still not sure what's really the issue, since comparing strings in a case sensitive way should be totaly fine. Have you checked it with @jsolovjo on what's really the issue?

tadayosi (changes requested):
Let me block merge of this until we can resolve the remaining questions on it.

packages/hawtio/src/util/strings.ts
```diff
@@ -79,10 +79,10 @@ export function trimQuotes(text: string): string {
 
 export function stringSorter(a: string, b: string, sortDesc?: boolean): number {
   let res = 0
-  if (a < b) {
+  if (a.toLowerCase() < b.toLowerCase()) {
     res = -1
   }
-  if (a > b) {
+  if (a.toLowerCase() > b.toLowerCase()) {
```
tadayosi, line 85: I'm not yet convinced that it's the right fix.

* This function is used in many places. Have you checked that this change doesn't cause any ill behaviours in other places?  I'm concerned about broad effect on this change.
* Even looking at HAWNG-7 I'm still not sure what's really the issue, since comparing strings in a case sensitive way should be totaly fine. Have you checked it with @jsolovjo on what's really the issue?

mmelko: Problem is that when you have case-sensitive strings for the ordering you might get wrong order because  `>` is comparing ASCII values. 

But I will change  this to `localCompare`:

```typescript
    expect('b' > 'A').toBe(true) // pass
    expect('B' > 'A').toBe(true) //pass
    expect('B' > 'a').toBe(true) //fail
    expect('b' > 'a').toBe(true) //pass

    expect('B'.localeCompare('a')).toBe(1) //pass
    expect('b'.localeCompare('a')).toBe(1) //pass
    expect('B'.localeCompare('A')).toBe(1) //pass
    expect('b'.localeCompare('A')).toBe(1) //pass 
    
```

So I think this will fix the sorting issue broadly not only the Threads table.

tadayosi (approved):
lgtm. could you write tests for the new stringSorter?

PR #682, opened 2023-11-24
refactor(Runtime-plugin): refactor filtering in Threads and SysProps components

mmelko:
This PR is essentially move `handleSearch` logic into the separate `useEffect`. This reduces little bit a complexity and improves readability.


tadayosi (approved):
great!

PR #703, opened 2023-12-06
feat: add URI encoding and additional check for the redirect to prevent XSS

mmelko:
1. I added additional check of the protocol to be only `http` or `https` to ensure no suspicious "protocols" as `javascript, livescript` will be executed 
2. add double check if hostname and port for the redirect is the same as origin
3.  added aditional encodeURI for the url
4. added decodeURI when retrieving `window.location.href` 

I tested this with both remote connect to localhost and to other host on the lan.

### DIFF
--- a/packages/hawtio/src/plugins/shared/connect-service.ts
+++ b/packages/hawtio/src/plugins/shared/connect-service.ts
@@ -243,9 +243,32 @@ class ConnectService implements IConnectService {
    */
   redirect() {
     const url = new URL(window.location.href)
-    const redirect = url.searchParams.get(PARAM_KEY_REDIRECT) ?? hawtio.getBasePath() ?? '/'
+    let redirect = url.searchParams.get(PARAM_KEY_REDIRECT) ?? '/'
+    let safeRedirect: boolean = false
+
+    try {
+      const { hostname, port, protocol, searchParams } = new URL(redirect)
+      const connectionKey = searchParams.get(PARAM_KEY_CONNECTION) ?? ''
+      safeRedirect =
+        hostname === url.hostname &&
+        port === url.port &&
+        ['http:', 'https:'].includes(protocol) &&
+        connectionKey !== '' &&
+        connectionKey === this.currentConnection
+    } catch (_e) {
+      log.error('Invalid URL')
+      eventService.notify({
+        type: 'danger',
+        message: 'Redirect parameter was modified',
+      })
+    }
+
+    if (!safeRedirect) {
+      redirect = hawtio.getBasePath() ?? '/'
+    }
+
     log.debug('Redirect to:', redirect)
-    window.location.href = redirect
+    window.location.href = encodeURI(redirect)
   }
 
   /**


oscerd (approved):

packages/hawtio/src/plugins/shared/connect-service.ts
```diff
@@ -52,21 +52,37 @@ const LOGIN_PATH = '/connect/login'
 
 export interface IConnectService {
   getCurrentConnectionName(): string | null
+
```
tadayosi, line 55: I'd prefer not to include this as it's unrelated to the fix.

packages/hawtio/src/plugins/shared/connect-service.ts
```diff
@@ -244,8 +260,15 @@ class ConnectService implements IConnectService {
...
-    log.debug('Redirect to:', redirect)
-    window.location.href = redirect
+    const { hostname, port, protocol } = new URL(redirect)
+
+    //check if it's correct redirect:
+    if (hostname === url.hostname && port === url.port && ['http:', 'https:'].includes(protocol)) {
+      log.debug('Redirect to:', redirect)
+      window.location.href = encodeURI(redirect)
+    } else {
+      log.error('Redirect param modified')
```
tadayosi, line 270: writing error log is great, but maybe then redirect to `/` (default) or showing the error message to the user with notification?

packages/hawtio/src/plugins/shared/jolokia-service.ts
```diff
@@ -312,14 +312,15 @@ class JolokiaService implements IJolokiaService {
       switch (xhr.status) {
         case 401:
         case 403: {
-          const url = new URL(window.location.href)
+          const url = new URL(decodeURIComponent(window.location.href))
```
tadayosi, line 315: why do we need this?  Isn't it enough to encode the value assigned to `window.location.href` below?

packages/hawtio/src/plugins/shared/jolokia-service.ts
```diff
@@ -312,14 +312,15 @@ class JolokiaService implements IJolokiaService {
...
           // If window was opened to connect to remote Jolokia endpoint
           if (url.searchParams.has(PARAM_KEY_CONNECTION)) {
             const loginPath = connectService.getLoginPath()
+
             if (url.pathname !== loginPath) {
               // ... and not showing the login modal
               this.jolokia?.then(jolokia => jolokia.stop())
-              const redirectUrl = window.location.href
+              const redirectUrl = decodeURIComponent(window.location.href)
```
tadayosi, line 323: same here

instead, shouldn't we encode `url.href` below?

tadayosi (commented):
A few comments that are not clear to me.

packages/hawtio/src/plugins/shared/jolokia-service.ts
mmelko: my understanding is that URL constructor isn't trying to decode the URL so if the URL is encoded this intentionally can throw an exception. If The URL isn't encoded  decode function returns same string. So there is no harm.

tadayosi: > " if the URL is encoded this intentionally can throw an exception."

The only encoded part should be query if there are any, so why do we need to prepare for the case?   Is there a valid case where the entire URL may be encoded?

mmelko: you're right probably not. At least I haven't found any and I just wanted to be sure.  So I will remove it.

packages/hawtio/src/plugins/shared/connect-service.ts
```diff
@@ -242,10 +242,27 @@ class ConnectService implements IConnectService {
    * Redirect to the URL specified in the query parameter {@link PARAM_KEY_REDIRECT}.
    */
   redirect() {
-    const url = new URL(window.location.href)
-    const redirect = url.searchParams.get(PARAM_KEY_REDIRECT) ?? hawtio.getBasePath() ?? '/'
+    const url = new URL(decodeURI(window.location.href))
```
tadayosi, line 245: This one as well. Why do we need decode here?

mmelko: because we are sending encoded URL from the jolokia-service: https://github.com/mmelko/hawtio-next/blob/600af730d342d388244afa01973562a92a72188a/packages/hawtio/src/plugins/shared/jolokia-service.ts#L326

tadayosi (approved):

PR #723, opened 2023-12-19
fix(springboot-plugin):fix exceptions when rendering components

mmelko:
This PR is fixing https://github.com/hawtio/hawtio-online/issues/270

by refactoring the Health component in the spring-boot plugin. Previously there was a `Record` to determine the corrent spans for the components, however in the openshift there is more probe components that was throwing exception. 

Now I introduced just list of the components that have 'fixed' position and rest is rendered underneath:
image

image

image

### DIFF
--- a/packages/hawtio/src/plugins/springboot/Health.tsx
+++ b/packages/hawtio/src/plugins/springboot/Health.tsx
@@ -1,16 +1,5 @@
 import React, { useEffect, useState } from 'react'
-import {
-  Card,
-  CardBody,
-  CardHeader,
-  Flex,
-  FlexItem,
-  Grid,
-  GridItem,
-  gridSpans,
-  PageSection,
-  Title,
-} from '@patternfly/react-core'
+import { Card, CardBody, CardHeader, Flex, FlexItem, Grid, GridItem, PageSection, Title } from '@patternfly/react-core'
 import { springbootService } from './springboot-service'
 import { HealthComponentDetail, HealthData } from './types'
 import { TableComposable, Tbody, Td, Tr } from '@patternfly/react-table'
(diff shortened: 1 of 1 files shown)


oscerd (approved):

packages/hawtio/src/plugins/springboot/Health.tsx
```diff
@@ -24,12 +13,8 @@ import {
   QuestionCircleIcon,
 } from '@patternfly/react-icons'
 
-const componentSpanRecord: Record<string, number[]> = {
-  diskSpace: [7, 2],
-  ping: [5, 0],
-  camelHealth: [5, 1],
-  camel: [5, 1],
-}
+const HEALTH_COMPONENTS = ['diskSpace', 'camelHealth', 'camel']
```
phantomjinx, line 16: Would this benefit from a more explicit name. HEALTH_COMPONENTS seems a little ambiguous given all of the compoments have come from `loadHealth` in the springboot-service?

packages/hawtio/src/plugins/springboot/Health.tsx
```diff
@@ -118,33 +103,53 @@ export const Health: React.FunctionComponent = () => {
...
-                        <ComponentDetails componentDetails={component.details} />
-                      ))}
-                  </Flex>
-                </CardBody>
-              </Card>
-            </GridItem>
-          ))}
+          {healthData?.components.map(
+            component =>
+              HEALTH_COMPONENTS.includes(component.name) && (
```
phantomjinx, line 108: You map everything here and then effectively use an `if` condition to decide whether display the component. However, underneath you filter and them map. Would it be better to use a uniform method across both or do a single map of all components then pick one component choice or other depending on the `includes` ?

phantomjinx (approved):
It solves the problem nicely but maybe consider my comments to further tighten it up?

packages/hawtio/src/plugins/springboot/Health.tsx
mmelko: yeah I know what do you mean. This is more-less a substitute for a simple sorting e.g health and camel on top with fixed span and rest on bottom but I will try to sort them.

PR #789, opened 2024-02-20
fix(core): cannot add productInfo to hawtconfig when about is absent

### LINKED ISSUE #715: ConfigManager expects at least an empty [] in hawtconfig.json
A `productInfo` property is added to most implementations of [hawtconfig.json](https://github.com/hawtio/hawtio-next/blob/main/app/public/hawtconfig.json#L22) to add additional version information that appears in the About component. It turns out that this array must exist in the hawtconfig.json, even if [empty](https://github.com/hawtio/hawtio-online/blob/main/packages/online-shell/public/hawtconfig.json#L4). Otherwise, any version information added using '[addProductInfo](https://github.com/hawtio/hawtio-next/blob/main/packages/hawtio/src/core/config-manager.ts#L211)' is not actually added.

The [About](https://github.com/hawtio/hawtio-next/blob/main/packages/hawtio/src/ui/about/HawtioAbout.tsx#L36) component handles the possibility of the `productInfo` array not existing by providing its own empty array which renders the version information in the display blank.

Suggestions for fixes:
- `loadConfig` in [ConfigManager.ts](https://github.com/hawtio/hawtio-next/blob/main/packages/hawtio/src/core/config-manager.ts#L132) should initialize an empty `productInfo` array
- `addProductInfo` in [ConfigManager.ts](https://github.com/hawtio/hawtio-next/blob/main/packages/hawtio/src/core/config-manager.ts#L211) should check for a null `productInfo` property and should initialize if necessary

tadayosi:
Fix #715

### DIFF
--- a/packages/hawtio/src/core/config-manager.test.ts
+++ b/packages/hawtio/src/core/config-manager.test.ts
@@ -159,4 +159,22 @@ describe('ConfigManager', () => {
     expect(product3?.name).toEqual('Hawtio React')
     expect(product3?.value).toEqual('1.0.0')
   })
+
+  test('adds product info to empty config', async () => {
+    // response for fetching hawtconfig.json
+    fetchMock.mockResponse(JSON.stringify({}))
+
+    configManager.addProductInfo('Hawtio React', '1.0.0')
+    const config = await configManager.getHawtconfig()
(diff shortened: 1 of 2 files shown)


phantomjinx (approved):
LGTM

PR #794, opened 2024-02-23
Azure OIDC wip

grgrzybek:
@tadayosi It is a big one, but I think I got it right.
Initially I was (wrongly) doing the OAuth2 in `useUser() -> fetchUser()` but then I realized I had to do it in the constructor (just like it works with Keycloak).

We can talk this over next week.

### DIFF
--- a/app/src/bootstrap.tsx
+++ b/app/src/bootstrap.tsx
@@ -7,7 +7,7 @@ import { reportWebVitals } from './reportWebVitals'
 // Configure the console
 const configure = () => {
   configManager.addProductInfo('Test App', '1.0.0')
-  hawtio.addUrl('plugin')
+  //COMMENT_FOR_NOW hawtio.addUrl('plugin')
 }
 configure()
 
--- a/packages/hawtio/package.json
+++ b/packages/hawtio/package.json
@@ -59,6 +59,7 @@
     "js-logger": "^1.6.1",
     "keycloak-js": "^23.0.7",
     "monaco-editor": "^0.46.0",
+    "oauth4webapi": "^2.10.3",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-markdown": "^8.0.7",
--- a/packages/hawtio/package.json
+++ b/packages/hawtio/package.json
@@ -57,6 +57,7 @@
     "jolokia.js": "^2.0.0",
     "jquery": "^3.7.1",
     "js-logger": "^1.6.1",
+    "jwt-decode": "^4.0.0",
     "keycloak-js": "^23.0.7",
     "monaco-editor": "^0.46.0",
     "oauth4webapi": "^2.10.3",
--- a/packages/hawtio/src/auth/hooks.ts
+++ b/packages/hawtio/src/auth/hooks.ts
@@ -8,20 +8,35 @@ export function useUser() {
   const [username, setUsername] = useState('')
   const [isLogin, setIsLogin] = useState(false)
   const [userLoaded, setUserLoaded] = useState(false)
+  // special, temporary status of user fetching operation. Required to avoid flickering when OAuth2
+  // redirect login is pending
+  const [userLoading, setUserLoading] = useState(true)
 
   useEffect(() => {
+    const abortController = new AbortController()
+    let proceed = true
+    const isProceed = () => proceed
     const fetchUser = async () => {
       // Try syncing the login status with the server here
-      await userService.fetchUser()
+      await userService.fetchUser(true, abortController.signal, () => isProceed())
 
       const username = await userService.getUsername()
       const isLogin = await userService.isLogin()
-      setUsername(username)
-      setIsLogin(isLogin)
-      setUserLoaded(true)
+      const isLoading = await userService.isLoading()
+      if (isProceed()) {
+        setUsername(username)
+        setIsLogin(isLogin)
+        setUserLoading(isLoading)
+        setUserLoaded(true)
+      }
     }
     fetchUser()
+
+    return () => {
+      abortController.abort()
+      proceed = false
+    }
   }, [])
 
-  return { username, isLogin, userLoaded }
+  return { username, isLogin, userLoaded, userLoading }
 }
--- a/packages/hawtio/src/auth/hooks.ts
+++ b/packages/hawtio/src/auth/hooks.ts
@@ -13,12 +13,11 @@ export function useUser() {
   const [userLoading, setUserLoading] = useState(true)
 
   useEffect(() => {
-    const abortController = new AbortController()
     let proceed = true
     const isProceed = () => proceed
     const fetchUser = async () => {
       // Try syncing the login status with the server here
-      await userService.fetchUser(true, abortController.signal, () => isProceed())
+      await userService.fetchUser(true, () => isProceed())
 
       const username = await userService.getUsername()
       const isLogin = await userService.isLogin()
@@ -33,7 +32,6 @@ export function useUser() {
     fetchUser()
 
     return () => {
-      abortController.abort()
       proceed = false
     }
   }, [])
--- a/packages/hawtio/src/auth/user-service.ts
+++ b/packages/hawtio/src/auth/user-service.ts
@@ -4,16 +4,17 @@ import { log, PATH_LOGOUT, PATH_USER, PUBLIC_USER } from './globals'
 type User = {
   username: string
   isLogin: boolean
+  isLoading: boolean
 }
 
 export type ResolveUser = (user: User) => void
-export type FetchUserHook = (resolve: ResolveUser) => Promise<boolean>
+export type FetchUserHook = (resolve: ResolveUser, signal: AbortSignal | null, proceed: (() => boolean) | null) => Promise<boolean>
 export type LogoutHook = () => Promise<boolean>
 
 export interface IUserService {
   addFetchUserHook(name: string, hook: FetchUserHook): void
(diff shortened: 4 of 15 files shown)


grgrzybek:
BTW I think with the throttling enabled I saw problems with multiple Bearer token authentications, but I have to investigate more...

grgrzybek:
I'm not sure how to handle this problem:
```
/data/sources/github.com/hawtio/hawtio-next/node_modules/oauth4webapi/build/index.js:19
export const clockSkew = Symbol();
^^^^^^

SyntaxError: Unexpected token 'export'

  2 | import { Logger } from "@hawtiosrc/core"
  3 | import { jwtDecode } from "jwt-decode"
> 4 | import * as oidc from "oauth4webapi"
    | ^
  5 | import { AuthorizationServer, Client, OAuth2Error } from "oauth4webapi"
  6 | import { fetchPath } from "@hawtiosrc/util/fetch"
  7 | import $ from "jquery"
```

packages/hawtio/src/plugins/auth/oidc/index.ts
```diff
@@ -0,0 +1,16 @@
...
+import { helpRegistry } from '@hawtiosrc/help/registry'
+import help from './help.md'
+import { oidcService } from './oidc-service'
+import { Logger } from '@hawtiosrc/core'
+
+const pluginName = 'hawtio-oidc'
+const log = Logger.get(pluginName)
+
+const oidc: HawtioPlugin = () => {
+  log.info('Loading oidc() plugin')
```
tadayosi, line 11: that's something we don't normally do. we can check whether a plugin is loaded or not by checking the hawtio core log.

packages/hawtio/src/plugins/index.ts
```diff
@@ -17,6 +18,7 @@ import { springboot } from './springboot'
 export const registerPlugins: HawtioPlugin = () => {
   // Auth plugins should be loaded before other plugins
   keycloak()
+  oidc()
```
tadayosi, line 21: Please also look at the line 34 and add `oidc` to the export there. Since we decided to export each plugin entry to users so that they can assemble their own hawtio console with selected builtin plugins. Some might want to use only `oidc` but not `keycloak` ;-)

packages/hawtio/tsconfig.json
```diff
@@ -20,6 +20,8 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
+    "suppressImplicitAnyIndexErrors": true,
+    "ignoreDeprecations": "5.0",
```
tadayosi, line 24: It looks a bit alarming. Why do we need these?

packages/hawtio/src/ui/page/HawtioPage.tsx
```diff
@@ -17,22 +17,26 @@ import { PageContext } from './context'
...
-  if (!userLoaded || !pluginsLoaded) {
+  // navigate should be used in effect
+  // otherwise "Cannot update a component (`BrowserRouter`) while rendering a different component" is thrown
+  useEffect(() => {
+    if (!isLogin && !userLoading) {
+      navigate('login')
+    }
+  }, [isLogin, navigate, userLoading])
+
+  if (!userLoaded || !pluginsLoaded || userLoading) {
```
tadayosi, line 34: I still don't get why we start to need this change. We've been doing this without that error and `useEffect()`. What changes made it to be needed?

tadayosi (commented):
Overall, it looks great!

There are some minor comments, some of whch are just for my understanding on the changes.

app/src/bootstrap.tsx
```diff
@@ -7,7 +7,7 @@ import { reportWebVitals } from './reportWebVitals'
 // Configure the console
 const configure = () => {
   configManager.addProductInfo('Test App', '1.0.0')
-  hawtio.addUrl('plugin')
+  //COMMENT_FOR_NOW hawtio.addUrl('plugin')
```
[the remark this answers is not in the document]
grgrzybek: With the plugin enabled, I saw duplicate initialization and script loading - and it's not related to React's `<StrictMode>` - I saw webpack mapping the same script with two paths:
```
__webpack_require__(/*! @hawtio/react */                    "../packages/hawtio/dist/index.js")
...
__webpack_require__(/*! ../packages/hawtio/dist/index.js */ "../packages/hawtio/dist/index.js")
```

But it was only a problem when I made huge mistake by doing OAuth2 login within `useUser` (which is `useEffect` underneath). I'll revert this change.

packages/hawtio/package.json
```diff
@@ -57,8 +57,10 @@
     "jolokia.js": "^2.0.0",
     "jquery": "^3.7.1",
     "js-logger": "^1.6.1",
+    "jwt-parser": "^2.0.0",
```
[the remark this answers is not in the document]
grgrzybek: Hmm, I was going to use (and I'm using it...) https://www.npmjs.com/package/jwt-decode...

packages/hawtio/src/auth/hooks.ts
```diff
@@ -8,20 +8,35 @@ export function useUser() {
   const [username, setUsername] = useState('')
   const [isLogin, setIsLogin] = useState(false)
   const [userLoaded, setUserLoaded] = useState(false)
+  // special, temporary status of user fetching operation. Required to avoid flickering when OAuth2
+  // redirect login is pending
+  const [userLoading, setUserLoading] = useState(true)
```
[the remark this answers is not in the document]
grgrzybek: Looks like this line was important - when I removed it, I had constant redirects (in React router) to `/login` in HawtioPage.tsx... I had to keep userLoading and this condition:
```
useEffect(() => {
  if (!isLogin && !userLoading) {
    navigate('login')
  }
}, [isLogin, navigate, userLoading])
```

packages/hawtio/src/auth/user-service.ts
```diff
@@ -4,16 +4,21 @@ import { log, PATH_LOGOUT, PATH_USER, PUBLIC_USER } from './globals'
...
   isLogin: boolean
+  isLoading: boolean
 }
 
 export type ResolveUser = (user: User) => void
-export type FetchUserHook = (resolve: ResolveUser) => Promise<boolean>
+export type FetchUserHook = (
+  resolve: ResolveUser,
+  signal: AbortSignal | null,
+  proceed: (() => boolean) | null,
```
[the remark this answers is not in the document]
grgrzybek: Sure - I'm still learning TS :)

packages/hawtio/src/auth/user-service.ts
```diff
@@ -47,10 +52,17 @@ class UserService implements IUserService {
   /**
    * Sync login status with the server by fetching login user.
    */
-  async fetchUser(retry = true): Promise<void> {
+  async fetchUser(
+    retry = true,
+    signal: AbortSignal | null = null,
+    proceed: (() => boolean) | null = null,
```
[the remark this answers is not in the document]
grgrzybek: done.

packages/hawtio/src/auth/user-service.ts
```diff
@@ -98,6 +110,12 @@ class UserService implements IUserService {
     return (await this.user).isLogin
   }
 
+  async isLoading(): Promise<boolean> {
```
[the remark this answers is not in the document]
grgrzybek: I switched to return boolean directly and kept async, so a promise is returned - now it's using `await` instead of promise-then.

packages/hawtio/src/auth/user-service.ts
```diff
@@ -98,6 +110,12 @@ class UserService implements IUserService {
...
+    return this.user.then(u => {
+      return u.isLoading || false
```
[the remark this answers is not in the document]
grgrzybek: done

packages/hawtio/src/plugins/auth/oidc/index.ts
grgrzybek: done

packages/hawtio/src/plugins/index.ts
grgrzybek: done

packages/hawtio/tsconfig.json
grgrzybek: oh wait - it must've been unnecessary `git add`...

packages/hawtio/src/ui/page/HawtioPage.tsx
grgrzybek: I was getting React error like this: "state modified when running component function".
`navigate('login')` is in fact updating state (react router) and I found that state should be updated in event handlers or effects.
But of course I may be wrong - I'm not (yet?) expert on async React... I'll monitor this fragment - anyway, you know where to find me ;)

grgrzybek:
Hmm, I tried different things with babel, jest and transformations (and ignoring it) and nothing...
* https://stackoverflow.com/questions/76428538/jest-syntaxerror-unexpected-token-export-oauth4webapi
* https://stackoverflow.com/questions/70883026/jest-and-babel-transpilation-syntaxerror-cannot-use-import-statement-outside
* https://stackoverflow.com/questions/49263429/jest-gives-an-error-syntaxerror-unexpected-token-export

tadayosi:
@grgrzybek What's the current status of this pull req?  What's now blocking it from merging?

grgrzybek:
Nothing blocks me ;) 3 conversations are not resolved - may I resolve them?

Issue #754, opened 2024-01-24
Consider lazy loading of Camel models


tadayosi:
Since we support multiple versions of Camel model #693, we need to bundle all the model defintions in the hawtio/react distribution. We lately found that it's bloating the distribution files and thus decreasing the performance of initial loading of the console.

One possible solution to deal with the performance issue would be to introduce lazy loading for the Camel models, so that they are not fetched from the server until it's really necessary and even that unused models are never loaded.

PR #801, opened 2024-03-04
Draft: Setup for improving load times of Hawtio Online

mmuzikar:
Exposed UI components that could be used for unified look of the app (I will use HawtioLoadingPage in Hawtio Online).
Lazy loading Camel models, the fetch starts on the module load and since the both model packages are not included in the resulting bundle it significantly improves the load times.

### DIFF
--- a/packages/hawtio/src/index.ts
+++ b/packages/hawtio/src/index.ts
@@ -8,6 +8,7 @@ export * from './core'
 export * from './help'
 export * from './plugins'
 export * from './preferences'
+export * from './ui'
 
 // Register Hawtio React component version
 configManager.addProductInfo('Hawtio React', '__PACKAGE_VERSION_PLACEHOLDER__')
--- a/packages/hawtio/src/plugins/camel/camel-service.ts
+++ b/packages/hawtio/src/plugins/camel/camel-service.ts
@@ -1,5 +1,3 @@
-import * as camel3 from '@hawtio/camel-model-v3'
-import * as camel4 from '@hawtio/camel-model-v4'
 import { eventService } from '@hawtiosrc/core'
 import { MBeanNode } from '@hawtiosrc/plugins/shared'
 import { jolokiaService } from '@hawtiosrc/plugins/shared/jolokia-service'
@@ -22,6 +20,9 @@ import {
   routesType,
 } from './globals'
 import { ROUTE_OPERATIONS } from './routes-service'
+
+let _camel3: CamelModel | null = null
+let _camel4: CamelModel | null = null
 
 // TODO: Should be provided by @hawtio/camel-model package
 export type CamelModel = {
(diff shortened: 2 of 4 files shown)


packages/hawtio/src/plugins/camel/camel-service.ts
```diff
@@ -333,6 +335,7 @@ export function getCamelVersions(): string[] {
  * the given node. Currently, it supports Camel v3 and v4.
  */
 export function getCamelModel(node: MBeanNode): CamelModel {
+  const { camel4, camel3 } = fetchCamelModules()
```
phantomjinx, line 338: When running I get an error that camel3 is null

phantomjinx (changes requested):

packages/hawtio/src/index.ts
```diff
@@ -8,6 +8,7 @@ export * from './core'
 export * from './help'
 export * from './plugins'
 export * from './preferences'
+export * from './ui'
```
tadayosi, line 11: Why do we need to export `ui`?

packages/hawtio/src/plugins/camel/camel-service.ts
tadayosi: Yeah, rather they should be `Promise` so that they will never return `null`.

packages/hawtio/src/plugins/camel/camel-service.ts
```diff
@@ -23,6 +21,9 @@ import {
 } from './globals'
 import { ROUTE_OPERATIONS } from './routes-service'
 
+let _camel3: CamelModel | null = null
+let _camel4: CamelModel | null = null
```
tadayosi, line 25: so if we change them to `Promise` them can be `const` not `let`.

packages/hawtio/src/plugins/camel/camel-service.ts
```diff
@@ -397,3 +400,23 @@ export function isCamelVersionEQGT(node: MBeanNode, major: number, minor: number
...
+function fetchCamelModules(): { camel3: CamelModel; camel4: CamelModel } {
+  if (!_camel3 || !_camel4) {
+    ;(async () => await camelModelPromises)()
+  }
+
+  return {
+    camel3: _camel3!,
+    camel4: _camel4!,
+  }
+}
```
tadayosi, line 422: I think this async -> await doesn't work as you expect. Async call is blocked only inside if, which leads to null pointer exception as Paul mentioned because the if block finishes immediately and tries to return the values no matter what values `_camel3` and `_camel4` have.

packages/hawtio/src/ui/index.ts
```diff
@@ -0,0 +1,5 @@
+export * from './about'
+export * from './icons'
+export * from './login'
+export * from './notification'
+export * from './page'
```
tadayosi, line 5: Why are they needed?  Note that it will change the Hawtio API we expose to external plugin authors. So generally it needs to be justified with a clear reason.

tadayosi (changes requested):

tadayosi:
@mmuzikar Also, this pull req should be intended to fix #754. If so, please don't forget to mention that it fixes #754.

tadayosi:
Should be fixed with #898. Let's close it.

PR #818, opened 2024-03-14
refactor(CLUX-495,496): Use the TableComposable for attributes and primary detail pattern instead of modal

mmelko:
- TableComposable for the attributes list 
  - `Update`changed to `Save` 
-  Used primary detail pattern instead of modal: 
  - Make attributes table scrollable so whole page isn't scrolling with the table 
Screenshot 2024-03-19 at 21 13 26

### DIFF
--- a/packages/hawtio/src/plugins/shared/attributes/AttributeModal.tsx
+++ b/packages/hawtio/src/plugins/shared/attributes/AttributeModal.tsx
@@ -75,18 +75,20 @@ export const AttributeModal: React.FunctionComponent<{
 
   const modalTitle = `Attribute: ${attributeName}`
 
-  const modalActions = [
-    <Button key='close' variant='primary' onClick={onClose}>
-      Close
-    </Button>,
-  ]
+  const modalActions = []
+
   if (isWritable) {
     modalActions.push(
       <Button key='update' variant='danger' onClick={updateAttribute}>
-        Update
+        Save
       </Button>,
     )
   }
+  modalActions.push(
+    <Button key='close' variant='link' onClick={onClose}>
+      Cancel
+    </Button>,
+  )
 
   return (
     <Modal variant='medium' title={modalTitle} isOpen={isOpen} onClose={onClose} actions={modalActions}>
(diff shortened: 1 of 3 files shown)


tadayosi:
@mmelko What's the current status of it?  If you're ready, please remove the `WIP` in the title as it's confusing.

tadayosi:
Closing as will be followed up with a newer PR.

packages/hawtio/src/plugins/shared/attributes/AttributeModal.tsx
```diff
@@ -75,18 +75,20 @@ export const AttributeModal: React.FunctionComponent<{
...
     modalActions.push(
       <Button key='update' variant='danger' onClick={updateAttribute}>
-        Update
+        Save
       </Button>,
     )
   }
+  modalActions.push(
+    <Button key='close' variant='link' onClick={onClose}>
+      Cancel
```
tadayosi, line 89: As I commented, `Cancel` doesn't make much sense to me. Anyway, isn't this part changed altogether with a following pull req?  Why do we bother changing this part then?

packages/hawtio/src/plugins/shared/attributes/Attributes.tsx
```diff
@@ -81,10 +78,28 @@ export const Attributes: React.FunctionComponent = () => {
 
   return (
     <Card isFullHeight>
-      <Table aria-label='Attributes' variant='compact' cells={columns} rows={rows}>
-        <TableHeader />
-        <TableBody onRowClick={selectAttribute} />
-      </Table>
+      {/*<DataListClickableRows />*/}
```
tadayosi, line 81: what's this comment?  shouldn't we remove it?

tadayosi:
@mmelko Can't the detail drawer overlap the attributes table?  The attached screenshot seems to shift the table instead of overlapping it.

mmelko:
> @mmelko Can't the detail drawer overlap the attributes table? The attached screenshot seems to shift the table instead of overlapping it.

It's overlaying only when the table is too tight. I tried both options but this with inline details looks better to me.

tadayosi:
@mmelko rebase it please

tadayosi:
So, the only failing E2E tests are related to the Attributes view changes and should be fixed with hawtio/hawtio#3326. Let's merge it now.

mmelko:
thanks!

PR #828, opened 2024-03-20
Fix E2E workflow & improve login form UX

mmuzikar:
With Hawtio 4.x we have app images only for Java 21 and 17, so 11 was removed from the testing matrix. 
Also the login form when connecting now is autofocused so user can login just from their keyboard + use Enter as submit. 
(Enter as submit should work automatically in forms, but the way modals are implemented the submit button is outside the form element so it doesn't trigger the submit event)

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -32,7 +32,7 @@ jobs:
       fail-fast: false
       matrix:
         node: ['18', '20']
-        java: ['11', '17']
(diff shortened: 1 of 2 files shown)


tadayosi (approved):
Thanks!! Let's see if it really fixes e2e.

tadayosi:
@mmuzikar Unfortunately E2E is still failing: https://github.com/hawtio/hawtio-next/actions/runs/8368402266

PR #834, opened 2024-03-23
fix(ci): disable OIDC for app

tadayosi:
Potentially fixes the E2E failures.

### DIFF
--- a/app/craco.config.js
+++ b/app/craco.config.js
@@ -168,7 +168,7 @@ module.exports = {
         res.redirect('/hawtio/login')
       })
 
-      const oidcEnabled = true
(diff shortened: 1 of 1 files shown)


grgrzybek:
is `app` used at all in e2e tests?
Regardless - setting `oidcEnabled` to true shouldn't have been pushed - my bad.

tadayosi:
@grgrzybek

> is app used at all in e2e tests?

Yes. https://github.com/hawtio/hawtio-next/blob/main/docs/e2e.md
But it didn't fix E2E anyway. I'm also looking into the E2E issue.

Issue #624, opened 2023-10-19
Improve Chart


tadayosi:
The current implementation of Chart tab in JMX plugin is not so nice. For instance, we can improve the loading time of the tab. We need to polish it before releasing 1.0.0.

![image](https://github.com/hawtio/hawtio-next/assets/156692/2bc44f2d-f4cf-4656-91da-fa295cd372a4)

Subtasks:
- [ ] Improve load time of Chart tab
- [x] https://github.com/hawtio/hawtio-next/issues/437
- [x] https://github.com/hawtio/hawtio-next/issues/489

Issue #489, opened 2023-08-04
Chart X Axis labels are not correct


lhein:
While the old HawtIO chart had a X Axis labeling of <hh:mm> to represent the current time the new chart has a label <mm:ss> which is not helpful. 

This issue is linked to #10 .

lhein:
Old HawtIO Screenshot:
![grafik](https://github.com/hawtio/hawtio-next/assets/435853/03d0b537-e800-496e-854f-13ac36712f8c)

New HawtIO Screenshot:
![grafik](https://github.com/hawtio/hawtio-next/assets/435853/34508e66-04dc-41ec-915a-a5eaa63893d1)

mmuzikar:
Probably a duplicate of  #437

jsolovjo:
I agree that the current view is not in the best shape. It would be great to have the values displayed without hovering and having an Y-axis would be great as well

Issue #437, opened 2023-07-18
Broken views in Charts


jsolovjo:
There are several issues with Charts

1. Time is missing hours

2. Missing values - you have to hover a mouse pointer to view a value

3. Missing an Y-Axis in the charts - it could help to see the values immediately

<img width="1728" alt="Screenshot 2023-07-18 at 10 02 23" src="https://github.com/hawtio/hawtio-next/assets/49488040/b2430979-bf2b-4ce9-ad08-b69013ddb57e">

lhein:
@joshiraez could you have a look at that issue once you are back?

lhein:
this issue relates to #10

joshiraez:
I'm on sick leave right now but I can address the questions.

1) Hours are not really worth it as most data gets wiped out after you get out of the screen. I tried having it but it clutters the timeline a lot, and most times you know that it's in your time proximity.

This limitation it's similar to the original, where data would get wiped out in a lot of instances - at least, now, it should be consistent in that you should have all the data while you are on the view.

Being able to pull more data was one of my original design goals at first but I was asked to just mimic the old hawtIO behaviour instead of improving it, so I could just make sure that it was at least consistent while you were on the view.

Adding the hours is easy enough if you feel like it's really needed. It's just that you could see that as soon as we have a lot of entries, you'll have very little ticks showing the times and you could most easily guess those.

On top of that, on hover you will actually see the full time with hours along with the values.

2) Yes. This is how it worked in the original where you had to hover over a point to see the actual values. How would you like those being shown? Although keep in mind that patternfly charts have some limitations, I would have to ask their team and Victory charts again, but it can be looked at.

3) Because the domains for the data are so wildly varying, adding the Y axis usually resulted in something like this.

- 389274987218937213
- 148390821309809482 
- 0

Which, on top of wasting a lot of space again, I felt like they didn't add any value. This is consistent with the original as well, as you don't have Y axis in none of the charts, only having a common time x-axis for every one.

If you'd like to have a diferent design, please tell me your suggestions and we can take a look at it in a different ticket to improve into it. Sorry for not being of more help on this, as most of the things are actually expected and takeovers from the original version.

PR #846, opened 2024-03-28
refactor(charts): Refactor Chart page

mmelko:
fix: #437
fix: #489
fix: #624 
In addition I added a possibility to show only charts for attributes with values actually changing in time. I also added dynamic  min value for y axis to actually show some scale. Charts should be responsive now.
Improved loading performance
Screenshot 2024-03-28 at 22 11 59

`branch: mmuzikar:mmuzikar-ci-test-branch`

### DIFF
--- a/packages/hawtio/src/plugins/shared/chart/Chart.tsx
+++ b/packages/hawtio/src/plugins/shared/chart/Chart.tsx
@@ -2,39 +2,107 @@ import { PluginNodeSelectionContext } from '@hawtiosrc/plugins/context'
 import { AttributeValues } from '@hawtiosrc/plugins/shared/jolokia-service'
 import { MBeanNode } from '@hawtiosrc/plugins/shared/tree'
 import { isNumber } from '@hawtiosrc/util/objects'
-import { ChartArea, ChartAxis, Chart as ChartDraw, ChartVoronoiContainer } from '@patternfly/react-charts'
+import {
+  ChartArea,
+  ChartAxis,
+  Chart as ChartDraw,
+  ChartVoronoiContainer,
+  getResizeObserver,
+} from '@patternfly/react-charts'
 import {
   Button,
   Card,
   CardActions,
   CardBody,
   CardHeader,
+  Grid,
+  GridItem,
   PageSection,
   PageSectionVariants,
+  Switch,
   Text,
+  Title,
 } from '@patternfly/react-core'
 import { InfoCircleIcon } from '@patternfly/react-icons'
-import { Response } from 'jolokia.js'
+import { Request, Response } from 'jolokia.js'
 import React, { useContext, useEffect, useRef, useState } from 'react'
 import { HawtioEmptyCard } from '../HawtioEmptyCard'
 import { HawtioLoadingCard } from '../HawtioLoadingCard'
 import { attributeService } from '../attributes/attribute-service'
 import { WatchableAttributesForm } from './WatchableAttributesForm'
+import { TableComposable, Tbody, Td, Tr } from '@patternfly/react-table'
 
 type MBeanChartData = {
-  [name: string]: MBeanChartDataEntriesTime[]
+  [name: string]: { attributes: AttributeChartEntries }
 }
 
-type MBeanChartDataEntriesTime = {
-  name: string
-  time: number
-  data: AttributeValues
+type AttributeChartEntries = {
+  [name: string]: {
+    data: {
+      time: number
+      value: number
+    }[]
+    min: number
+    hasConstantValue: boolean
+  }
 }
+type AttributesEntry = { [attName: string]: { time: number; value: number } }
 
 export type AttributesToWatch = {
-  [name: string]: {
-    [attributeName: string]: boolean
+  [name: string]: { [attributeName: string]: boolean }
+}
+const AttributeChart = ({
+  name,
+  data,
+  min,
+}: {
+  name: string
+  data: { name: string; x: number; y: number }[]
+  min?: number
+}) => {
+  const containerRef = useRef<HTMLDivElement>(null)
+  const [width, setWidth] = useState<number>(0)
+  const handleResize = () => {
+    if (containerRef.current && containerRef.current.clientWidth) {
+      setWidth(containerRef.current.clientWidth)
+    }
   }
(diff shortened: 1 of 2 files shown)


packages/hawtio/src/plugins/shared/chart/Chart.tsx
```diff
@@ -20,9 +29,10 @@ import { HawtioEmptyCard } from '../HawtioEmptyCard'
 import { HawtioLoadingCard } from '../HawtioLoadingCard'
 import { attributeService } from '../attributes/attribute-service'
 import { WatchableAttributesForm } from './WatchableAttributesForm'
+import { TableComposable, Tbody, Td, Tr } from '@patternfly/react-table'
 
 type MBeanChartData = {
-  [name: string]: MBeanChartDataEntriesTime[]
+  [name: string]: { data?: MBeanChartDataEntriesTime[]; min: number }
```
tadayosi, line 35: why do we need to make `data` optional (`?`)?  we can possibly always initialise it with `[]`.

packages/hawtio/src/plugins/shared/chart/Chart.tsx
```diff
@@ -68,11 +78,16 @@ export const Chart: React.FunctionComponent = () => {
...
 
   function updateNode(mbeanObjectName: string, data: MBeanChartDataEntriesTime): void {
-    if (!chartData[mbeanObjectName]) chartData[mbeanObjectName] = []
+    if (!chartData[mbeanObjectName]) chartData[mbeanObjectName] = { data: [], min: Number.MAX_SAFE_INTEGER }
 
     // Don't add repeated responses to avoid duplicate points
-    if (!chartData[mbeanObjectName]?.find(value => value.time === data.time)) {
-      chartData[mbeanObjectName]?.push(data)
+    if (!chartData[mbeanObjectName]?.data?.find(value => value.time === data.time)) {
+      chartData[mbeanObjectName]?.data?.push(data)
```
tadayosi, line 85: does ts still require these `?`?  it seems redundant after we check their existence at the above `if`.

packages/hawtio/src/plugins/shared/chart/Chart.tsx
```diff
@@ -181,39 +195,60 @@ export const Chart: React.FunctionComponent = () => {
     )
   }
 
-  const AttributeChart = ({ name, data }: { name: string; data: { name: string; x: number; y: number }[] }) => (
-    <div style={{ height: '140px', display: 'flex', justifyContent: 'space-between', alignItems: 'center' }}>
-      <div style={{ width: '20%', overflow: 'hidden', fontSize: 10, paddingRight: 10, flexGrow: 1, flexShrink: 0 }}>
-        {name}
-      </div>
-      <div style={{ width: '80%', maxWidth: '80%', height: '140px', flexGrow: 0, flexShrink: 0 }}>
+  const AttributeChart = ({
```
tadayosi, line 198: this is now too big to have it inside a component definition. push it out to the top level of the same file. making it a independent component is also better for isolating component rendering as you know.

packages/hawtio/src/plugins/shared/chart/Chart.tsx
```diff
@@ -181,39 +195,60 @@ export const Chart: React.FunctionComponent = () => {
...
+    name,
+    data,
+    min,
+  }: {
+    name: string
+    data: { name: string; x: number; y: number }[]
```
tadayosi, line 204: why don't you just use `MBeanChartDataEntriesTime[]` here?

packages/hawtio/src/plugins/shared/chart/Chart.tsx
```diff
@@ -181,39 +195,60 @@ export const Chart: React.FunctionComponent = () => {
...
+    min: number
+  }) => {
+    const containerRef = useRef<HTMLDivElement>(null)
```
tadayosi, line 207: what is this for?

tadayosi (commented):
there are comments, but first let's start with fixing lint and build errors.

tadayosi:
> In addition I added a possibility to show only charts for attributes with values actually changing in time.

image

Now you are using a checkbox but it looks a bit primitive. You can instead use a switch: https://v4-archive.patternfly.org/v4/components/switch

packages/hawtio/src/plugins/shared/chart/Chart.tsx
mmelko: This is required for change the width after window serize afaik

mmelko:
> > In addition I added a possibility to show only charts for attributes with values actually changing in time.
> 
> image
> 
> Now you are using a checkbox but it looks a bit primitive. You can instead use a switch: https://v4-archive.patternfly.org/v4/components/switch

great suggestion! thanks

packages/hawtio/src/plugins/shared/chart/Chart.tsx
mmelko: hmm it looks I haven't commited everything. In my last iteration I don't have this. Same as  above.

packages/hawtio/src/plugins/shared/chart/Chart.tsx
```diff
@@ -20,89 +30,125 @@ import { HawtioEmptyCard } from '../HawtioEmptyCard'
 import { HawtioLoadingCard } from '../HawtioLoadingCard'
 import { attributeService } from '../attributes/attribute-service'
 import { WatchableAttributesForm } from './WatchableAttributesForm'
+import { TableComposable, Tbody, Td, Tr } from '@patternfly/react-table'
+
+// type MBeanChartData = {
+//   [name: string]: { data?: MBeanChartDataEntriesTime[] }
+// }
```
tadayosi, line 37: Why is this comment?

mmelko: that's going to be removed

mmelko:
@tadayosi I forgot to remove `WIP` yesterday. Can you check ? Performance is now better, although still not ideal due to processing batch of requests.

tadayosi (approved):

tadayosi:
The failing tests should be only related to Charts.

PR #847, opened 2024-04-01
refactor(app): migrate from craco to webpack for console assembly

### LINKED ISSUE #455: Consider stopping the use of CRA
Create React App (CRA) seems to be deprecated.
- https://github.com/facebook/create-react-app/issues/13072
- https://dev.to/ag2byte/create-react-app-is-officially-dead-h7o

It shouldn't impact our development soon, because we use it only at the last stage of building a Hawtio app and `@hawtio/react` is built with tsup. It only affects the `app` test package in this project, and downsteam it should affect the way we bundle the hawtio-next console in the hawtio standalone project, and custom plugins that are built with hawtio-next.

But let's consider switching from CRA to some other framework in the mid term. Options include:
- Eject CRA and use Webpack directly
- Migrate to other framework like Vite or Next.js

tadayosi:
Fix #455


tadayosi:
E2E are still very flaky, we don't want to merge it until it becomes more stable.

tadayosi:
Now it's become stable enough. Let me merge it.

PR #869, opened 2024-04-10
Migrate Contexts,Exchanges,Attributes to TableComposable 

### LINKED ISSUE #180: Upgrade instance of Patternfly Table to TableComposable
[Table](https://www.patternfly.org/v4/components/table/react-legacy) component is being replaced with [TableComposable](https://www.patternfly.org/v4/components/table). While some table instances have been upgraded, some, eg. InFlightExchanges, still use the legacy version. At some point, these remaining examples should be upgraded.

### LINKED ISSUE #845: Camel context selection dissapears after a while
After checking a CamelContexts the selection dissapears after a while as seen in the video: 
[Peek 2024-03-27 11-15.webm](https://github.com/hawtio/hawtio-next/assets/46345469/2033aa50-2f67-4470-b9a2-6f745cc9bad1)

mmelko:
I migrated tables to use TableComposable instead of old PF4 table.
Also fixed #845 by refactoring selected const to use list of strings instead Context objects. 
close #180

### DIFF
--- a/packages/hawtio/src/plugins/camel/contexts/Contexts.tsx
+++ b/packages/hawtio/src/plugins/camel/contexts/Contexts.tsx
@@ -4,7 +4,7 @@ import { HawtioLoadingCard } from '@hawtiosrc/plugins/shared'
 import { AttributeValues } from '@hawtiosrc/plugins/shared/jolokia-service'
 import { Card, CardBody, Text } from '@patternfly/react-core'
 import { InfoCircleIcon } from '@patternfly/react-icons'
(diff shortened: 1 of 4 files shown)


oscerd (approved):

tadayosi (approved):
It's ok for now, but ideally #845 should be fixed by splitting the component into smaller chunks so that the effect of re-rendering is narrowed to a specific component.

mmelko:
@tadayosi I don't think that was a problem actually. Contexts component is fairly simple. My impression was that problem was that for selectedState was used actually array of the `ContextState` objects. I assume it could be a problem of references in the array. Refactoring the selected state to actually to the array of the strings did the trick

PR #881, opened 2024-04-16
chore(Routes): Add the state label to the routes list

mmelko:
Screenshot 2024-04-16 at 16 22 44

### DIFF
--- a/packages/hawtio/src/plugins/camel/routes/CamelRoutes.tsx
+++ b/packages/hawtio/src/plugins/camel/routes/CamelRoutes.tsx
@@ -6,6 +6,7 @@ import {
   Dropdown,
   DropdownItem,
   KebabToggle,
+  Label,
   Modal,
   Toolbar,
   ToolbarContent,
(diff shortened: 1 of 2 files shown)


packages/hawtio/src/plugins/camel/routes/CamelRoutes.tsx
```diff
@@ -215,7 +220,9 @@ export const CamelRoutes: React.FunctionComponent = () => {
                   }}
                 />
                 <Td dataLabel='Name'>{route.routeId}</Td>
-                <Td dataLabel='State'>{route.state}</Td>
+                <Td dataLabel='State'>
+                  <RouteStateLabel state={route.state!} />
```
tadayosi, line 224: Why `state!`?  I believe it's a non-null assertion, but you don't seem to really assert the state is not null nor undefined. Instead, isn't it better to provide a default value with `route.state ?? <something>`?

tadayosi (approved):
lgmt, but one comment.

packages/hawtio/src/plugins/camel/routes/CamelRoutes.tsx
mmelko: you,re right. Or default unknown state in the Label component.

tadayosi: Or simply leave it empty?

tadayosi: e.g. `{route.state ? <RouteStateLabel state={route.state} /> : ''}`

mmelko: Well I'd like to have a gray unknown label instead  😇

mmelko:
I also refactored the test in the BlockedExchanges. Looks like there were a change of the behaviour after the upgrade of the react/testing-library in findBy...

PR #882, opened 2024-04-17
WIP: refactor(CamelContent): Refactor to use tabs instead of Navigation and React routes

mmelko:
`branch: jsolovjo:e2e-tests-fix-ux-changes-pr-882`

I used Tabs elements instead of the Tertiary navigation which is deprecated in the patternfly 5. Tabs are rendering only on click and are demounted from the DOM so the behaviour is the same except we don't have the items in the path as we had before. But for the camel it didn't work anyway. 

Using tabs instead of the navigation makes styles bit more tight: 
Screenshot 2024-04-17 at 21 45 00

Screenshot 2024-04-17 at 21 45 10

I have to sort out some issues, but the question is if we want to go this route.

### DIFF
--- a/packages/hawtio/src/plugins/camel/CamelContent.tsx
+++ b/packages/hawtio/src/plugins/camel/CamelContent.tsx
@@ -4,20 +4,19 @@ import {
   EmptyState,
   EmptyStateIcon,
   EmptyStateVariant,
-  Nav,
   NavItem,
-  NavList,
   PageGroup,
-  PageNavigation,
   PageSection,
   PageSectionVariants,
+  Tab,
+  Tabs,
   Text,
   Title,
 } from '@patternfly/react-core'
 import { CubesIcon } from '@patternfly/react-icons'
 import { Response } from 'jolokia.js'
 import React, { useContext, useEffect, useState } from 'react'
-import { NavLink, Navigate, Route, Routes, useLocation, useNavigate } from 'react-router-dom'
+import { useNavigate } from 'react-router-dom'
 import './CamelContent.css'
 import * as camelService from './camel-service'
 import { CamelContext } from './context'
@@ -30,7 +29,7 @@ import { BrowseMessages } from './endpoints/BrowseMessages'
 import { EndpointStats } from './endpoints/EndpointsStats'
 import { SendMessage } from './endpoints/SendMessage'
 import { Exchanges } from './exchanges'
-import { log, pluginPath } from './globals'
+import { log } from './globals'
 import { Profile } from './profile/Profile'
 import { Properties } from './properties'
 import { RestServices } from './rest-services/RestServices'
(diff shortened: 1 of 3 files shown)


packages/hawtio/src/plugins/camel/CamelContent.tsx
```diff
@@ -47,11 +46,90 @@ type NavItem = {
   component: JSX.Element
   isApplicable: (node: MBeanNode) => boolean
 }
+/*
+ * Test if nav should contain general mbean tabs
+ */
```
tadayosi, line 51: Top level should use jsdoc format.

packages/hawtio/src/plugins/camel/CamelContent.tsx
```diff
@@ -47,11 +46,90 @@ type NavItem = {
...
+  return (
+    camelService.hasMBean(node) &&
+    !camelService.isContextsFolder(node) &&
+    !camelService.isRoutesFolder(node) &&
+    !camelService.isRouteXmlNode(node)
+  )
+}
 
+// The order of the items in the following list is the order in will the tabs will be visualized.
+// For more info check: https://github.com/hawtio/hawtio-next/issues/237
```
tadayosi, line 62: top-level should use jsdoc format.

packages/hawtio/src/plugins/camel/CamelContent.tsx
```diff
@@ -47,11 +46,90 @@ type NavItem = {
   component: JSX.Element
   isApplicable: (node: MBeanNode) => boolean
 }
+/*
+ * Test if nav should contain general mbean tabs
+ */
+const isDefaultApplicable = (node: MBeanNode) => {
```
tadayosi, line 52: this should be function instead of function object.

packages/hawtio/src/plugins/camel/CamelContent.tsx
```diff
@@ -47,11 +46,90 @@ type NavItem = {
...
+    title: 'Type Converters',
+    component: <TypeConverters />,
+    isApplicable: camelService.hasTypeConverter,
+  },
+  { id: 'chart', title: 'Chart', component: <Chart />, isApplicable: isDefaultApplicable },
+  // Applicable for same criteria as trace
+  { id: 'profile', title: 'Profile', component: <Profile />, isApplicable: camelService.canTrace },
+  { id: 'trace', title: 'Trace', component: <Trace />, isApplicable: camelService.canTrace },
+  { id: 'debug', title: 'Debug', component: <Debug />, isApplicable: camelService.canGetBreakpoints },
+]
```
tadayosi, line 118: as const

packages/hawtio/src/plugins/camel/CamelContent.tsx
```diff
@@ -66,122 +144,42 @@ export const CamelContent: React.FunctionComponent = () => {
...
-        {navItems.length > 1 && <PageNavigation>{camelNav}</PageNavigation>}
-      </PageGroup>
-
-      <PageSection id='camel-content-main'>
-        {navItems.length > 0 && (
-          <Routes>
-            {camelNavRoutes}
-            <Route key='root' path='/' element={<Navigate to={navItems[0]?.id ?? ''} />} />
-          </Routes>
+    <PageGroup style={{ height: '100%' }}>
```
tadayosi, line 162: why should we keep using PageGroup?

packages/hawtio/src/plugins/camel/CamelContent.tsx
```diff
@@ -66,122 +144,42 @@ export const CamelContent: React.FunctionComponent = () => {
...
-            <NavLink to={{ pathname: nav.id, search }}>{nav.title}</NavLink>
-          </NavItem>
-        ))}
-      </NavList>
-    </Nav>
+  const camelTabs = (
+    <Tabs mountOnEnter unmountOnExit activeKey={activeTab} onSelect={(_, eventKey) => setActiveTab(eventKey as string)}>
+      {tabItems.map(nav => (
+        <Tab eventKey={nav.id} key={nav.id} title={nav.title}>
+          {camelService.canViewRouteDiagram(selectedNode) ? (
```
tadayosi, line 151: why do we put this here? it makes the code a bit more cluttered.

packages/hawtio/src/plugins/camel/CamelContent.tsx
```diff
@@ -47,11 +46,90 @@ type NavItem = {
...
   const routeDiagramContext = useRouteDiagramContext()
-  const { pathname, search } = useLocation()
+  const [activeTab, setActiveTab] = useState<string>('')
+  const [tabItems, setTabItems] = useState<NavItem[]>([])
+
+  useEffect(() => {
+    if (selectedNode) {
+      /* Filter the nav items to those applicable to the selected node */
+      const tabItems = allNavItems.filter(nav => nav.isApplicable(selectedNode))
+      setActiveTab(tabItems[0]?.id ?? '')
```
tadayosi, line 129: it appears always the first tab is selected. shouldn't we use pathname?

tadayosi (commented):
first please fix ci errors

packages/hawtio/src/plugins/camel/CamelContent.tsx
mmelko: Yes you're right, but this was the always the case actually with the camel plugin. I will fix that though.

packages/hawtio/src/plugins/camel/CamelContent.tsx
mmelko: for passing the Context. I moved the NavItems outside of the component.

mmelko:
Closing for now,. I will do the style changes in the different PR without migration to tabs component.

PR #894, opened 2024-04-23
fix(connect): HAWNG-487 apply authentication throttling to connect login

tadayosi:
https://issues.redhat.com/browse/HAWNG-487

### DIFF
--- a/packages/backend-middleware/src/proxy.ts
+++ b/packages/backend-middleware/src/proxy.ts
@@ -27,6 +27,7 @@ export async function proxy(uri: string, req: Request, res: Response) {
       switch (res2.status) {
         case 401:
         case 403:
+        case 429:
           log.info('Authentication failed on remote server:', error.status, error.message, uri)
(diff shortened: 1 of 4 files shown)


oscerd (approved):

grgrzybek:
Looks good - there are no changes required at hawtio/hawtio side, right?

tadayosi:
@grgrzybek Thanks.

> Looks good - there are no changes required at hawtio/hawtio side, right?

No, no changes required. It's based on the existing capability in hawtio/hawtio.

However, just FYI, if we would look into HAWNG-487 more seriously, ideally there can be a few more enhancements from hawtio/hawtio and jolokia sides, including:
- Enable throttling on Jolokia endpoint in a Quarkus application with Quarkus Security: https://github.com/hawtio/hawtio/tree/4.x/examples/quarkus
- Implement the same throttling mechanism even at Jolokia itself in JVM agent mode

Anyway, we should work on them after product GA if we do.

PR #898, opened 2024-04-24
refactor(camel): lazy load camel models

### LINKED ISSUE #754: Consider lazy loading of Camel models

tadayosi:
Fix #754


tadayosi:
This should supercede #801 by @mmuzikar

mmuzikar (approved):
looking great!

PR #920, opened 2024-05-03
fix(Source): Use dynamic height instead of fixed height in pixels

mmelko:


tadayosi (approved):
Next time, please provide before/after screenshots so that reviewers can better assess whether it's the right way to fix the issue.

PR #930, opened 2024-05-08
HAWNG-646: Default connection modal to https scheme

phantomjinx:
* Makes the switch on the ConnectionModal default to https

### DIFF
--- a/packages/hawtio/src/plugins/shared/connect-service.ts
+++ b/packages/hawtio/src/plugins/shared/connect-service.ts
@@ -26,7 +26,7 @@ export type Connection = {
 
 export const INITIAL_CONNECTION: Connection = {
   name: '',
-  scheme: 'http',
+  scheme: 'https',
   host: 'localhost',
   port: 8080,
   path: '/hawtio/jolokia',


phantomjinx:
`https` selected by default:

Screenshot_20240508_133246

packages/hawtio/src/plugins/connect/remote/Remote.tsx
```diff
@@ -61,6 +61,8 @@ const RemoteToolbar: React.FunctionComponent = () => {
   }
 
   const initialConnection = { ...INITIAL_CONNECTION }
+  // default connection modal to 'https'
+  initialConnection.scheme = 'https'
```
tadayosi, line 65: why don't we change the INITIAL_CONNECTION directly instead?

phantomjinx: Did so initially but used elsewhere so was being cautious. Will change accordingly.

phantomjinx:
@jsolovjo 
Looks like this broke the e2e tests?

jsolovjo:
> @jsolovjo Looks like this broke the e2e tests?

seems yes
as the tests use simple http connection
I can fix it

tadayosi (approved):

Issue #705, opened 2023-12-07
How to stop login from retrying once connection is lost


tadayosi:
It's good idea to provide an option at the Connect preferences that disables retrying connections on connection lost.

### Discussed in https://github.com/orgs/hawtio/discussions/3123

<div type='discussions-op-text'>

<sup>Originally posted by **ilxyz** December  7, 2023</sup>
Hello,

I noticed once a session expires and you are already logged into jolokia/remote jmx inside hawtio, a popup appears with the message "Connection lost. Retrying...", which keeps attempting login to hawtio with the jolokia/remote jmx credentials. This is generating a login error message per minute, which I am trying to stop from happening.

Is there a way to NOT have hawtio keep retrying the connection every minute and completely kill the session (in all windows)? Or a way to set a default refresh rate across the board in hawtio?</div>

----
related:
* #635
* hawtio/hawtio#3178

grgrzybek:
With improvements from #906 I'll simply disable the toast. Failed connection will be indicated by red "disconnected" icon in the top bar.

grgrzybek:
PR: #936
I will continue the fix for credential mismatch here: hawtio/hawtio#3178

PR #936, opened 2024-05-09
feat(connect): Display "connection lost" toast only once. Fixes (#705)

grgrzybek:


packages/hawtio/src/plugins/shared/jolokia-service.ts
```diff
@@ -309,8 +309,7 @@ class JolokiaService implements IJolokiaService {
   }
 
   private ajaxError(resolve?: AjaxErrorResolver): JQueryAjaxError {
-    const errorThreshold = 2
```
tadayosi, line 312: I still think the threshold is needed. It's used to ignore first 2 failures for some unknown reasons to make the connection robust. If we removed this, I wonder if it would then show error toast more frequently.

tadayosi (commented):
I think it's ok to make toast displayed only once, but I still think the threshold is needed. Please check it again.

packages/hawtio/src/plugins/shared/jolokia-service.ts
grgrzybek: ah sure - I'll display once, keep the status in the header (which was my main change), but display the toast after the threshold.

grgrzybek: @tadayosi how about now?

tadayosi (approved):
lgtm now!

PR #960, opened 2024-05-30
refactor(Dropdowns):Refactor the components across the hawtio to use the PF5 dropdown instead of old deprecate ones

mmelko:
This PR gets rid of deprecated Dropdown components from the PF4 which is now at `'@patternfly/react-core/deprecated'` with the recent ones. 
`branch: jsolovjo:e2e-tests-fix-pf5-dropdowns`

### DIFF
--- a/packages/hawtio/src/plugins/connect/remote/Remote.tsx
+++ b/packages/hawtio/src/plugins/connect/remote/Remote.tsx
@@ -13,14 +13,18 @@ import {
   DataListItem,
   DataListItemCells,
   DataListItemRow,
+  Dropdown,
+  DropdownItem,
+  DropdownList,
   Icon,
+  MenuToggle,
+  MenuToggleElement,
   Modal,
   ModalVariant,
   Toolbar,
   ToolbarContent,
   ToolbarItem,
 } from '@patternfly/react-core'
-import { Dropdown, DropdownItem, DropdownPosition, KebabToggle } from '@patternfly/react-core/deprecated'
 import { PluggedIcon, PlusIcon, UnpluggedIcon } from '@patternfly/react-icons'
 import React, { useContext, useEffect, useState } from 'react'
 import { DELETE } from '../connections'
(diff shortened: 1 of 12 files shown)


packages/hawtio/src/plugins/connect/remote/Remote.tsx
```diff
@@ -218,19 +227,23 @@ const ConnectionItem: React.FunctionComponent<{
           </Button>
           <Dropdown
             key={`connection-action-dropdown-${id}`}
-            isPlain
```
tadayosi, line 221: sometimes it's left, sometimes it's removed on Dropdown. is there any difference?

tadayosi (approved):

packages/hawtio/src/plugins/connect/remote/Remote.tsx
mmelko: yeah, when is plain it's missing borders in the menu. I removed that from all dropdowns now because it looked better in my opinion.

tadayosi: Can you show us the screenshots?

tadayosi: Maybe changing the styles can be done under a separate issue from this refactoring to PF 5, as it's doing two different things at once if it's not just mandatory for PF 5.

mmelko: It's actually not about changing the style but rather preserving the current one. It seems that in PF5 `isPlain` changed the behaviour and it draws the whole dropdown without the border, not only the 'header': dropdown with keeping `isPlain`(on the left PF5 with isPlain): 
Screenshot 2024-06-06 at 15 43 42

dropdown without `isPlain`:
Screenshot 2024-06-06 at 15 40 57
so as you can see the last picture removing `isPlain` is completely fine as is preserves the current look and feel and that's why I removed them.

Only change is in the header where there is a thin side border:
Screenshot 2024-06-06 at 15 50 14

But it follows the PF5 quickstarts.

tadayosi: Makes sense. Let's remove all the `isPlain` here then.

tadayosi:
Please look at hawtio/hawtio#3514 first, as unless we fix Juri's E2E at hawtio/hawtio we won't be able to go anywhere further.

PR #970, opened 2024-06-07
feat(shared,jmx): support serialising long to string in JSON response from Jolokia

### LINKED ISSUE #292: UI numeric type (long type) issue
Make sure this issue is fixed with hawtio-next: https://github.com/hawtio/hawtio/issues/2746

tadayosi:
* Upgrade jolokia.js to 2.0.3 (that supports `serializeLong` option)
* Create JMX preferences and provide `Serialize long to string` option, which can affect Jolokia requests within JMX plugin

image

This fix limits the `serializeLong=string` option to JMX, because the change of the bahaviour might cause unexpected errors in other plugins. JMX is a generic MBeans manipulation plugin, so it should make sense to have such an option plugin-wide.

Each plugin should decide whether it's worth applying the option or not.

Fix #292

### DIFF
--- a/packages/hawtio/package.json
+++ b/packages/hawtio/package.json
@@ -54,7 +54,7 @@
     "@types/react-router-dom": "^5.3.3",
     "dagre": "^0.8.5",
     "eventemitter3": "^5.0.1",
-    "jolokia.js": "^2.0.1",
+    "jolokia.js": "^2.0.3",
     "jquery": "^3.7.1",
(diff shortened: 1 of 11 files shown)


tadayosi:
A bit more screenshots to demonstrate how it fixes the long values issue:

#### Without `Serialize long to string` (default)
image

#### With `Serialize long to string`
image

image

grgrzybek (approved):

Issue #971, opened 2024-06-07
Warnings in browser console after upgrading to PF 5


tadayosi:
Using `yarn start` and connecting to a Quarkus app with `mvn quarkus:dev -pl examples/quarkus` in hawtio/hawtio:

### Camel

![image](https://github.com/hawtio/hawtio-next/assets/156692/55b68437-43b1-485c-aaac-b1036df04372)

![image](https://github.com/hawtio/hawtio-next/assets/156692/e3ecd948-0a5d-4e7f-bb91-89df7fa3d56d)

### Runtime

![image](https://github.com/hawtio/hawtio-next/assets/156692/ded6cd99-21a2-48e9-b18e-b40724e0f584)

mmelko:
That `ChartVoronoiContainer` error comes from the charts dependency. It was there before and it has to be fixed there. Anyway it's not showing after build. 

With others PR is on the way.

PR #982, opened 2024-06-12
fix: add missing aria-labels 

mmelko:
This PR adds missing aria-labels that caused warning in the console. Also remove defaultProps for the CamelIcon and replace it with JS defaults.

fix: #971


tadayosi (approved):

tadayosi:
@mmelko Have you also tested that it fixes #971 at hawtio/hawtio?  Like the console in a Quarkus or Spring Boot application.

mmelko:
Yes, I double-checked that.

PR #984, opened 2024-06-14
fix(Runtime-metrics): fix incorrect memory conversions

mmelko:
fixes HAWNG-709
Screenshot 2024-06-14 at 16 17 36
Screenshot 2024-06-14 at 16 32 17


tadayosi (approved):
We can further extract the unit resolution logic to the service's method, but let's merge it for now.

PR #993, opened 2024-06-24
feat(camel-source): Make Source editor editable when it's enabled in the CamelContext

mmelko:
When updating the routes is enabled then the editor in the Source tab under particular route becomes writable additional control/check is shown: 
Screenshot 2024-06-24 at 15 54 45
When the code is changed:
Screenshot 2024-06-24 at 15 54 26
When it's saved the notification is fired and control button is changed back: 
Screenshot 2024-06-24 at 15 54 33

### DIFF
--- a/packages/hawtio/src/plugins/camel/routes/Source.tsx
+++ b/packages/hawtio/src/plugins/camel/routes/Source.tsx
@@ -1,24 +1,81 @@
 import { CamelContext } from '@hawtiosrc/plugins/camel/context'
-import { CodeEditor, Language } from '@patternfly/react-code-editor'
+import { CodeEditor, CodeEditorControl, Language } from '@patternfly/react-code-editor'
 import React, { useContext, useEffect, useState } from 'react'
 import { log } from '../globals'
+import { CheckCircleIcon, SaveIcon } from '@patternfly/react-icons'
+import { isRouteNode, isRoutesFolder } from '@hawtiosrc/plugins/camel/camel-service'
+import { routesService } from './routes-service'
+import { eventService } from '@hawtiosrc/core'
 
 export const Source: React.FunctionComponent = () => {
   const { selectedNode } = useContext(CamelContext)
   const [xmlSource, setXmlSource] = useState('')
+  const [isUpdateEnabled, setIsUpdateEnabled] = useState(false)
+  const [codeChanged, setCodeChanged] = useState(false)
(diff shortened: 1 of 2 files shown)


tadayosi:
@mmelko Which issue are you supposed to fix with this pull req?  Please link it if there is.

packages/hawtio/src/plugins/camel/routes/Source.tsx
```diff
@@ -1,24 +1,76 @@
...
-    const xml = selectedNode?.getMetadata('xml')
-    if (xml) {
-      setXmlSource(xml)
-    } else {
-      log.warn('Source - Unable to fetch XML from', selectedNode)
+    if (selectedNode) {
+      const xml = selectedNode?.getMetadata('xml')
+
+      if (isRoute) {
+        routesService.isRouteUpdateEnabled(selectedNode!).then(enabled => {
```
tadayosi, line 22: why is `!` needed?  I don't think it's needed.

packages/hawtio/src/plugins/camel/routes/Source.tsx
```diff
@@ -1,24 +1,76 @@
...
+  }, [isRoute, selectedNode])
+
+  const onCodeChange = (code: string) => {
+    setCodeChanged(code !== xmlSource)
+  }
+
+  const onSaveClick = (code: string) => {
+    if (isRoute && code !== xmlSource) {
+      try {
+        routesService.saveRoute(selectedNode!, code)
```
tadayosi, line 42: same here: `!`

If we need `selectedNode` to exist, check it's existence right after the `useEffect` and then you are free from null-checking of it afterwards. See how the other components handle that.

tadayosi (commented):
A few comments.

Also, have you checked that updating the route works when the context has multiple routes?

mmelko:
> A few comments.
> 
> Also, have you checked that updating the route works when the context has multiple routes?

Well, the quickstart I tested has 2 routes. Also this operation is available not on context level but on route level.

As I mentioned, you won't be able to edit the source when you open it in the Routes view, you have to open source for specified route. Sorry I cut the screenshots too much so it's not obvious from there.

packages/hawtio/src/plugins/camel/routes/Source.tsx
mmelko: you're right.

tadayosi (approved):

PR #1012, opened 2024-07-08
feat(Source): Add warning about the route being editable

mmelko:
Screenshot 2024-07-08 at 13 50 03

Also the save button behaviour is changed. When there is no change in the source, the button is disabled.


tadayosi (approved):
thx!

PR #1022, opened 2024-07-12
feat(RouteDiagram): Proper visualisation of parallel processing of the multicast

mmelko:
Also I added tests for the multicast rendering and did small variables refactor. 

Route visualisation without parallelProcessing=true:
Screenshot 2024-07-12 at 09 46 17

with parallelProcessing=true:
Screenshot 2024-07-12 at 09 45 50


tadayosi (approved):

tadayosi:
@mmelko Don't forget to point to the issue that the pull req is supposed to fix from here.

PR #1063, opened 2024-08-15
Refactor paginated and filetered tables into one component

### LINKED ISSUE #1004: Runtime Thread view switches to page 1 every few seconds
It looks like when the thread data are refreshed the page counter resets back to 1:
[Peek 2024-06-25 10-34.webm](https://github.com/hawtio/hawtio-next/assets/46345469/b81f1663-c522-4deb-8ad0-07a9ebc35861)

### LINKED ISSUE #1005: Small enhancements for the UI
I've found several small issues with the UI that could be fixed
Runtime plugin:
- [ ] The properties with long values should be truncated by default and shown the whole value only on user request
- [ ] The JVM info is formatted like `Header<space>:<value>`

Help, preferences links in the header
- [ ] The first link is underscored on hover, but not the rest

mmuzikar:
Created a FilteredTable component that can be used for generic rendering of tables with filters, sorting and search. Fixes the pagination issues where the page would always reset and is keeping the same features so far

Fixes #1004 
Fixes #1005


tadayosi (approved):
great!

tadayosi:
@mmelko Could you review this as well, as this touches the parts you developed?

mmelko:
@mmuzikar I noticed that the Logs table is quite changed from the previous. There is a one small issue, when you click on the Clear All Filters there .. the Log level isn't cleared.

mmuzikar:
@mmelko fixed

mmelko (approved):

PR #1087, opened 2024-08-29
fix: Disable help for disabled plugins

mmuzikar:
Dev build connected to quarkus looks like this:
image
with `"disabledRoutes": ["/disabled", "/camel", "/quartz"]`:
image


tadayosi (approved):
awesome

PR #1111, opened 2024-09-10
Jolokia 2.1

grgrzybek:


mmuzikar:
@grgrzybek 
I will take a look at it, from the reports there's a ton of the same error: `JavaScript error: http://localhost:3000/hawtio/static/js/vendors-node_modules_module-federation_utilities_dist_index_cjs_js-node_modules_patternfly_re-2ef800.chunk.js line 16585 > eval, line 26: TypeError: Jolokia is undefined
`

grgrzybek:
@mmuzikar I only saw that after trying to run `yarn start`. Thanks for your time, but it's definitely my fault with this line: https://github.com/jolokia/jolokia/blob/bf354b1c80fa5decb8c46c78117243273725a144/client/javascript-esm/packages/jolokia-simple/package.json#L23 which I changed from:
```
"default": "./src/jolokia-simple.ts"
```

to:
```
"default": "./dist/jolokia-simple.js"
```

grgrzybek:
darn... ES modules vs. node `require()` again...

packages/hawtio/src/plugins/camel/CamelContent.tsx
```diff
@@ -209,14 +209,21 @@ const CamelContentContextToolbar: React.FunctionComponent = () => {
...
-
-        // Replace the context in the existing set with the new one
-        const attrs = response.value as AttributeValues
-        const newAttrs = contextsService.toContextState(selectedNode, attrs)
-        if (newAttrs) setContextState(newAttrs)
-      })
+      contextsService.register(
+        { type: 'read', mbean: objectName },
+        (response: JolokiaSuccessResponse | JolokiaErrorResponse) => {
+          if (Jolokia.isError(response)) {
```
tadayosi, line 215: Not related to this pull req, but ideally `Jolokia.isError` should then narrow down the type of response to `JolokiaErrorResponse` with `response is JolokiaErrorResponse` at return so that you don't need to cast the type again inside the if statement.
https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

tadayosi: Since this type guard is used in many places, we can declare some custom `isJolokiaError` type guard function in `util/jolokia.ts`, and we'll no longer need this type casting for both error and success types.

We may also declare a type alias to compact the argument definitions:
```
type JolokiaResponse = JolokiaSuccessResponse | JolokiaErrorResponse
```

packages/hawtio/src/plugins/shared/jolokia-service.test.ts
```diff
@@ -23,12 +24,14 @@ describe('JolokiaService', () => {
...
+    Jolokia.prototype.list = jest.fn((path?: string | string[] | RequestOptions, opts?: SimpleRequestOptions) => {
       if (typeof path !== 'string') throw new Error('String is expected for path')
       if (!opts) throw new Error('No options set')
-      if (!opts.success) throw new Error('No success option set')
+      if (!opts.success) {
+        throw new Error('No success option set')
+      }
 
-      opts.success({
+      ;(opts.success! as SimpleResponseCallback)({
```
tadayosi, line 34: this `;` at the beginning doesn't look cool. Isn't it something we can remove?

packages/hawtio/src/plugins/shared/jolokia-service.test.ts
```diff
@@ -44,20 +47,37 @@ describe('JolokiaService', () => {
...
-      opts.error({
-        status: -1,
-        timestamp: 123456789,
-        request: { type: 'list', path: path },
-        value: null,
-        error_type: 'non-exist',
-        error: 'ERROR HAS OCCURRED',
-        stacktrace: 'not available',
-      })
+      ;(opts.error! as ErrorCallback)(
```
tadayosi, line 70: same here

packages/hawtio/src/util/jolokia.ts
```diff
@@ -91,14 +126,14 @@ export function onGenericSuccessAndError<S, E, O extends BaseRequestOptions>(suc
...
-  return (response: ErrorResponse) => {
+function defaultErrorHandler(): ErrorCallback {
+  return (response: JolokiaErrorResponse, _index: number) => {
     if (!response.error) {
       return
     }
 
     const req = response.request
-    const operation = req?.type === 'exec' ? req.operation : 'unknown'
+    const operation = req?.type === 'exec' ? 'exec/' + req.operation : req ? req.type : 'unknown'
```
tadayosi, line 136: is it correct?

tadayosi (approved):

tadayosi:
@grgrzybek Btw, can you please follow the commit message convention?  Otherwise, the commit-and-tag-version cannot pick up the changes correctly for automated releases.

grgrzybek:
> Btw, can you please follow the commit message convention

ah, sorry - old habits...

packages/hawtio/src/plugins/camel/CamelContent.tsx
grgrzybek: I really tried the type guard (`response is ...`), but somehow it didn't work... In my simple scenario I understood how it works and tells TS about the types in `if` and `else` branches.
But somehow it didn't work with Jolokia types.

But now I checked in this particular file and it worked... ;)

packages/hawtio/src/plugins/shared/jolokia-service.test.ts
grgrzybek: This was added by `yarn format:fix` somehow... Let me check

packages/hawtio/src/util/jolokia.ts
grgrzybek: It's only for logging - I wanted to have actual type when type is not `exec`.

packages/hawtio/src/plugins/shared/jolokia-service.test.ts
grgrzybek: checked again - I removed these semicolons and I got:
```
09:32 $ yarn format:check
Checking formatting...
[warn] packages/hawtio/src/plugins/shared/jolokia-service.test.ts
[warn] packages/hawtio/src/plugins/shared/jolokia-service.ts
[warn] packages/hawtio/src/ui/util/ExpandableText.tsx
[warn] Code style issues found in 3 files. Run Prettier with --write to fix.
```

`format:fix` got the semicolons back...

tadayosi: maybe removing the type cast can remove the necessity of `;`

packages/hawtio/src/plugins/shared/jolokia-service.test.ts
grgrzybek: I'm getting:
```
TS2349: This expression is not callable.
  Not all constituents of type "ignore" | ErrorCallback are callable.
    Type "ignore" has no call signatures.
```
without casting

packages/hawtio/src/plugins/camel/CamelContent.tsx
grgrzybek: I made `Jolokia.isError` an actual type predicate here: https://github.com/jolokia/jolokia/commit/29eeeafab8e7ddf838c7055afcbd84ef509245a5

packages/hawtio/src/plugins/shared/jolokia-service.test.ts
grgrzybek: I'm getting:
```
TS2349: This expression is not callable.
  Not all constituents of type "ignore" | ErrorCallback are callable.
    Type "ignore" has no call signatures.
```
without casting

packages/hawtio/src/plugins/shared/jolokia-service.test.ts
grgrzybek: this was also added here:
https://github.com/hawtio/hawtio-next/blob/9125bb287c0bbc4a8e54f2dd720bcd6750ef341b/packages/hawtio/src/ui/util/ExpandableText.tsx#L297-L298

packages/hawtio/src/plugins/shared/jolokia-service.test.ts
tadayosi: ok, I can check that later

packages/hawtio/src/plugins/shared/jolokia-service.test.ts
tadayosi: ok, I can check that later

tadayosi (approved):
LGTM.

Just curious about what was the culprit of the last e2e issue?

tadayosi:
Great job!

grgrzybek:
> LGTM.
> 
> Just curious about what was the culprit of the last e2e issue?

it was silly mistake related to switch from `$.ajax` to `fetch()` in Jolokia.

We had this:
https://github.com/hawtio/hawtio-next/blob/c6fe03c4311dbed5914e4bd26f9d042a148e547d/packages/hawtio/src/plugins/shared/jolokia-service.ts#L268

and I simply removed it. The fix was simply to actually set up the credentials to be passed directly to `Jolokia` constructor:
https://github.com/hawtio/hawtio-next/blob/dab6b29b13f668c92115b9079cac0d24d6345aae/packages/hawtio/src/plugins/shared/jolokia-service.ts#L322

PR #1114, opened 2024-09-12
fix(shared): Stop using jquery (for ajax and selection) (closes #76)

### LINKED ISSUE #76: Stop using jQuery/ajax and use Fetch API instead
Relates #8

grgrzybek:


grgrzybek:
This one is easier than Jolokia 2.1.x migration.

oscerd (approved):

grgrzybek:
> Maybe we should then check whether the changes don't break integration with Keycloak and OIDC.
I have hawtio/hawtio#3404 for that.

Actually one of the big arguments to move to fetch() API (and expose headers configuration for Jolokia) was to handle XSRF tokens - we didn't have this before. And this caused problems with Spring Security + OIDC.

I'll merge first this one and then continue the work.

Issue #1125, opened 2024-09-19
Implement session management in hawtio OIDC


grgrzybek:
Without this, any hard refresh (CTRL-R) leads to full OAuth2 login redirect cycle...

BTW, when checking non-generic Keycloak implementation again with containerized Keycloak (as in Hawtio docs) 25.0.2, I found problems with current hawtio version:
* keycloak/keycloak#27624 - not having this fixed leads to redirect loop. Works with Keycloak 23.
* keycloak/keycloak#26651 - already fixed in jolokia.js 24+, but we still use 23. Update fails because of some `crypto` requirements...

grgrzybek:
cc: @tadayosi

grgrzybek:
The fix won't involve handling `check_session_iframe` URL from OIDC discovery endpoint. Simply because I don't see (where I can see) support for it in Entra ID.
However the refresh problem (enforcing reauthentication) was caused by explicit `prompt=login` URL query param sent to authorization endpoint.

PR #1126, opened 2024-09-20
fix(#1125): Fix scenario with OIDC and remote, authenticated Jolokia …

grgrzybek:
…agent

### DIFF
--- a/packages/hawtio/src/plugins/auth/keycloak/keycloak-service.ts
+++ b/packages/hawtio/src/plugins/auth/keycloak/keycloak-service.ts
@@ -117,6 +117,11 @@ class KeycloakService implements IKeycloakService {
     const initOptions: KeycloakInitOptions = {
       onLoad: 'login-required',
       pkceMethod,
+      // required for Keycloak 23
+      // see: https://github.com/keycloak/keycloak/issues/26651
+      // and we can't switch to Keycloak 25
+      // see: https://github.com/keycloak/keycloak/issues/27624
+      useNonce: false,
     }
     return initOptions
   }
--- a/packages/hawtio/src/plugins/auth/oidc/oidc-service.ts
+++ b/packages/hawtio/src/plugins/auth/oidc/oidc-service.ts
@@ -1,5 +1,5 @@
 import { ResolveUser, userService } from '@hawtiosrc/auth/user-service'
-import { Logger } from '@hawtiosrc/core'
+import { hawtio, Logger } from '@hawtiosrc/core'
 import { jwtDecode } from 'jwt-decode'
 import * as oidc from 'oauth4webapi'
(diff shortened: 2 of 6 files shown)


packages/hawtio/src/plugins/auth/oidc/oidc-service.ts
```diff
@@ -195,9 +214,11 @@ export class OidcService implements IOidcService {
...
       // authorizationUrl.searchParams.set('login_hint', 'hawtio-viewer@fuseqe.onmicrosoft.com')
       // authorizationUrl.searchParams.set('hsu', '1')
-      if (config.prompt) {
-        authorizationUrl.searchParams.set('prompt', config.prompt)
-      }
+      // it's better to stick to default, as 'prompt=login' always
+      // shows login screen...
+      // if (config.prompt) {
+      //   authorizationUrl.searchParams.set('prompt', config.prompt)
+      // }
```
tadayosi, line 221: too many comment outs here and there. maybe we can clean them up now.

tadayosi (approved):

packages/hawtio/src/plugins/auth/keycloak/keycloak-service.ts
```diff
@@ -117,6 +117,11 @@ class KeycloakService implements IKeycloakService {
     const initOptions: KeycloakInitOptions = {
       onLoad: 'login-required',
       pkceMethod,
+      // required for Keycloak 23
+      // see: https://github.com/keycloak/keycloak/issues/26651
+      // and we can't switch to Keycloak 25
+      // see: https://github.com/keycloak/keycloak/issues/27624
+      useNonce: false,
```
tadayosi, line 124: so should we pin the version of keycloak.js to 23.x?  related: #1133

grgrzybek: I'm not sure what to do yet (will get back to it later), so I only left the dev notes - I hope you don't mind ;)

packages/hawtio/src/plugins/auth/oidc/oidc-service.ts
grgrzybek: I've removed the comment ;) Thanks for spotting!

packages/hawtio/src/plugins/auth/keycloak/keycloak-service.ts
grgrzybek: @tadayosi are you ok with this explanation? :)

tadayosi: It's not a change request but a pure question :-)  I wonder what we should do for #1133. Feel free to resolve it.

Issue #870, opened 2024-04-15
Migrate to ESLint v9


tadayosi:
ESLint v9 has breaking changes #861 which require the config file to be transformed to the new flat config format `eslint.config.js`.
https://eslint.org/docs/latest/use/migrate-to-9.0.0

PR #1148, opened 2024-10-07
fix: Upgrades eslint to v9 and migrate to flat config #870

phantomjinx:
* Conversion of .eslintrc.js to eslint.config.mjs

* Incorporation of .eslintignore into config file

* Upgrades eslint package and respective plugins

* eslint.config.mjs
 * Some plugins are still not eslint 9 compatible so the compat library is used as a go-between

* [*.ts, *.js]
 * Updates to satisfy new warnings produced by running `yarn lint`


phantomjinx:
Some of these resolutions to warnings, eg. eslint-disable comments removed, seem a little unsettling but they 'seem' to be the right course of action. Anyway, please review and we can take further iterations to get this right.

Note: eslint plugins such as testing-library and react-hooks are not fully compliant with eslint v9.

tadayosi (approved):
Awesome, thank you!

PR #1169, opened 2024-10-16
fix: Adds better error handling of workspace loading errors

phantomjinx:
* Adds a status plugin for displaying errors messages if the workspace failed to load. Such errors can get lost in the console log and a blank page is the result in the client.

* workspace.ts
  * If an error occurs then add them to a collection in the workspace
  * Provides an api for quizzing the workspace on whether it has errored

* app-status
  * Plugin that is only active if
    * there is a current connection (no point in displaying if nothing has been connected to yet)
    * there are no mbeans collected from the jolokia connection
    * Errors were flagged from the workspace
  * Assuming the plugin is active, offer a component that provides notification of the workspace *

### DIFF
--- a/packages/hawtio/src/plugins/shared/workspace.ts
+++ b/packages/hawtio/src/plugins/shared/workspace.ts
@@ -12,6 +12,8 @@ const HAWTIO_REGISTRY_MBEAN = 'hawtio:type=Registry'
 const HAWTIO_TREE_WATCHER_MBEAN = 'hawtio:type=TreeWatcher'
 
 export interface IWorkspace {
+  hasErrors(): Promise<boolean>
+  getErrors(): Promise<Error[]>
   refreshTree(): Promise<void>
   getTree(): Promise<MBeanTree>
   hasMBeans(): Promise<boolean>
@@ -26,9 +28,25 @@ class Workspace implements IWorkspace {
   private pluginUpdateCounter?: number
   private treeWatchRegisterHandle?: Promise<number>
   private treeWatcherCounter?: number
+  private _errors: Error[] = []
+
+  async hasErrors(): Promise<boolean> {
+    await this.getTree()
+    return this._errors.length > 0
+  }
+
+  async getErrors(): Promise<Error[]> {
+    await this.getTree()
+    return this._errors
+  }
+
+  addError(error: Error) {
+    this._errors.push(error)
+  }
 
   async refreshTree() {
     this.tree = undefined
+    this._errors = []
     await this.getTree()
     eventService.refresh()
   }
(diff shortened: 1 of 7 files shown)


phantomjinx:
So I found I was having issues with the workspace being empty (due to an unrelated error) and the console message that the workspace was empty was buried in the console. What was more obvious was the blank page when a connection was attempted.

Therefore, the idea is to have a status plugin pop up that checks on the workspace. If it succeeds then the plugin never appears. However, if there is a failure then at least a single plugin is loaded and displayed giving the user some indication of the issue.

Screenshot_20241016_161139

packages/hawtio/src/plugins/app-status/globals.ts
tadayosi, line 1: I'm not sure if the name "app status" is the good one for the plugin. Because "app" can mean the server application that embeds the Hawtio console. What about naming it "Console Status"?

packages/hawtio/src/plugins/shared/workspace.ts
```diff
@@ -44,11 +62,13 @@ class Workspace implements IWorkspace {
 
   private async loadTree(): Promise<MBeanTree> {
     if (!(await userService.isLogin())) {
-      throw new Error('User needs to have logged in to use workspace')
+      this.addError(new Error('User needs to have logged in to use workspace'))
+      return MBeanTree.createEmpty(pluginName)
```
tadayosi, line 66: This changes the behaviour when the user is not logged in yet. Is it really safe?  Can't we just reserve the original behaviour while recording the error in the workspace?

packages/hawtio/src/plugins/shared/workspace.ts
```diff
@@ -26,9 +28,25 @@ class Workspace implements IWorkspace {
   private pluginUpdateCounter?: number
   private treeWatchRegisterHandle?: Promise<number>
   private treeWatcherCounter?: number
+  private _errors: Error[] = []
```
tadayosi, line 31: Why does this need to be `Error` class?  Can't we just use `string[]` with only error messages here?

packages/hawtio/package.json
tadayosi, line 1: As this pull req is introducing a new plugin, the commit message should be `feat:` rather than `fix:` so that it triggers minor version increase when a new version of hawtio/react is released (i.e. 1.6.0 rather than 1.5.1).

tadayosi (changes requested):
Overall the enhancement looks great, but there are a few considerations we'd like to take care of a bit.

packages/hawtio/src/plugins/shared/workspace.ts
phantomjinx: The benefit was the preservation of the actual cause of the error caught by adding it to the `cause` property of an Error.

PR #1190, opened 2024-10-28
Pr 1185 fix

grgrzybek:


grgrzybek:
keycloak/keycloak#33080 broke Jest tests, because in `keycloak-js` <=25 we had:
```json
"exports": {
  ".": {
    "import": {
      "types": "./dist/keycloak.d.mts",
      "default": "./dist/keycloak.mjs"
    },
    "require": {
      "types": "./dist/keycloak.d.ts",
      "default": "./dist/keycloak.js"
    }
  },
},
```

and in 26+ we have:
```json
"exports": {
  ".": {
    "types": "./lib/keycloak.d.ts",
    "default": "./lib/keycloak.js"
  },
},
```

Thanks to the `mjs` extension, Jest was working well with ESM code and now ESM code is used in `js` file, which breaks Jest...

cc: @tadayosi

packages/hawtio/tsconfig.json
```diff
@@ -16,7 +16,7 @@
     "noImplicitThis": true,
     "noUncheckedIndexedAccess": true,
     "module": "esnext",
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
```
tadayosi, line 19: is it required? I am a little bit worried about any potential effects with it.

tadayosi (approved):

PR #1286, opened 2025-01-06
feat(connect): enable auto-connect based on preset connections from backend

tadayosi:
Required by hawtio/hawtio#3731

This enhancement introduces support for the new backend API `/preset-connections` which can return a set of preset connections from the Hawtio CLI `--connection` option and automatically connecting to them.

### DIFF
--- a/packages/hawtio/src/plugins/shared/__mocks__/connect-service.ts
+++ b/packages/hawtio/src/plugins/shared/__mocks__/connect-service.ts
@@ -1,19 +1,23 @@
-import 'jolokia.js'
 import Jolokia, { IJolokiaSimple } from '@jolokia.js/simple'
+import 'jolokia.js'
 import {
+  ConnectStatus,
   Connection,
   ConnectionCredentials,
   ConnectionTestResult,
   Connections,
   IConnectService,
   LoginResult,
-  ConnectStatus,
 } from '../connect-service'
 
 class MockConnectService implements IConnectService {
   constructor() {
     // eslint-disable-next-line no-console
     console.log('Using mock connect service')
+  }
+
+  getCurrentConnectionId(): string | null {
+    return null
   }
 
   getCurrentConnectionName(): string | null {
@@ -72,7 +76,7 @@ class MockConnectService implements IConnectService {
     return ''
   }
 
-  getJolokiaUrlFromName(name: string): string | null {
+  getJolokiaUrlFromId(name: string): string | null {
     return null
   }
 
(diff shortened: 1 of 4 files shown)


packages/hawtio/src/plugins/shared/__mocks__/connect-service.ts
```diff
@@ -72,7 +76,7 @@ class MockConnectService implements IConnectService {
     return ''
   }
 
-  getJolokiaUrlFromName(name: string): string | null {
+  getJolokiaUrlFromId(name: string): string | null {
```
grgrzybek, line 79: Thanks - I should've done it when adding ID support for connections...

packages/hawtio/src/plugins/shared/connect-service.ts
```diff
@@ -62,35 +63,44 @@ export const SESSION_KEY_CURRENT_CONNECTION = 'connect.currentConnection'
 export const PARAM_KEY_CONNECTION = 'con'
 export const PARAM_KEY_REDIRECT = 'redirect'
 
-const LOGIN_PATH = '/connect/login'
+const PATH_LOGIN = '/connect/login'
+const PATH_PRESET_CONNECTIONS = 'preset-connections'
```
grgrzybek, line 67: Shouldn't it be also an URI with slash? (to make it less context-dependent)?

packages/hawtio/src/plugins/shared/connect-service.ts
```diff
@@ -351,13 +433,13 @@ class ConnectService implements IConnectService {
...
       // Authorization header is already "taken" by OIDC/Keycloak authenticator
-      const headers = {
+      const headers: Record<string, string> = {
         'X-Jolokia-Authorization': basicAuthHeaderValue(username, password),
       }
       const token = getCookie('XSRF-TOKEN')
       if (token) {
         // For CSRF protection with Spring Security
-        ;(headers as Record<string, string>)['X-XSRF-TOKEN'] = token
+        headers['X-XSRF-TOKEN'] = token
```
grgrzybek, line 442: Much better than my struggle with TypeScript ;)

packages/hawtio/src/plugins/shared/__mocks__/connect-service.ts
tadayosi: 👍

packages/hawtio/src/plugins/shared/connect-service.ts
tadayosi: 🙂

packages/hawtio/src/plugins/shared/connect-service.ts
tadayosi: Thanks, you are right. Fixed.

PR #1354, opened 2025-02-14
fix(jolokia): stop Jolokia instance in JolokiaService.reset() (fixes …

grgrzybek:
…#1353)

### DIFF
--- a/packages/hawtio/src/plugins/shared/jolokia-service.ts
+++ b/packages/hawtio/src/plugins/shared/jolokia-service.ts
@@ -178,6 +178,10 @@ class JolokiaService implements IJolokiaService {
 
   reset() {
     this.jolokiaUrl = undefined
+    this.jolokia?.then(jolokia => {
+      jolokia.stop()
+      return
+    })
     this.jolokia = undefined
     this.config = {
       method: JolokiaListMethod.DEFAULT,


phantomjinx:
That should work. The asynchronous might mean 1 or 2 more updates are made before it does stop but that should be fine I think.

packages/hawtio/src/plugins/shared/jolokia-service.ts
```diff
@@ -178,6 +178,10 @@ class JolokiaService implements IJolokiaService {
 
   reset() {
     this.jolokiaUrl = undefined
+    this.jolokia?.then(jolokia => {
+      jolokia.stop()
+      return
```
tadayosi, line 183: it's not needed

tadayosi (approved):

packages/hawtio/src/plugins/shared/jolokia-service.ts
grgrzybek: hmm, we don't need to call `stop()`?

tadayosi: I mean `return` is not necessary here.

grgrzybek: I'll remember ;)

PR #1356, opened 2025-02-14
Adds hawt config properties to change appearance of the header and sidebar

phantomjinx:
### Overview
Since the Hawtio-Online dynamic console plugin has much more limited space, it is necessary to re-organise the UI components. Therefore,
 - Being able to show/hide the header bar. This is no longer strictly necessary as the latest version of the console plugin still makes use of the header bar but figured this option was still useful, hence its inclusion;
 - Being able to hide the brand logo. This is out of place in a sub-component tab of the OCP console and takes up a significant amount of space so essential to hide it;
 - Being able to hide the User dropdown. This is redundant in the OCP console since the user is displayed at the top of the console, the Preferences are integrated in the console and logout action is redundant;
 - Being able to hide the sidebar. The sidebar is wide and unnecessary. The console-plugin provides a toolbar dropdown that is applied to the header bar so much more compact.

### Commit Changes
- Creates 'appearance' property group in hawtconfig with properties that determine whether
  - showHeader: to show/hide the main header bar
  - showBrand: to show/hide the brand logo in the header bar
  - showUserHeader: to show/hide the user header dropdown in the header bar
  - showSideBar: to show/hide the sidebar

- config-manager.ts
  - Introduction of Appearance type
  - Functions to fetch the value of the properties from a hawt config with the defaults being true if properties are not specified

- HawtioHeader.tsx / HawtioPage.tsx
  - Make the various components conditional according to the properties

### DIFF
--- a/packages/hawtio/src/core/config-manager.ts
+++ b/packages/hawtio/src/core/config-manager.ts
@@ -16,6 +16,11 @@ export type Hawtconfig = {
   branding?: BrandingConfig
 
   /**
+   * Configuration for the placement and structure of the UI
+   */
+  appearance?: AppearanceConfig
+
+  /**
    * Configuration for the built-in login page.
    */
   login?: LoginConfig
@@ -53,6 +58,23 @@ export type BrandingConfig = {
   appLogoUrl?: string
   css?: string
   favicon?: string
+}
+
+/**
+ * Appearance configuration type.
+ */
+export type AppearanceConfig = {
+  // Whether to display the header bar (default: true)
+  showHeader?: boolean
+
+  // Whether to display the brand logo on the header bar (default: true)
+  showBrand?: boolean
+
+  // Whether to display the user header dropdown on the header bar (default: true)
+  showUserHeader?: boolean
+
+  // Whether to display the sidebar (default: true)
+  showSideBar?: boolean
 }
 
 /**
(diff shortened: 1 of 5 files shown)


packages/hawtio/src/core/config-manager.ts
```diff
@@ -236,6 +258,66 @@ class ConfigManager {
...
+   * eg. the #useHawtconfig hook
+   */
+  isSideBarShown(hawtconfig: Hawtconfig): boolean {
+    if (!hawtconfig) return true
+
+    const appearance = hawtconfig.appearance
+    if (!appearance) return true
+
+    return appearance.showSideBar ?? true
+  }
```
tadayosi, line 320: These `isXxx` methods don't add any values so not necessary. We can just chain the fields with `xxx?.yyy` in one line. Please remove them. The design of the methods is not cool as well, as they accepts `hawtconfig` while it should be already there in the object field. This aspect further clutters the interface API.

packages/hawtio/src/ui/page/HawtioPage.tsx
```diff
@@ -63,14 +64,17 @@ export const HawtioPage: React.FunctionComponent = () => {
     sessionService.userActivity()
   }
 
+  const headerShown = configManager.isHeaderShown(hawtconfig)
+  const sideBarShown = configManager.isSideBarShown(hawtconfig)
```
tadayosi, line 68: here you can just do:
```
const sideBarShown = hawtconfig.appearance?.showSideBar ?? true
```
Same for the others.

tadayosi (approved):
Well, as I suggested before, the other way is to provide a `HawtioHeadlessPage` along with `HawtioPage` and the downstream app assembers can choose which page to use. I thought that approach was cleaner than introducing an omni option like the `appearance` to Hawtconfig.

But let's go with the option you proposed.

PR #1405, opened 2025-03-17
fix(camel): default view should be Route Diagram

### LINKED ISSUE #1378: Camel - Default view should be Route Diagram
Currently the default view for Camel plugin is the first tab in the content pane, which is Attributes. It's not the most interesting view for most of the users. Generally, the users would like to view what the Camel application is like at a glance, so the Route Diagram should serve the best.

We should either:
* Simply make the Route Diagram the default for Camel plugin, or
* Provide some config option for each plugin to specify which view is default.

tadayosi:
Fix #1378

### DIFF
--- a/packages/hawtio/src/plugins/camel/CamelContent.tsx
+++ b/packages/hawtio/src/plugins/camel/CamelContent.tsx
@@ -3,6 +3,7 @@ import { AttributeValues, Attributes, Chart, JmxContentMBeans, MBeanNode, Operat
 import {
   Divider,
   EmptyState,
+  EmptyStateHeader,
   EmptyStateIcon,
   EmptyStateVariant,
   Nav,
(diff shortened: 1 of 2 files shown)


grgrzybek (approved):

tadayosi:
Two issues found by @jsolovjo:

1. > I tried to reproduce it manually, I added a new endpoint from URI and UI doesn't add it automatically, you have to either refresh the page or change a view, e.g. go to another page and return back, and only then the newly added endpoint is listed
this covers the scenario when you add a new endpoint from URI on endpoints page
and the same applies when you add an endpoint from Data

2. > Also, what I noticed is by some reason when you start/stop a route on routes (page) -> Routes (tab) it redirects you to Route Diagram tab

tadayosi:
All checks have passed thanks to @jsolovjo. The issues will be addressed at #1409 separately.

PR #1412, opened 2025-03-21
fix(springboot-plugin): Set background to PF5 style.

joshiraezcode:
Sets the background/padding to the PF5 style.

I also added a bit of padding to the icons as they were too close to the margin.

Without padding:

Screenshot 2025-03-21 011640

With the padding:

Screenshot 2025-03-21 012756


tadayosi (approved):
Thanks. Please fix CI error(s) and we are good to go.

mmelko (approved):

PR #1414, opened 2025-03-24
update(router): Upgrade router to 7 (#1282)

### LINKED ISSUE #1282: Fix React Router future flag warnings
Screenshot 2024-12-27 at 15 16 44

joshiraezcode:
Fixes #1282 

Updates the router component to use new features found in version 7. The guidelines https://reactrouter.com/upgrading/v6

There is a new warning, but it contradicts the fixes needed for https://reactrouter.com/upgrading/v6#v7_relativesplatpath. Trying to follow it is like we had before, and breaks the page. For now I have ignored it as I think that when we upgrade to 7 it should be removed.

426104904-014f05f6-aefe-4eca-b0d2-1a6ec161dfbe

### DIFF
--- a/packages/hawtio/src/Hawtio.tsx
+++ b/packages/hawtio/src/Hawtio.tsx
@@ -17,7 +17,7 @@ export const Hawtio: React.FunctionComponent<HawtioProps> = props => {
   }
 
   return (
-    <BrowserRouter basename={hawtio.getBasePath()}>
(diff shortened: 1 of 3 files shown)


tadayosi:
Oops, sorry I accidentally merged it before fully reviewing it. It looks ok to me, but Paul had been struggling with the React Router version when working on the OCP dynamic plugin, so I think it's better to ask his opinion. @phantomjinx What do you think about this change?  Doesn't it affect the dynamic plugin in any way?

PR #1454, opened 2025-04-09
fix(runtime, logs, shared): Fixed colum width changing in tables (#1452)

### LINKED ISSUE #1452: Filtered table changes column width when interacting with it
Spotted while working on #1435 by @jsolovjo 

Column width on FilteredTable component seems to change width when interacting with it, and it's not apparent what sets the width.

https://github.com/user-attachments/assets/4bafad47-ebef-4bce-a122-744d3e614313

joshiraezcode:
Fixes #1452 

Sets a new field to configure FilteredTable to add the width desired for the table. It's recommended to use it to ensure that the columns stay at a fixed width and the content is adjusted accordingly. Even then, there is no guarantees that a specially long string wont break the width but for now it seems working fine.

https://github.com/user-attachments/assets/adef98f1-9d8e-492e-be3c-b94c10a8c6a7

Also sets a width for JMXContentMBeans because that table was jumping widths while exploring through the MBeans in JMX and Camel, which looked odd.

I haven't touched other tables as they can't be filtered or paginated as far as I've seen, so the width will stay fixed for that table as all the content has already been loaded. Also, as the tables are different depending on the attributes we are viewing, it's normal for those tables to have variable widths depending on the data shown.

### DIFF
--- a/packages/hawtio/src/ui/util/FilteredTable.tsx
+++ b/packages/hawtio/src/ui/util/FilteredTable.tsx
@@ -29,15 +29,13 @@ import { SearchIcon } from '@patternfly/react-icons'
 import { ExpandableText } from './ExpandableText'
 import { isNumber, objectSorter } from '@hawtiosrc/util/objects'
 
-type AvailablePercentageWidths = 10 | 15 | 20 | 25 | 30 | 35 | 40 | 45 | 50 | 60 | 70 | 80 | 90 | 100
-
 interface Props<T> {
   extraToolbar?: React.ReactNode
   tableColumns: {
     name?: string
     key?: keyof T
     renderer?: (value: T) => React.ReactNode
-    percentageWidth?: AvailablePercentageWidths
+    percentageWidth?: ThProps['width']
   }[]
   rows: T[]
   searchCategories: { name: string; key: keyof T }[]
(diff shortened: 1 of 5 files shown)


packages/hawtio/src/ui/util/FilteredTable.tsx
```diff
@@ -29,9 +29,16 @@ import { SearchIcon } from '@patternfly/react-icons'
 import { ExpandableText } from './ExpandableText'
 import { isNumber, objectSorter } from '@hawtiosrc/util/objects'
 
+type AvailablePercentageWidths = 10 | 15 | 20 | 25 | 30 | 35 | 40 | 45 | 50 | 60 | 70 | 80 | 90 | 100
```
tadayosi, line 32: Instead of introducing a new type, you can simply use `ThProps['width']` here.

tadayosi (approved):
One minor comment, but otherwise it looks good to me.

packages/hawtio/src/ui/util/FilteredTable.tsx
joshiraezcode: OH, Thanks a lot Tada, I wanted something like that but couldn't find how to do it. I'm going to change it, reupload and merge.

PR #1457, opened 2025-04-10
fix: Decrements @patternfly/react-code-editor back to version 5

phantomjinx:
* react-code-editor was updated by Dependabot rather than a specific issue but doing this has introduced both patternfly 5 and 6 react-core and other libraries to the dependencies.

* Better if all of the @patternfly libraries were updated in a uniform way to reduce the possibility of mutliple dependent versions being included.


grgrzybek (approved):

tadayosi:
@phantomjinx FYI, if you want to include it to the next product release, you'd need to backport it to 1.9.x branch as well.

PR #1483, opened 2025-04-23
refactor(logs): Swapped positions of fixed search filter and text filter (#1479)

### LINKED ISSUE #1479: Swap fixed set filters and text filters on search box (Subtask #1436)
Subtask #1436 

Swap the fixed set filters to the left of the text filters on the search boxes of Filtered Table to mimic OpenShift search boxes

Image

joshiraezcode:
Fixes #1479 

image

I think I'm going to add a bit more effort and refactor the fixed set so the logic is inside the Filtered Table component. Is that okay @tadayosi ?

I think I'll be able to do that in the day. If it's not desirable, we can just pick the previous commit which has just the fix.

### DIFF
--- a/packages/hawtio/src/plugins/logs/Logs.tsx
+++ b/packages/hawtio/src/plugins/logs/Logs.tsx
@@ -217,7 +217,7 @@ const LogsTable: React.FunctionComponent = () => {
           setSelected(row.logEntry)
           setIsModalOpen(true)
         }}
-        extraToolbar={levelToggles}
+        fixedSetToolbar={levelToggles}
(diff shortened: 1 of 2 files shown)


joshiraezcode:
It's a bit harder than I thought and the changes merit their own task. I created a new issue and I'll be setting a different PR for that issue for the refactor. https://github.com/hawtio/hawtio-next/blob/main/packages/hawtio/src/plugins/logs/Logs.tsx#L140

It's useful to get with it now before the other search input subtask as the other pages have some fixed string searchs, and it'll ease development on those.

tadayosi (approved):

PR #1485, opened 2025-04-25
refactor(util, logs): Added fixed search prop to filtered table (#1484)

### LINKED ISSUE #1484: Add fixed string search logic to FilteredTable (Subtask #1436)
Subtask #1436 

Currently, FilteredTable is adding the fixed search input as an extra ReactNode being defined out of the class. For example, the Log class creates its own select and then passes it down to FilteredTable to be shown, having some of the filter logic on the Logs component.

https://github.com/hawtio/hawtio-next/blob/main/packages/hawtio/src/plugins/logs/Logs.tsx#L140

Although is useful to have the extraToolbar prop for extra elements like in Runtime > Threads, I feel like the fixed set search is more intrinsic to the filtered table and will be used in more places, making sense to have the logic inside the component (and will be used on the other tasks for the search bar)

This was going to be included on #1483 but it has proven a bit more work than I expected and to unblock #1483  and to properly document the change I decided to create a new issue with it with its own separate Pull Request.

joshiraezcode:
Fixes #1484 

I'd like to spend a bit more time on this but it's kinda complicated and I'd like to unblock the next tasks. This has minimal changes needed to have the fixed set search running using the component's logic. I'll revisit and do a refactor later.

https://github.com/user-attachments/assets/be82744a-a34f-43b2-80de-9146e028e1d4


packages/hawtio/src/plugins/logs/Logs.tsx
```diff
@@ -63,20 +57,15 @@ const LogsTable: React.FunctionComponent = () => {
...
-
   // Modal
   const [isModalOpen, setIsModalOpen] = useState(false)
   const [selected, setSelected] = useState<LogEntry | null>(null)
 
   const rows = useMemo(() => {
-    return logsService.filter(logs, filters).map(log => new LogRowData(log))
-  }, [logs, filters])
+    return logsService
+      .filter(logs, { level: [], logger: '', message: '', properties: '' })
```
tadayosi, line 66: I don't think this `.filter` is needed any more.

tadayosi: Also, now this `rows` changes only when `logs` is changed. Do we still need `useMemo` for it?

In addition, is `logsService.filter(..)` still needed? If no place other than this uses this, it should be removed altogether.

tadayosi (approved):
Overall, it seems nice.

I just want to remind that sometimes there are cases where separating react components are the best practice to avoid rerendering the entire page as one big component. Please double check that it is not the case here.

packages/hawtio/src/plugins/logs/Logs.tsx
joshiraezcode: You are right, I left it because of how it used to append new elements, and to minimize changes, but I might have minimized too much.

I have left the logs state still to use logService.append, and then update the rows from the new log value. The rows are now updated together with any logs update, so no need to use useMemo

joshiraezcode:
Hi @tadayosi , you are right and that's one of the things I want to improve.  Although right now the behaviour is the same as it was before as the update graph would be the same as before in the way the select was inserted.

One of the things I want to do is separate some of the components to ensure that the toggles only affect the clicked search, for example (although is a minimal gain, it's also more for clarity of code). As well as clean up a bit more of the code because right now a lot of the coupled logic has been inserted into the new Select component that will be created for every fixedSearchCategory.

But it was proven difficult as I'd have to change more parts of the components. That's why I went with just decoupling plus refactoring the logic into Filtered Table to be able to use it in Quartz plugin, and revisit it later to clean it now that it is decoupled.

tadayosi (approved):

PR #1486, opened 2025-04-26
fix: catch exception on http code 403 properly

GChuf:
Testing through activemq artemis console.

The exception is not handled properly in the current version, see screenshot.

This also causes some issues with website/browser when developer tools are open, since it takes a lot of time to open the debug tab and the webpage in the browser can freeze.

I tested this fix with hawtio v4.4 and it works.
Please check.
debug

### DIFF
--- a/packages/hawtio/src/plugins/shared/connect-service.ts
+++ b/packages/hawtio/src/plugins/shared/connect-service.ts
@@ -423,17 +423,15 @@ class ConnectService implements IConnectService {
 
   private async forbiddenReasonMatches(response: Response, reason: string): Promise<boolean> {
     // Preserve compatibility with versions of Hawtio 2.x that return JSON on 403 responses
-    return response
-      .text()
-      .then(txt => {
+    return response.text().then(txt => {
+      try {
         const json = JSON.parse(txt)
-        // exception will propagate to .catch()
         return json['reason'] === reason
-      })
-      .catch(_ => {
+      } catch (_) {
         // Otherwise expect a response header containing a forbidden reason
         return response.headers.get('Hawtio-Forbidden-Reason') === reason
-      })
+      }
+    })
   }
 
   connect(connection: Connection, current = false) {


tadayosi (approved):
Thank you! LGTM.

Please fix the lint and we can merge it. Maybe `_e` should work.

tadayosi:
OK, it's trivial so I'll take care of it. Thanks for your contribution!

GChuf:
@tadayosi sorry, didn't have time to fix it yet. I saw you fixed it, but I also did some testing.
I think you could just do this - just use catch without variable:

```
  private async forbiddenReasonMatches(response: Response, reason: string): Promise<boolean> {
    // Preserve compatibility with versions of Hawtio 2.x that return JSON on 403 responses
    return response.text().then(txt => {
      try {
        const json = JSON.parse(txt)
        return json['reason'] === reason
      } catch {
        // Otherwise expect a response header containing a forbidden reason
        return response.headers.get('Hawtio-Forbidden-Reason') === reason
      }
    })
  }
```

This compiles into this javascript:
```
  forbiddenReasonMatches(response, reason) {
    return __async(this, null, function* () {
      return response.text().then((txt) => {
        try {
          const json = JSON.parse(txt);
          return json["reason"] === reason;
        } catch (e) {
          return response.headers.get("Hawtio-Forbidden-Reason") === reason;
        }
      });
    });
  }
```

GChuf:
@tadayosi I know - I just thought maybe the above code looks cleaner.
Anyway, I tested it now with hawtio v4.4.1 and @hawtio/react 1.9.4 and everything seems OK!

I'll wait until https://github.com/hawtio/hawtio/pull/3874 is merged and then I'll include all the changes into activemq artemis.

tadayosi:
@GChuf Is hawtio/hawtio#3874 so urgent? It doesn't seem so to me. Unless otherwise, I'd wait for the next regular release of Hawtio 4.4.2 or 4.5.0 for that fix.

GChuf:
@tadayosi no worries, not urgent, do it whenever you think it's okay.
I just want to avoid updating twice :)
Do you have a fixed date for the next regular release?

If another version of artemis gets published before that, I'll update it with 4.4.1.

tadayosi:
No fixed date, but unless there is any urgent fix to be released, it will be in 2-3 months.

PR #1494, opened 2025-04-30
refactor(quartz): Table ported to Filtered Table (#1480)

### LINKED ISSUE #1480: Implement Filtered Table on Quartz Plugin > Triggers (Subtask #1436)
Subtask #1436 

Implement Filtered Table on Quartz Plugin Triggers as its style doesnt match HawtIO or OpenShift

Image

joshiraezcode:
Fixes #1480 

https://github.com/user-attachments/assets/8363e5d2-6c07-4602-961c-add24b529e48


joshiraezcode:
@tadayosi Should I remove TriggerFilter and filter Triggers from quartz service? As the table is the one doing the filtering right now, it's not used, but I feel like is an utility function that maybe some other functionality will use.

Currently no one is using these

tadayosi (approved):

tadayosi:
@joshiraezcode Yes, if there is nothing that uses the type(s) and service method(s) it makes sense to remove them.

PR #1512, opened 2025-05-14
fix(connect): When testing connections (top bar) add credentials if p…

### LINKED ISSUE #1504: Periodic check for connection availability should not spam 403 in network tab
Currently after connecting to remote Jolokia agent, topbar's behavior is to indicate connection status (red/green).
But because the same code is used for initial check of the connection (when connection is being created), there's no credentials sent (on purpose)...

We need to do it better.

grgrzybek:
…resent (closes #1504)


phantomjinx (approved):
LGTM

PR #1513, opened 2025-05-15
fix(connect): HAWNG-1008 copy paste error

tadayosi:
If agent.agent_id is undefined, it should always become 'discover-undefined'. Then it should be rather simply ''.


tadayosi:
@grgrzybek Please review it

grgrzybek (approved):

PR #1514, opened 2025-05-15
upgrade(node): Upgraded to 22

joshiraezcode:


joshiraezcode:
E2E tests not firing on the correct version. Does it pick the configuration for E2E from the main branch @tadayosi ? The configuration seems fine.

It's working for lint and test

image

joshiraezcode:
Ok I'm afraid that I have no idea why the E2E tests are not picking the correct values.

image

The commit on which is being done the tests is the correct one

https://github.com/hawtio/hawtio-next/commit/d383c40b82cc169bbc56473e9892c58cd03c8693

joshiraezcode:
I can't figure it out. I'm able to get changes or add actions in the other yaml files, but I can't figure out from where the github action for e2e tests is retrieving the values from, even though the checkout says that it's on the recent commit.

tadayosi:
@joshiraezcode Just because E2E tests use `pull_request_target` which only reads from the branch's CI configuration for security reasons. Otherwise, anonymous pull reqs can easily tweak the workflow and steal the credentials.

tadayosi (approved):

joshiraezcode:
Much appreciated for the explanation, I wasn't able to figure it out from the code.

Issue #1543, opened 2025-06-06
Size of Camel diagram nodes appear too big when viewed through console plugin UI


phantomjinx:
- see related -> https://github.com/hawtio/hawtio-online-console-plugin/issues/18

Make the diagram nodes fit in proportion to the size of the bounding box of the parent diagram.

phantomjinx:
Unfortunately, further testing of more complex Route examples (application with more than one Route), has shown the solution in https://github.com/hawtio/hawtio-next/pull/1544 to be insufficient and introduces a regression which lays the route diagram nodes from the independent routes on top of one another in the centre of the X axis. Will need to at a minimum revert the commit [9212f15](https://github.com/hawtio/hawtio-next/pull/1544/commits/9212f15280f9d3494c14647ea8c6dced198cf784) if I cannot improve it further.

Issue #1542, opened 2025-06-06
Modify css / js to align colours with theme of browser


phantomjinx:
While working on the console-plugin, some of the colour scheme chosen assumes a light theme and so some css has assumed this to be the case, eg. [diagram node popup](https://github.com/hawtio/hawtio-next/blob/main/packages/hawtio/src/plugins/camel/route-diagram/RouteDiagram.css#L64).

Also, the HawtioMainPage has its background colour [css](https://github.com/hawtio/hawtio-next/blob/main/packages/hawtio/src/ui/page/HawtioPage.css#L2) set to `--pf-v5-global--BackgroundColor--100`  but does not change to black when loaded into a dark themed browser. However, this does happen correctly in the console-plugin using the same css.

Altogether, there needs some investigation into correctly theming the UI to ensure it corresponds with the dark/light theme of the browser.

phantomjinx:
Addressed in https://github.com/hawtio/hawtio-next/pull/1544

PR #1544, opened 2025-06-06
Addresses both node sizing and theming of application

phantomjinx:
Apologies this is 2 commits for 2 issues but if they are submitted separately then one or other would have to be rebased once one was merged. Therefore, if you can indulge me and review the 2 commits separately for 1 PR, I'd appreciate it.

Commits

1.  fix: Aligns the size of the camel diagram content with parent bounds (#1543)
    * When opening the route diagram, gets the bounds of the parent div and
      creates the camel nodes in proportion and centres them appropriately
    * RouteDiagram.tsx
      * Splits the diagram component to provide an outer div with a reference
         that can maintain the bound box dimensions. The inner
         ReactFlowRouteDiagram is then refreshed with the correctly sized nodes
         and 'onLoad' sizes the viewport correctly using the reference's bounds
         before fitting the view.

2.  fix: Aligns the theme of the application with browser setting (#1542)
    
    * Camel Diagram
      * Colours the node popup table (odd/even) according to the theme of the
        browser
      * Reduces the padding of the table to make the popup not so prominent and
        more in keeping with the size of the diagram nodes
      * Defines the hawtio-diagram-node--BackgroundColor var to determine the
        background colour of the nodes by default or and if mode is dark
    
    *  theme.ts
      * Defines a window listener that updates the html class to the requisite
        Patternfly class (see [1]) if the browser theme is changed
      * Adds a localStorage flag if clients want to disable this functionality
         * hawtio.disableThemeListener = true in LocalStorage
    
     * index.ts
       * Initialises the theme listener when the library is loaded
    
    [1] https://www.patternfly.org/developer-resources/dark-theme-handbook/#enabling-dark-theme

### DIFF
--- a/packages/hawtio/src/Hawtio.tsx
+++ b/packages/hawtio/src/Hawtio.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useEffect } from 'react'
 import { BrowserRouter, Route, Routes } from 'react-router-dom'
 import { hawtio } from './core'
 import './Hawtio.css'
@@ -15,6 +15,24 @@ export const Hawtio: React.FunctionComponent<HawtioProps> = props => {
   if (basepath) {
     hawtio.setBasePath(basepath)
   }
+
+  /*
+   * Initialise a window theme listener to update the application theme
+   * depending on the browser's chosen / default theme.
+   *
+   * Note: Will be ignored if a hawtio.disableThemeListener flag is set
+   * to true in localStorage.
+   */
+  hawtio.addWindowThemeListener()
+
+  useEffect(() => {
+    return () => {
+      /*
+       * Clean up the window listener on unmount
+       */
+      hawtio.removeWindowThemeListener()
+    }
+  }, [])
 
   return (
     <BrowserRouter future={{ v7_startTransition: true, v7_relativeSplatPath: true }} basename={hawtio.getBasePath()}>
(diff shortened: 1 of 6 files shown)


packages/hawtio/src/util/theme.ts
```diff
@@ -0,0 +1,47 @@
...
+    document.documentElement.classList.remove(PATTERNFLY_THEME_CLASS)
+  }
+}
+
+/**
+ * Adds an event listener to the window theme to update
+ * css values in the event of a change in theme
+ */
+export function initWindowThemeListener() {
+  const disableListener = localStorage.getItem(HAWTIO_DISABLE_THEME_LISTENER) || 'false'
```
tadayosi, line 35: use `??` instead of `||`

packages/hawtio/src/util/theme.ts
tadayosi, line 1: we can regard them as a core function and move all of them to methods in `HawtioCore`.
I don't think it's good to overusing utils.

packages/hawtio/src/index.ts
```diff
@@ -12,3 +13,7 @@ export * from './ui'
 
 // Register Hawtio React component version
 configManager.addProductInfo('Hawtio React', '__PACKAGE_VERSION_PLACEHOLDER__')
+
+// Initialise a listener to update the application theme
+// depending on the browser's chosen theme.
+initWindowThemeListener()
```
tadayosi, line 19: Since the theme is a UI concern, why can't we invoke it at `Hawtio` react component instead of here?

You should be able to invoke it before or after this part:
https://github.com/hawtio/hawtio-next/blob/4f29a26c14802d36023f2be9f1473423c46e2bf7/packages/hawtio/src/Hawtio.tsx#L15-L17

tadayosi (changes requested):
Looks great but I think we shouldn't overuse utils. Please read my comments.

PR #1571, opened 2025-06-17
Changes the test app source map generation to reduce the memory footprint

phantomjinx:
* eval-source-map is heavy on memory and not meant for production. See https://webpack.js.org/configuration/devtool

* For development, this option can be changed locally if developers require it. Most of the time, it seems people do not so disable entirely

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,30 @@ yarn lint && yarn format:check
 
 Read [Running E2E tests locally](./docs/e2e.md).
 
+### Debugging with Source-Maps
+
+When `yarn start` is run, the test application located in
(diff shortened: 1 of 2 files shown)


grgrzybek (approved):
Fully agree!

tadayosi (changes requested):
Sorry, but I don't understand the reason for the change. `app` is always meant for development, not production.

tadayosi:
@phantomjinx OK, then can you add an instruction for how to turn it on here?  I'm fine with it then.
https://github.com/hawtio/hawtio-next?tab=readme-ov-file#developing

tadayosi (approved):
awesome, thank you!

Issue #1570, opened 2025-06-17
Exception thrown when route diagrams are grouped


phantomjinx:
When there are a number of routes under the context and the diagrams are grouped then the assumption that the parent of the [diagram](https://github.com/hawtio/hawtio-next/blob/main/packages/hawtio/src/plugins/camel/routes-service.ts#L217) has the context in its metadata falls down. This results in the dumping of the xm stats failing with an error.

![Image](https://github.com/user-attachments/assets/1575059b-1b57-4548-8919-a7148f58e199)

PR #1572, opened 2025-06-17
Fix: Handles route statistics in multi route applications (#1570)

phantomjinx:
* If an application has multiple routes and grouped then a group node will sit between the route nodes used for diagramming and the Routes node than has access to the statistics XML. This means the Routes node is not selected, the context node not found and an undefined mbean used for retrieving the XML.

* Since the function must support both the Routes diagram and Route diagrams, uses a while loop to traverse up the tree to the Routes node.

* Adds a iteration limit on the loop as a safety feature in case the tree has in same way become cyclical, thereby avoiding possiblity of an infinite loop.


tadayosi (approved):
Looks great but can we have a test for the issue?

It's a bit hard to understand what the routes having the issue really look like.

We should definitely have a test scenario at E2E for hawtio/hawtio as well. cc @jsolovjo

PR #1573, opened 2025-06-18
fix: Sort out the node alignment to support all node diagrams (#1543)

phantomjinx:
* Fits the view, zooms and centres nodes correctly in both Routes and Route diagrams

* visualization-service.ts
  * Reverts to previous version prior to c1520188b

* RouteDiagram.tsx
  * Moves all state back into single component although retains the outer RouteDiagram to allow for the use of the ReactFlowProvider which is required for the useReactFlow and useNodesInitialized hooks
  * Adds useEffect on nodesInitialized - only when true is fitView called as this ensures the nodes have been given their dimensions. Calling any sooner, fitView does not work properly (see [1])
  * Separate out the refreshing of stats from the loading of nodes

* RouteDiagram.css
  * Removes absolute measurements in view height to ensure the portability of the component and ensures that fitView correctly renders the viewport correctly

[1] https://github.com/xyflow/xyflow/issues/533

### DIFF
--- a/packages/hawtio/src/plugins/camel/CamelContent.css
+++ b/packages/hawtio/src/plugins/camel/CamelContent.css
@@ -7,9 +7,14 @@
 }
 #camel-content {
   height: 100%;
+  display: flex;
+  flex-direction: column;
 }
 #camel-content-header > #camel-contexts-toolbar {
   float: right;
   padding-top: 0;
   padding-bottom: 0.5em;
+}
+#camel-content-main {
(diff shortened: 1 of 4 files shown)


phantomjinx:
These changes fix the original regression in 11c1d09fcc1b and ensures the diagram canvas is portable to the console-plugin

grgrzybek (approved):
Wow - this looks complex - must be correct ;)

PR #1575, opened 2025-06-23
fix: Makes the showVerticalNav preference shown only if sidebar is visible

phantomjinx:
If the hawtconfig.json has disabled the sidebar, eg. in the console-plugin, then the preference for setting the sidebar as shown/collapsed is irrelevant. Better to not display it at all.

### DIFF
--- a/packages/hawtio/src/preferences/HomePreferences.tsx
+++ b/packages/hawtio/src/preferences/HomePreferences.tsx
@@ -15,19 +15,31 @@ import {
 import React, { useState } from 'react'
 import { useNavigate } from 'react-router-dom'
 import { preferencesService } from './preferences-service'
+import { useHawtconfig } from '@hawtiosrc/core'
+import { HawtioLoadingPage } from '@hawtiosrc/ui'
 
-export const HomePreferences: React.FunctionComponent = () => (
-  <CardBody>
-    <Form isHorizontal>
-      <FormSection title='UI' titleElement='h2'>
-        <UIForm />
-      </FormSection>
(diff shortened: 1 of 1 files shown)


grgrzybek (approved):
:+1:

jsolovjo:
> Detailed summary

Hi @tadayosi 
Actually, I see 2 test scenarios are failing and for some reason they are connected to breakpoints in Debug mode
I will have a look

jsolovjo:
Hi @phantomjinx @tadayosi 
I checked the screenshot attached during the E2E and I also ran hawtio-next Paul's dev branch and it seems the issue is with the Debug page, see the screenshots both from the automation and my manual check

Screenshot 2025-06-24 at 10 08 34

joshiraezcode:
I followed @jsolovjo steps and got to his same screen, where he can see the debug diagram.

When trying to stop and start debugging, I also found this error

image

Interestingly, this doesn't happen if you try to checkout paul's branch. I just set up Paul's as a new remote and checkout his branch, git log seems the same but in this case the diagram works correctly

We are currently investigating to see what else is there. So far it seems Debug diagram is broken.

joshiraezcode:
The original PR - https://github.com/hawtio/hawtio-next/pull/1573

Was merged even though these tests were failing there

https://github.com/hawtio/hawtio-next/runs/44352590504

Why did github flag this as green even though tests were failing????????? 

Every PR after that has been red and merged anyways, with the same tests failing

https://github.com/hawtio/hawtio-next/runs/44352590504

tadayosi (approved):

Issue #1574, opened 2025-06-20
Make OIDC plugin more reliable


grgrzybek:
When working on:
* #1556

I've already added some improvements:
* OIDC login is performed on demand - there's a button at new login page
* Because we no longer should rely on [OpenID Connect Session Management 1.0](https://openid.net/specs/openid-connect-session-1_0.html), we implement silent login by explicit _silent login_ with `prompt=none` - without it we couldn't even refresh the page. But it's working now - on browser refresh we check special flag (no sensitive information) in local storage which tells Hawtio to try starting _silent_ Authorization Flow - for a second we see redirect state in browser location (state, iss, ... params) which are used to get new access/refresh/id tokens - this works when Keycloak specific cookies are available for Keycloak origin
* logout uses `post_logout_redirect_uri` parameter so we end up on Hawtio, not on Keycloak page

However there are few more things to improve (which weren't working before #1556 anyway):
* [x] OIDC provider may be down when Hawtio starts, so OIDC initialization code shouldn't run _only_ on Hawtio initialization
* [x] OIDC provider server may be down after proper initialization (i.e., fetching `.well-known/openid-configuration`
* [x] OIDC provider may be down during refresh - Hawtio will fail silent authentication, so main login page will be presented
* [x] OIDC provider may be down during logout - without special check we end up with "Unabled to connect" browser page
* [x] We should handle all errors related to non successful authorization with properly configured OIDC/Keycloak
* [x] Currently user-service's `user` promise is readonly and resolved only once. If we want to let user login using different method after failed OIDC attempt, we need to change the way this promise is resolved (cc: @tadayosi, @phantomjinx)

Issue #1556, opened 2025-06-13
Support multiple authentication methods at runtime


grgrzybek:
This is a client counterpart of:
* hawtio/hawtio#2941

Issue #1535, opened 2025-06-02
Duplicate initialization of allegedly shared module


grgrzybek:
There's something wrong with Module Federation shared modules.

In https://github.com/hawtio/hawtio-next/blob/d4748c33cfea7ff67a57d561413adfe11952e89b/app/webpack.config.cjs#L151-L158 we declare a "plugin" which is loaded using Module Federation.

However I observed that some configuration fetches (`/auth/config`) are performed twice - with different stack trace:

1:
```
	fetchPath
OidcService
../packages/hawtio/dist/chunk-GYBN24F5.js
__webpack_require__
../packages/hawtio/dist/index.js
__webpack_require__
__webpack_require__.I/</</<
__webpack_require__.f.consumes/</onFactory/__webpack_require__.m[id]
__webpack_require__
__webpack_require__.t
(Async: promise callback) ./src/bootstrap.tsx
__webpack_require__
(Async: promise callback) ./src/index.ts
__webpack_require__
<anonymous>
<anonymous>
```

2.
```
	fetchPath
OidcService
../packages/hawtio/dist/chunk-GYBN24F5.js
__webpack_require__
../packages/hawtio/dist/index.js
__webpack_require__
__webpack_require__.I/</</<
__webpack_require__.f.consumes/</onFactory/__webpack_require__.m[id]
__webpack_require__
./src/examples/remote/index.ts
__webpack_require__
webpack/container/entry/app/./remote/</<
../node_modules/@module-federation/utilities/dist/index.cjs.js/_ref</<
step
../node_modules/@module-federation/utilities/dist/index.cjs.js/_ts_generator$1/verb/<
asyncGeneratorStep$1
_next
(Async: promise callback) asyncGeneratorStep$1
_next
../node_modules/@module-federation/utilities/dist/index.cjs.js/_async_to_generator$1/</<
../node_modules/@module-federation/utilities/dist/index.cjs.js/_async_to_generator$1/<
importRemote
../packages/hawtio/dist/chunk-PORUKUKV.js/loadExternalPlugins/<
loadExternalPlugins
loadPlugins
bootstrap
./src/bootstrap.tsx/<
(Async: promise callback) ./src/bootstrap.tsx
__webpack_require__
(Async: promise callback) ./src/index.ts
__webpack_require__
<anonymous>
<anonymous>
```

when I removed shared module, everything is fine.

I need to analyze, because errors like this cause problems with React Router and react-chart...

Issue #1328, opened 2025-01-31
hawtio-auth-keycloak doesn't show any error when trying to log in in insecure context with keycloak


mmuzikar:
When hawtio is running in an http mode (not localhost) and user logs in via keycloak they get redirected to "normal" Hawtio login page. Which causes some confusion leaving the user thinking that they must've not configured the application correctly, however in the console there's an error message: 
```

[hawtio-auth-keycloak] Failed to initialise Keycloak: Error: Web Crypto API is not available.
    m http://172.17.0.1:8080/hawtio/static/js/4240.1d53be00.chunk.js:2
    createLoginUrl http://172.17.0.1:8080/hawtio/static/js/4240.1d53be00.chunk.js:2
    login http://172.17.0.1:8080/hawtio/static/js/4240.1d53be00.chunk.js:2
    login http://172.17.0.1:8080/hawtio/static/js/4240.1d53be00.chunk.js:2
    t http://172.17.0.1:8080/hawtio/static/js/4240.1d53be00.chunk.js:2
    d http://172.17.0.1:8080/hawtio/static/js/4240.1d53be00.chunk.js:2
    m http://172.17.0.1:8080/hawtio/static/js/4240.1d53be00.chunk.js:2
4240.1d53be00.chunk.js:2:4215134

```

If possible there should be some toast notification saying that keycloak login doesn't work in insecure contexts.

----
Related issues:
* #1556
* #1574

PR #1580, opened 2025-06-25
Big Authentication Changes

grgrzybek:
Please forgive me, but I can't split it ;)

This is a PR which combines several issues:
* #1535
* #1556
* #1574
* #1328

Here are key changes in this PR:
* `<HawtioLogin>` content is based on the JSON returned from `/auth/config/login` endpoint which lists available login methods
* `form` is the default (if not specified) and works like before (but we can configure login/logout endpoints used)
* I've separated Patternfly and non-Patternfly code:
  * modules that use Patternfly _should_ be imported with `import()` instead of `import`
  * this speeds up loading
  * this allowed me to implement `<HawtioInitialization>` React (but not Patternfly) component, which shows initialization process
  * `configManager` exposes `initItem()` methods, so Hawtio internals can _mark_ states of initialization
  * main UI (`<Hawtio>` component) is passed to `root.render()` _only_ after full initialization:
```javascript
m.hawtio.bootstrap().then(() => {
  import("@hawtio/react/ui").then(m => {
    root.render(
      <React.StrictMode>
        <m.Hawtio />
      </React.StrictMode>,
    )
  })
})
```
* because full initialization is performed before showing UI, we don't have to load `keycloak` and `oidc` plugins first (← @phantomjinx)
* neither `keycloak` nor `oidc` plugins start the Authorization Flow without user interaction - they can perform _silent login_ if needed
* user can start Keycloak / OIDC login after clicking a button - this is consistent with OpenShift login
* I solved the problem with loading MF modules - these work both dynamically and through webpack _remotes_ - see `webpack.config.js` in `app/`.

### DIFF
--- a/packages/hawtio/src/core/config-manager.ts
+++ b/packages/hawtio/src/core/config-manager.ts
@@ -508,6 +508,7 @@ export class ConfigManager implements IConfigManager {
 
     const { branding } = await this.getHawtconfig()
     if (!branding) {
+      this.initItem('Applying branding', TaskState.skipped, 'config')
       return false
     }
 
@@ -556,6 +557,11 @@ export class ConfigManager implements IConfigManager {
     if ('disabled' in elm) {
       elm.disabled = false
     }
+  }
+
+  async isRouteEnabled(path: string): Promise<boolean> {
+    const { disabledRoutes } = await this.getHawtconfig()
+    return !disabledRoutes || !disabledRoutes.includes(path)
   }
 
   /**
--- a/packages/hawtio/jest.config.ts
+++ b/packages/hawtio/jest.config.ts
@@ -19,10 +19,11 @@ const config: Config = {
     'monaco-editor': path.resolve(__dirname, './src/__mocks__/monacoEditor.js'),
     '@monaco-editor/react': path.resolve(__dirname, './src/__mocks__/monacoEditor.js'),
     '@patternfly/react-code-editor': path.resolve(__dirname, './src/__mocks__/codeEditorMock.js'),
-    '@patternfly/react-icons/dist/esm/icons/ellipsis-v-icon': path.resolve(
-      __dirname,
-      './src/__mocks__/ellipsis-v-icon.js',
-    ),
+    // Jest prefers cjs version of react-icons
+    '@patternfly/react-icons/dist/esm/icons/(.*)':
+      '<rootDir>/../../node_modules/@patternfly/react-icons/dist/js/icons/$1',
+    // Jest prefers cjs version of react-tokens
+    '@patternfly/react-tokens/dist/esm/(.*)': '<rootDir>/../../node_modules/@patternfly/react-tokens/dist/js/$1',
     oauth4webapi: path.resolve(__dirname, './src/__mocks__/oauth4webapi.js'),
   },
 
--- a/packages/hawtio/src/__mocks__/ellipsis-v-icon.js
+++ b/packages/hawtio/src/__mocks__/ellipsis-v-icon.js
@@ -1,4 +1,0 @@
-exports = {
-  // no idea why only this one, but it fixes all test issues
-  EllipsisVIcon: jest.fn(),
-}


grgrzybek:
Sure - I'll check the test failures, but if you have time please check my changes - I'd be happy to discuss them live.

packages/hawtio/src/plugins/connect/init.ts
```diff
@@ -24,25 +25,35 @@ export async function isConnectionStatusActive(): Promise<boolean> {
...
+
   try {
+    configManager.initItem('Checking proxy', TaskState.started, 'config')
     const res = await fetch(PATH_PROXY_ENABLED)
     if (!res.ok) {
-      // Silently ignore and enable it when fetch failed
+      configManager.initItem('Checking proxy', TaskState.skipped, 'config')
       log.debug('Failed to fetch', PATH_PROXY_ENABLED, ':', res.status, res.statusText)
-      return true
+      return false
```
tadayosi, line 41: shouldn't we assign to `proxyEnabled` here as well?

packages/hawtio/src/plugins/connect/init.ts
```diff
@@ -24,25 +25,35 @@ export async function isConnectionStatusActive(): Promise<boolean> {
   return connectService.getCurrentConnectionId() !== null
 }
 
+let proxyEnabled: boolean | null = null
+
 async function isProxyEnabled(): Promise<boolean> {
+  if (proxyEnabled != null) {
+    return proxyEnabled
+  }
```
tadayosi, line 33: Be mindful that even in the single threaded model this guard doesn't perfectly eliminate the possibility of the following code being executed twice, as there are `await`s before setting the variable in the following.

packages/hawtio/src/plugins/shared/attributes/AttributeTable.tsx
```diff
@@ -1,13 +1,11 @@
...
-import { JmxContentMBeans } from '@hawtiosrc/plugins/shared/JmxContentMBeans'
 import { AttributeValues } from '@hawtiosrc/plugins/shared/jolokia-service'
 import { humanizeLabels } from '@hawtiosrc/util/strings'
 import { Panel } from '@patternfly/react-core'
 import { Table, Tbody, Td, Th, Thead, Tr } from '@patternfly/react-table'
 import { JolokiaErrorResponse, JolokiaSuccessResponse } from 'jolokia.js'
 import React, { useContext, useEffect, useState } from 'react'
-import { HawtioEmptyCard } from '../HawtioEmptyCard'
-import { HawtioLoadingCard } from '../HawtioLoadingCard'
+import { HawtioEmptyCard, HawtioLoadingCard, JmxContentMBeans } from '@hawtiosrc/plugins/shared'
```
tadayosi, line 8: they are within the same `shared` sub package, so I'd prefer to keep the relative path specifications. My rules are: use `@hawtiosrc` when cross-referencing, otherwise use relative.

packages/hawtio/src/plugins/shared/jolokia-service.ts
```diff
@@ -215,17 +215,19 @@ class JolokiaService implements IJolokiaService {
...
+            (version: JolokiaResponseValue) => {
+              log.info('Jolokia version:', {
+                client: jolokia.CLIENT_VERSION,
+                agent: (version as VersionResponseValue).agent,
+              })
+            },
+            error => log.error('Failed to fetch Jolokia version:', error),
+          ),
+        )
+        .then(() => true)
```
tadayosi, line 230: why do we need this?

packages/hawtio/src/ui/page/HawtioPage.tsx
```diff
@@ -36,13 +39,12 @@ export const HawtioPage: React.FunctionComponent = () => {
   // navigate should be used in effect
   // otherwise "Cannot update a component (`BrowserRouter`) while rendering a different component" is thrown
   useEffect(() => {
-    if (!isLogin && !userLoading) {
+    if (!isLogin && userLoaded) {
```
tadayosi, line 42: why this change? if user is loaded it should be already logged in. this change introduces a very different meaning to the behaviour here.
shouldn't we simply remove `!userLoading`?

tadayosi (approved):
there are minor comments but I trust you and let's approve it ;-)

tadayosi:
@grgrzybek Btw, now is not a perfect time for the huge comments to merge, as there are bugs in the E2E check now. @phantomjinx and @joshiraezcode are working to fix them. Please wait for the E2E fix as the CI results are not reliable otherwise.

joshiraezcode:
There is a lot of errors but I agree with Tadayosi that is better to wait for a bit before merging while we figure out the state of the E2E tests pipelines and fix the bug in main that is breaking E2E tests

joshiraezcode:
I don't have any idea as why the tests are failing with an exit code of 124 Error: Process completed with exit code 124. Google tells me that usually means the pod got stuck badly. There is no more data in the actions that I can see. @jsolovjo ?

It seems totally unrelated to Paul's problems. The tests are able to be built correctly

From what I can see, it seems that it fails when starting webpack. I remember there were some changes related to that, does it ring any bells @grgrzybek ?

image
image

jsolovjo:
I don't have much expertise here. Whether a webpack is failing, or it needs more time as the timeout is set to 30 seconds or the expected message is different when checking for its startup. I will try to check further.

jsolovjo:
Ok, it seems the issue might be here - https://github.com/hawtio/hawtio-next/blob/main/.github/workflows/e2e-test.yml#L77

The test text is `You can now view app in the browser.` which is not there as you can/can't see from the logs below.

I tried to follow the steps from our e2e workflow file and when I start the app, I am getting this:
```
<i> [webpack-dev-server] Project is running at:
<i> [webpack-dev-server] Loopback: http://localhost:3000/, http://[::1]:3000/
<i> [webpack-dev-server] On Your Network (IPv4): http://192.168.0.186:3000/
<i> [webpack-dev-server] Content not from webpack is served from '/Users/.../hawtio-next/app/public' directory
<i> [webpack-dev-server] 404s will fallback to '/hawtio'
```

So it seems the test message has changed for some reason.
What do you think @grgrzybek ?

grgrzybek:
sorry I didn't check the outcome. I'll get back to this on Monday.

app/src/bootstrap.tsx
```diff
@@ -1,29 +1,58 @@
...
 
-// Bootstrap Hawtio
-registerPlugins()
+hawtio.addUrl('plugin')
 
-// You can also select which builtin plugins to load
-//connect()
-//jmx()
-//camel()
+import('@hawtio/react').then(async m => {
```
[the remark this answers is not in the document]
grgrzybek: Actually it's not needed, because there's no `await`. Changing this.

app/src/bootstrap.tsx
```diff
@@ -1,29 +1,58 @@
...
+  //jmx()
+  //camel()
 
-const root = ReactDOM.createRoot(document.getElementById('root') as HTMLElement)
-root.render(
-  <React.StrictMode>
-    <Hawtio />
-  </React.StrictMode>,
-)
+  import('./examples').then(m => {
```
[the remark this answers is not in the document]
grgrzybek: hmm, you're right... Looks like bootstrap has to be delayed...

tadayosi:
@grgrzybek Is it intentional that it still includes the loading checklist at the start, or is it a leftover for debugging?

grgrzybek:
> @grgrzybek Is it intentional that it still includes the loading checklist at the start, or is it a leftover for debugging?

I thought it'd look professional ;)

But I'm thinking about adding something like:
```
<HawtioInitialization verbose={true} />
```

and by default (when verbose=false) we could see some spinner.

grgrzybek:
I've already implemented `<HawtioInitialization verbose={false} />` and it simply shows dotted progress bar instead of full information.'
Will show you later ;)

packages/hawtio/package.json
```diff
@@ -1,9 +1,28 @@
...
+      "types": "./dist/index.d.ts",
+      "require": "./dist/index.js",
+      "default": "./dist/index.js"
+    },
+    "./init": {
+      "types": "./dist/init.d.ts",
+      "require": "./dist/init.js",
+      "default": "./dist/init.js"
+    },
+    "./ui": {
```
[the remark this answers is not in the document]
grgrzybek: It makes sense from the point of view of Webpack.
This allows to do things like:

```javascript
import '@hawtio/react/init`
...
import('@hawtio/react').then(() => {
  // ...
})
```

This is handled correctly by webpack and the web application loads/evaluates the _init_ part of Hawtio synchronously (from smaller .js files) and then heavier processing of bigger _chunks_ is done asynchronously.

packages/hawtio/src/plugins/shared/jolokia-service.ts
grgrzybek: It was my silly way to get rid `Promise returned from request is ignored` warnings from IDEA.
But right - I think warning is less confusing than this `.then()`.

packages/hawtio/src/core/config-manager.ts
```diff
@@ -140,11 +141,203 @@ export type OnlineConfig = {
...
+  started,
+  skipped,
+  finished,
+  error,
+}
+export type InitializationTasks = Record<string, { ready: TaskState; group: string }>
+
+/** Supported authentication methods */
+export type AuthenticationKind =
+  /**
```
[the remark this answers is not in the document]
tadayosi: Alright, let's keep it as is then. Thanks for letting me know that!

PR #1582, opened 2025-06-25
Debug page

phantomjinx:
In relation to comments posted at https://github.com/hawtio/hawtio-next/pull/1575.

Fix: Get Debug diagram to display
    
    * Due to hardcoded height values being removed, the
      #route-diagram-breakpoint-view <div> lacked a parent with a height value
      to inherit. Therefore, a value had to be passed down the hierarchy from
      the nearest ancestor with an established height, namely #camel-content-main
    
    * So using CSS rules, define the height of all the containers in the
      hierarchy using flex / flex-grow
    
    * MessageDrawer
     * Gives the elements id attributes to enable easier debugging / inspection
    
    * Exchanges
     * Since panels have been tasked with growing, this has the drawback of
       other instances of <panel> also growing in other pages. Therefore, a
       large gap is generated between Inflight Exchanges panel and the empty
       Blocked Exchanges <card>.
     * However, the UI is inconsistent in its use of a <card> for empty instances
       and a <panel> for populated instances. Therefore, by making both
       components use <card>, this makes the UI consistent and avoids the gap
       generated by using a <panel>.


phantomjinx:
I'll do a video to demo the changes but feel free to have a review of the code now if you wish.

tadayosi (approved):

joshiraezcode (approved):
I would have to double check something with flex-grow because it's implied to be 1 by default, and only comes to play if there are other siblings. It might be needed at the top parent component but I see it being used down a nested list of single enclosing components and I'm not sure if that's needed. I would need to investigate a bit on it.

That said, it seems this fixes main as the E2E tests are green again (not very helpful as we still have to sit down and check if there are false positives but still) and I'm sure you would have check that the diagram is showing, so for me let's go ahead and merge as the CSS is enclosed enough to always be checked out later nor is a very important part of our page right now so for me let's go ahead and merge it.

And I'm not confident on my take either so yeah, if it works let's go ahead and merge it for now.

joshiraezcode:
image

phantomjinx:
> I can't watch the video. Downloading doesn't allow me to watch it either.
> 

How odd? Do you not have the H.265 codec installed?

Opening video decoder: [ffmpeg] FFmpeg's libavcodec codec family
libavcodec version 61.19.101 (external)
Selected video codec: [ffhevc] vfm: ffmpeg (FFmpeg HEVC / H.265)

PR #1619, opened 2025-07-15
Dev peer dependencies

grgrzybek:
This may look scary, but I took much care to organize the dependencies correctly. Read the details in: https://github.com/hawtio/hawtio-next/blob/dev-peer-dependencies/app/README.md


app/README.md
```diff
@@ -8,12 +8,92 @@ This is a JavaScript application which uses (in `dependencies` of its `package.j
...
 
 - should be structured
 - should use `webpack` and configure [Module Federation](https://webpack.js.org/concepts/module-federation/)
 - should initialize and bootstrap `hawtio` classes
-- should render initial React application
+- should render main React component (`<Hawtio>`)
+
+## dependencies, dev dependencies, peer dependencies
+
+This `app` NPM package is bundled using Webpack and produces an _application_ that runs in browser. This is _not_ a published NPM packages and will never be a dependency of any other NPM project.
```
phantomjinx, line 23: - in browser ---> in the browser
- packages  ---> package

app/README.md
```diff
@@ -8,12 +8,92 @@ This is a JavaScript application which uses (in `dependencies` of its `package.j
...
+
+In terms of dependencies we have to remember few things:
```
phantomjinx, line 25: - few things ---> a few things

app/README.md
```diff
@@ -8,12 +8,92 @@ This is a JavaScript application which uses (in `dependencies` of its `package.j
...
+
+- This `app` definitely depends on `react`, `react-dom` and Patternfly libraries and `@hawtio/react` package
```
phantomjinx, line 27: and Patternfly ---> , Patternfly

app/README.md
```diff
@@ -8,12 +8,92 @@ This is a JavaScript application which uses (in `dependencies` of its `package.j
...
+- This `app` transitively (implicitly) depends on other libraries (dependencies of the above)
+- `webpack.config.cjs` configuration of this `app` uses some _shared modules_ (with the help of [Module Federation mechanism](https://webpack.js.org/concepts/module-federation/)).
+
+Application bundled with Webpack should not have any `peerDependencies`, because it's not used by other packages in their `dependencies`.
```
phantomjinx, line 31: Application ---> Applications
it's ---> it is (not high priority but always taught not to use such abbreviations in offical docs)

app/README.md
```diff
@@ -8,12 +8,92 @@ This is a JavaScript application which uses (in `dependencies` of its `package.j
...
+
+This `app` should declare a `dependency` only if it's used directly in its code (for example `src/bootstrap.tsx` imports React functions, so `react` should be a direct dependency). This `app` should not add Jolokia to it's `dependencies` - Webpack will bundle it together with `@hawtio/react` without problems.
```
phantomjinx, line 33: it's ---> it is

app/README.md
```diff
@@ -8,12 +8,92 @@ This is a JavaScript application which uses (in `dependencies` of its `package.j
...
+
+If there's a need to select specific version of a transitive package (for example due to security reasons or known CVE), such dependency should be added to `resolutions` config in `package.json`.
```
phantomjinx, line 35: - specific version ---> a specific version
- such dependency ---> such a dependency
- `resolutions` config ---> the `resolutions` property

app/README.md
```diff
@@ -8,12 +8,92 @@ This is a JavaScript application which uses (in `dependencies` of its `package.j
...
+
+`@patternfly/react-core` and React libraries should be _peer dependencies_ of `@hawtio/react` package:
+
+- these are declared as `shared` dependencies in Module Federation configuration in `webpack.config.cjs`
+- the exact version should be specified by the _application_, not by the _library_
+- another plugin library (for example the plugin from [Artemis Console](https://github.com/apache/activemq-artemis-console/)) could be used by this `app`, so it's easier to manage single version of React, Patternfly at the _application_ level than at the level of both `@hawtio/react` and `artemis-console-extension`.
```
phantomjinx, line 41: - it's ---> it is
- single version ---> single versions
- React, Patternfly ---> React and Patternfly

app/README.md
```diff
@@ -8,12 +8,92 @@ This is a JavaScript application which uses (in `dependencies` of its `package.j
...
+
+And finally there's one confusing, but recommended configuration - a _peer dependency_ usually should be added also as _dev dependency_. There's only one reason - some package managers like [Yarn](https://yarnpkg.com/) or NPM before version 7 do **not** install peer dependencies. As a proof of this confusing convention, here's a fragment from `react-router` package:
```
phantomjinx, line 43: - there's ---> there is
- also as ---> also as a
- There's ---> There is
- here's ---> here is
- fragment from ---> fragment from the

app/README.md
```diff
@@ -8,12 +8,92 @@ This is a JavaScript application which uses (in `dependencies` of its `package.j
...
+"peerDependencies": {
+  "react": ">=16.8"
+},
+```
+
+## Patternfly dependencies
+
+The first dependency on Patternfly library is `@patternfly/react-core`. It directly depends on (declaration in its `dependencies`) 3 other libraries: `@patternfly/react-icons`, `@patternfly/react-styles` and `@patternfly/react-tokens`.
+
+There other Patternfly libraries we use: `@patternfly/react-charts`, `@patternfly/react-code-editor` and `@patternfly/react-table`, which also depend on icons, styles and tokens.
```
phantomjinx, line 59: There ---> The

app/README.md
```diff
@@ -8,12 +8,92 @@ This is a JavaScript application which uses (in `dependencies` of its `package.j
...
+
+## A note about semantic versioning
+
+For Yarn, `defaultSemverRangePrefix` option defaults to `"^"`. When adding dependencies you can also use one of
+
+    yarn add <dependency> --caret
+    yarn add <dependency> --tilde
+    yarn add <dependency> --exact
+
+It may be confusing to distinguish between caret and tilde versioning, so here's the definition:
```
phantomjinx, line 77: - here's ---> here is

app/README.md
```diff
@@ -8,12 +8,92 @@ This is a JavaScript application which uses (in `dependencies` of its `package.j
...
+
+- `"^"` preserves left-most non-zero digit
```
phantomjinx, line 79: - left-most ---> the left-most

packages/hawtio/jest.config.ts
```diff
@@ -1,41 +1,59 @@
 import type { Config } from 'jest'
-import * as path from 'path'
 
 const config: Config = {
+  // https://jestjs.io/docs/ecmascript-modules
```
phantomjinx, line 4: Should this use multiline comment /* */ style instead?

phantomjinx (changes requested):
Nothing major - just syntax corrections.

app/package.json
```diff
@@ -21,13 +22,10 @@
     "copy-webpack-plugin": "^13.0.0",
     "css-loader": "^7.1.2",
     "html-webpack-plugin": "^5.6.3",
-    "jest": "^30.0.3",
-    "monaco-editor": "^0.52.2",
     "monaco-editor-webpack-plugin": "^7.1.0",
     "source-map-loader": "^5.0.0",
     "style-loader": "^4.0.0",
     "swc-loader": "^0.2.6",
-    "typescript": "^5.4.5",
```
tadayosi, line 30: why no typescript?

app/src/setupTests.ts
tadayosi, line 1: why do we remove this? maybe it's still beneficial to put some test code in app so that the user knows how they can develop tests for it.

packages/hawtio/package.json
```diff
@@ -52,74 +52,70 @@
     "prepack": "yarn build && yarn replace-version"
   },
   "dependencies": {
-    "@hawtio/camel-model-v4_10": "npm:@hawtio/camel-model@~4.10.2",
-    "@hawtio/camel-model-v4_8": "npm:@hawtio/camel-model@~4.8.5",
+    "@hawtio/camel-model-v4_10": "npm:@hawtio/camel-model@^4.10.6",
+    "@hawtio/camel-model-v4_8": "npm:@hawtio/camel-model@^4.8.8",
```
tadayosi, line 56: These intentionally use `~`, as you can see in the dep names (v4_8, v4_10). Please don't change them.

tadayosi: This part is critical.

app/README.md
grgrzybek: oops - I have to check everywhere then (s/Patternfly/PatternFly)

app/README.md
grgrzybek: true, but if there's a CVE in transitive dependency I think we can use `resolutions` instead of saying "this is our 1st level dependency".
Scanners just scan node_modules/ which may have nested node_modules/ and so on...

packages/hawtio/jest.config.ts
grgrzybek: it was easier for me to do `ctrl+/`...
But sure, let's change it ;)

app/package.json
grgrzybek: swc-loader doesn't use tsc. confirmed.
https://rspack.rs/guide/tech/typescript#transpile-only

> For maximum speed, builtin:swc-loader transpiles TypeScript source code without performing any type checking. An external tool such as tsc must be used for type checking.

app/src/setupTests.ts
grgrzybek: I removed it only because we're not using - there were no `jest` tests in the app (and no jest config).
Also the app is bundled with webpack, so ideally there should be _higher level_ tests. For now the app effectively only contains single `bootstrap.tsx` file which is the _core_ of the application.

I can bring it back when we develop some tests.

grgrzybek: wdyt?

packages/hawtio/package.json
grgrzybek: you're right - thanks for catching! `^4.10.6` would match `4.11.0` too, while `~4.10.6` would match only micro/patch version changes. I documented it, but didn't follow my instructions ;)

app/src/setupTests.ts
tadayosi: ok, let's remove them for now.

PR #1693, opened 2025-09-13
build: Add ESM build alongside CJS for hawtio-next

GChuf:
This is just a working example - right now backend-middleware already builds CJS + ESM in separate folders (main and module).
I used .js and .mjs extensions in the same "dist" folder for hawtio-next. I'm not sure which is best - same folder or separate folders. The maintainers should decide on this anyway.
I also got rid of "main", "typings" and "module" top-level keys in package json and only used "exports".

According to the docs:
https://nodejs.org/api/packages.html#main

This decreases the bundle sizes and should also help with tree-shaking "downstream".

More info about export:
> When a package has an "exports" field, this will take precedence over the "main" field when importing the package by name.
> 
> The "exports" field allows defining the entry points of a package when imported by name loaded either via a node_modules lookup or a self-reference to its own name. It is supported in Node.js 12+ as an alternative to the "main" that can support defining subpath exports and conditional exports while encapsulating internal unexported modules.
> Conditional Exports can also be used within "exports" to define different package entry points per environment, including whether the package is referenced via require or via import.

### DIFF
--- a/packages/backend-middleware/package.json
+++ b/packages/backend-middleware/package.json
@@ -2,9 +2,13 @@
   "name": "@hawtio/backend-middleware",
   "version": "1.0.7",
   "description": "An Express middleware that implements Hawtio backend",
-  "main": "build/main/index.js",
-  "typings": "build/main/index.d.ts",
-  "module": "build/module/index.js",
+  "exports": {
+    ".": {
+      "types": "./build/main/index.d.ts",
+      "require": "./build/main/index.js",
+      "import": "./build/module/index.js"
+    }
+  },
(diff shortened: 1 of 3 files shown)


GChuf:
Looks like I'll have to fix imports.
I was only testing yarn build, but I can see yarn build:all fails.

GChuf:
Sorry for not answering sooner ...
Like usually, I was messing around with artemis console and then I got here.
It seems the changes I've done here in this branch are enough for artemis, but yes - not enough for hawtio-next.
I didn't want to remove this, so I just left it as draft ...

PR #1725, opened 2025-10-06
Proxy fixes

grgrzybek:


packages/hawtio/src/plugins/shared/connect-service.ts
```diff
@@ -388,9 +388,7 @@ class ConnectService implements IConnectService {
         }
         fetch(this.getJolokiaUrl(connection), {
           method: 'post',
-          // with application/json, I'm getting "CanceledError: Request stream has been aborted" when running
-          // via hawtioMiddleware...
-          headers: { ...headers, 'Content-Type': 'text/json' },
+          headers: { ...headers, 'Content-Type': 'application/json' },
```
tadayosi, line 391: doesn't this affect anything with the real jolokia backend and/or proxy servlet?

tadayosi (approved):

PR #1741, opened 2025-10-13
HAWNG-1248: Handles error conditions on Route XML loading failures

phantomjinx:
* xml.ts
  * If an error has occurred in the parsing of the XML, ensure this is thrown by the parse function and not remained hidden in the XML Doc

* RouteDiagram.tsx
  * Better handle XML parsing errors. Rather than just a blank screen, displays an error message conveying the problem in an error in the diagram canvas.

### DIFF
--- a/packages/hawtio/src/plugins/camel/route-diagram/RouteDiagram.tsx
+++ b/packages/hawtio/src/plugins/camel/route-diagram/RouteDiagram.tsx
@@ -1,7 +1,8 @@
 import { eventService } from '@hawtiosrc/core'
 import { MBeanNode } from '@hawtiosrc/plugins/shared'
-import { Switch, Title } from '@patternfly/react-core'
+import { Switch, Title, EmptyState, EmptyStateBody, EmptyStateHeader, EmptyStateIcon } from '@patternfly/react-core'
 import { Table, Tbody, Td, Tr } from '@patternfly/react-table'
+import { ExclamationCircleIcon } from '@patternfly/react-icons'
 import React, { RefObject, useCallback, useContext, useEffect, useMemo, useRef, useState } from 'react'
 import {
   Connection,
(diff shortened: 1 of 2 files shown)


packages/hawtio/src/plugins/camel/route-diagram/RouteDiagram.tsx
```diff
@@ -163,8 +178,27 @@ const ReactFlowRouteDiagram: React.FunctionComponent<ReactFlowRouteDiagramProps>
...
+        <CardBody>
+          <Alert variant='warning' title='Route could not be drawn.' />
+
+          <Alert variant='danger' title={error.message}>
+            {hasCause(error) && <p>Cause: {(error.cause as Error).message}</p>}
+          </Alert>
+        </CardBody>
+      </Card>
+    )
+  }
```
tadayosi: why can't we use the existing `eventService` function to notify errors instead?

tadayosi: Yeah, as far as I remember there are no other places that use a card for error. For consistentcy it's better to use the `eventService` for error notification. We can still keep the warning message if a blank area isn't comfortable, or there's indeed a dedicated component for the empty state.
https://v5-archive.patternfly.org/components/empty-state

jsolovjo:
> Something (not this) broke the e2e tests. Any thought @jsolovjo @grgrzybek ??

On Spring Boot the tracing test is a bit flaky and moreover, it's enabled by default now (not sure why it's changed now), so the test tries to enable it but it's already enabled. In my new PR in the hawtio project I changed the test behaviour.

Regarding `Account Lockout (Throttling).User account gets locked out after multiple failed login attempts` error - it's another flaky story. If I am not mistaken, @kariuwu worked on this, so perhaps she knows more about it.

PR #1759, opened 2025-10-20
fix(plugins): Skip /plugin endpoint if not available (fixes hawtio/hawtio#4014)

grgrzybek:


grgrzybek:
merging without tests, because npm is down.

PR #1792, opened 2025-11-13
fix(jolokia): reject the promise on jolokia.request() error instead of resolving with empty value (fixes #1791)

### LINKED ISSUE #1791: jolokia-service.ts methods should NOT resolve the promises with empty objects
When checking:
* apache/activemq-artemis-console#134

we started seeing errors like this in the browser console:
```
other.856fe495.chunk.js:2 [hawtio-shared] Error during readAttributes: {request: {…}, error_type: 'java.lang.Exception', error: 'java.lang.Exception : User not authorized to access attribute: Uptime', status: 403}

react.52f8e43c.chunk.js:2 Uncaught TypeError: Cannot read properties of undefined (reading 'indexOf')
```

After doing some research, I found that the change that caused this problem is:
* jolokia/jolokia#893

Before this feature, Jolokia, when asked to "get all the attributes of an MBean" was listing all the attributes from an MBeanInfo and iterating over them and was calling: `javax.management.MBeanServerConnection#getAttribute()` for each of them (for example there's ~90 atributes for Artemis Broker MBean).

The problem is (was) that when there was an error (for example caused by RBAC interceptor from Artemis), the error message was added as the value of the attribute! In practice, when user saw an object similar to:
```json
{
  "Name": "User xxx can't access property \"Name\"",
  "Id": "User xxx can't access property \"Id\"",
...
}
```

it was obvious that there's some security issue, but in theory it's not a good design.

jolokia/jolokia#893 changed Jolokia's Read Handler so `javax.management.MBeanServerConnection#getAttributes()` (plural version) is called and potential SecurityException is thrown immediately and proper Jolokia error is returned:
```json
{
  "request": {
    "mbean": "org.apache.activemq.artemis:broker=\"0.0.0.0\",component=acceptors,name=\"artemis\"",
    "type": "read"
  },
  "error_type": "java.lang.Exception",
  "error": "java.lang.Exception : User not authorized to access attribute: Name",
  "status": 403
}
```

The problem is that Hawtio (jolokia-service.ts) is using this code:
https://github.com/hawtio/hawtio-next/blob/74f06ef1d75d87b662340f4c6da0c5a98eb31fc2/packages/hawtio/src/plugins/shared/jolokia-service.ts#L816-L835

when a Jolokia exception is handled in the `.catch()` of the promise, the higher promise is simply resolved with `{}`...

@tadayosi I think (and we've checked it's much better from the perspective of entire Hawtio application) to simply reject the promise with the Jolokia error, so we can handle it like this:

Image

WDYT?

grgrzybek:
Pretty please check, ask, understand and object, because it worked for me, fixed https://issues.apache.org/jira/browse/ARTEMIS-5760 but I may have missed something...

### DIFF
--- a/packages/hawtio/src/plugins/shared/jolokia-service.ts
+++ b/packages/hawtio/src/plugins/shared/jolokia-service.ts
@@ -826,7 +826,7 @@ class JolokiaService implements IJolokiaService {
           },
           error => {
             log.error('Error during readAttributes:', error)
-            resolve({})
+            reject(error)
           },
           options,
         ),
@@ -856,7 +856,7 @@ class JolokiaService implements IJolokiaService {
           },
           error => {
             log.error('Error during readAttributes:', error)
-            resolve({})
+            reject(error)
           },
           options,
         ),
@@ -882,7 +882,7 @@ class JolokiaService implements IJolokiaService {
           },
           error => {
             log.error('Error during readAttribute:', error)
-            resolve(null)
+            reject(error)
           },
           options,
         ),
@@ -909,7 +909,7 @@ class JolokiaService implements IJolokiaService {
           },
           error => {
             log.error('Error during writeAttribute:', error)
-            resolve(null)
+            reject(error)
           },
           options,
         ),


tadayosi:
@grgrzybek Looks ok, but have you checked that all the callers of the changed methods are ready and can safely handle the promise rejected errors?  In general, with the changes the callers should try/catch the promise and send the error to the eventService's error notification.

tadayosi:
The `readAttribute(s)` methods are used in many places, so the changes can cause unexpected issues in other plugins. We need to test it entirely especially where RBAC takes effect.
Note that Artemis is not the only target that RBAC is effective, Hawtio Online also supports RBAC, so testing is very important for Hawtio Online, which our E2E doesn't cover automatically.

grgrzybek:
I did the 1.11.3 release, but I'll check the code more now. (Which is very unprofessional).

PR #1826, opened 2025-12-01
feat(auth): Support multiple OIDC login methods (fixes #1581)

### LINKED ISSUE #1581: Support multiple OIDC configurations
With #1556 we support more authentication methods (but only one per type). It'd be great to support more OIDC providers configured (e.g., Keycloak, EntraID and Spring Authorization Server).

grgrzybek:
Generally the change is that `oidc-service.ts` supports an array of configs and metadata instead of a single config and metadata.

When multiple OIDC configs are available, we present this UI:
image

and each OIDC login method has an ID which selects proper provide, login and logout methods.

### DIFF
--- a/packages/hawtio/src/plugins/auth/oidc/oidc-service.ts
+++ b/packages/hawtio/src/plugins/auth/oidc/oidc-service.ts
@@ -99,13 +99,18 @@ type AuthData = {
 
 class OidcService implements IOidcService {
   // promises created during construction
-  // configuration from Hawtio backend - should include provider URL
-  private readonly config: Promise<OidcConfig | null>
-  // OIDC plugin is enabled when there's provider URL in the config - may not be reachable though
+
+  // OIDC plugin is enabled when there's any configuration with a provider URL in the config - may not be reachable though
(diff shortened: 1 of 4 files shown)


packages/hawtio/src/plugins/auth/oidc/oidc-service.ts
```diff
@@ -425,26 +473,47 @@ class OidcService implements IOidcService {
     }
   }
 
+  private oidcLoginFunction = (oidcIndex: number): (() => Promise<AuthenticationResult>) => {
+    log.info('GRGR', `Creating login function ${oidcIndex}`)
+    return () => {
+      log.info('GRGR', `Invoking login function ${oidcIndex}`)
```
tadayosi: GRGR?
How about making the label more generic and changing the level to debug so that it can be kept in the codebase?

grgrzybek: oh my - I thought I removed these!!!! Sorry and thanks for your sharp eye ;)

that's my way to later find given code in browser JS debugger ;)

PR #1905, opened 2026-02-05
Fixes for PF6 taking into account theming in dark mode

phantomjinx:
### Fix: background of page should conform to theme of browser
- In PF6, vars ending in numbers often map to specific levels of the color palette or legacy mappings. In many contexts, 100 resolves to white (or near-white) and stays that way regardless of the theme.
- Changes to a semantic token, which tells PF to give the main background color according to the browser theme:
  - Light Mode: Resolves to White (#ffffff)
  - Dark Mode: Resolves to the dark background color (e.g., #151515)

### Fix: Theme changing failure with Patternfly 6
- core.ts
  - The window listener is tasked with adding the PF dark CSS property to hawtio's document element. This was hard-coded to the PF5 version so does not function with PF6
  - Makes use of build-time constant that construct the correct PF CSS property to ensure that change of theme functions correctly
    
- tsup.config.ts
  - Converts defineConfig to a function to allow for build-time scripting
  - Resolves the build-time location of the patternfly package.json and extracts its version then injecting this into a process.env property for use at runtime by the hawtio core.
  - Build-time
    - Detect version of PF, eg. 6 and inject into node constant
  - Runtime
    - HawtioCore registers window listener, detecting browser theme
    - If theme is dark then correct version of CSS theme class is appended to document element

### DIFF
--- a/packages/hawtio/src/ui/page/HawtioPage.css
+++ b/packages/hawtio/src/ui/page/HawtioPage.css
@@ -1,3 +1,3 @@
 #hawtio-main-page {
-  background-color: var(--pf-t--global--background--color--100);
+  background-color: var(--pf-t--global--background--color--primary--default);
 }


tadayosi (commented):
That's great, but can we simply hardcode the change to `v6`?

It doesn't make sense to introduce this tsup config change only for this single part, where there are many other styles that also refer to `v6`.

Ultimately such styles should be updated/migrated manually whenever upgrading to a new PF version (which shouldn't occur so often).

phantomjinx:
> That's great, but can we simply hardcode the change to `v6`?
> 
> It doesn't make sense to introduce this tsup config change only for this single part, where there are many other styles that also refer to `v6`.
> 
> Ultimately such styles should be updated/migrated manually whenever upgrading to a new PF version (which shouldn't occur so often).

The point is that it was overlooked for PF6 despite the update to all the other styles so I am mitigating this by doing it automatically as/when PF7 comes along.

grgrzybek (approved):

PR #1906, opened 2026-02-06
[E2E] Fix the e2e workflow file.

jsolovjo:
This PR fixes an issue where incorrect syntax in the workflow file caused test app images to be selected incorrectly during github actions.

It should definitely help to pass checks in this PR - https://github.com/hawtio/hawtio-react/pull/1905 - as checking the logs I noticed that 4.x images were used for testing (with older PF version) instead of newer 5.x images.

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -96,16 +96,16 @@ jobs:
         env:
           body: ${{ github.event.pull_request.body || ''}}
         run: |
-          if [[ "${body}" =~ \`branch:[[:space:]]([[:alnum:]\/_.-]+)(:([[:alnum:]\/_.-]+))?\` ]]; then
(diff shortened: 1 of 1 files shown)


grgrzybek (approved):

tadayosi (approved):

jsolovjo:
Hi @tadayosi yes, it seems it uses the old workflow in this PR for a security reason. I can try it in my forked project in a separate branch.

jsolovjo:
Hi @tadayosi 
I tested it in my forked project with a test branch

- The run - https://github.com/jsolovjo/hawtio-next/actions/runs/21860557050
- The commit which was tested - https://github.com/jsolovjo/hawtio-next/commit/b3d5143b3c088485672a49a542f65a3453dae1f4 (I had to add my test branch there to trigger the actions and check how `if` statement works there - so based in the condition it picked 5.x images correctly now)

And in the given PR it fails because of security reasons - we don't run the modified workflows in the PRs.

tadayosi:
@jsolovjo Thanks. That makes sense. Let's merge it then.

PR #1917, opened 2026-02-10
fix(ui): adjust loading page layout

tadayosi:


jsolovjo:
Interesting, it see that the **old** e2e workflow was used in the actions 
It still has the issue with brackets in if [[ "${body}" =~ \`branch:[[:space:]]([[:alnum:]\/_.-]+)(:([[:alnum:]\/_.-]+))?\` ]]; then and missing an extra step I added.
However, I see that the e2e workflow is updated in 2.x and in your branch.

Issue #903, opened 2024-04-24
Navigation doesn't work properly in the plugins involves Tree


mmelko:
This might be caused by resolution of the path, where selected mbeans aren't included into the path, so i.e source of the route1 is translated to `[host]/hawtio/camel/source`. So when we paste this link to the browser we are automatically redirected to the root of the camel plugin with the context table. 

Similar behaviour when we selected different mbean of the same type. Instead of perceive the selected tab it's automatically redirected to the first tab.

tadayosi:
@mmelko Look at how Hawtio v2 did for the persistent navigation. I remember it used query parameters such as `?nid=` for keeping the current mbean location. That said, I don't think using query params is a good design in 2024. So the first thing we should do would be to design how we encode the path location into URL. Do you have any ideas?

tadayosi:
I wouldn't say it's a bug so let me change it to enhancement.

mmelko:
Sure, I marked is a bug because the functionality is lacking compared to v2. 

My idea from top of my head is to use path param and mbean names (i.e `[host]/hawtio/camel/source/routes`) That would require extending either current context or create new with something like searchNode and then obtain the objectName.

tadayosi:
@mmelko Hmm, then how do we encode the selected node into the path?  Should we still use a query parameter for the node selection?

mmelko:
so bad example .. routes would be the name of the mbean (not the object name). so for route1 the path would look like `[host]/hawtio/camel/source/route1`  That's how I'd like to see it.

tadayosi:
@joshiraezcode I don't think so. You are free to investigate it but be aware that it's really a complicated one to implement correctly. I'd propose to first investigate it and write some solution plan for review before working on an actual fix.

pseudaverse:
> [@mmelko](https://github.com/mmelko) Look at how Hawtio v2 did for the persistent navigation. I remember it used query parameters such as `?nid=` for keeping the current mbean location. That said, I don't think using query params is a good design in 2024. So the first thing we should do would be to design how we encode the path location into URL. Do you have any ideas?

@tadayosi This thing was quite useful in hawtio v2, with it and  "con=" query param we could share link to exact mbean/etc.. I guess the only way to do so is to add some state/location storing around JmxTreeView, and nid query param looks like a straightforward solution in this case. Why do you think its a bad idea? I would implement this logic (maybe firstly under some config param in hawtconfig.json/whatever), what do you think?

[...]

PR #1939, opened 2026-02-18
fix(jmx): refactor and align nid usage on JMX tree

tadayosi:
Relates #903

Addresses the following:

1. Use the builtin node id for `nid` instead of introducing a new one
2. No double URL encoding on `nid` parameter
3. More robust detection of previously selected nodes for complicated object names such as `org.xnio:type=Xnio,provider="nio",worker="XNIO-1",address="/[0:0:0:0:0:0:0:0]:10000"`

### DIFF
--- a/packages/hawtio/src/plugins/jmx/context.ts
+++ b/packages/hawtio/src/plugins/jmx/context.ts
@@ -2,7 +2,7 @@ import { EVENT_REFRESH, eventService } from '@hawtiosrc/core'
 import { PluginNodeSelectionContext } from '@hawtiosrc/plugins'
 import { MBeanNode, MBeanTree, workspace } from '@hawtiosrc/plugins/shared'
 import { createContext, useContext, useEffect, useRef, useState } from 'react'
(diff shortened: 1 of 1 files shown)


tadayosi:
@grgrzybek Please review

packages/hawtio/src/plugins/jmx/context.ts
```diff
@@ -110,6 +84,17 @@ export function useMBeanTree() {
   return { tree, loaded, selectedNode, setSelectedNode }
 }
 
+/**
+ * Build URL query string with nid parameter, preserving other existing params
+ * @param node - The node to encode
+ */
+export function pluginPathWithNodeId(node: MBeanNode): string {
+  const searchParams = new URLSearchParams(window.location.search)
```
grgrzybek: I don't know enough about react-router - shouldn't we use the state from `useSearchParams()` hook of `react-router-dom` here?

tadayosi: good point. let me see if we can use `useSearchParams` here instead.

PR #1954, opened 2026-02-19
fix(jmx): Parse MBean names using parser, not RegExp (fixes #1952)

### LINKED ISSUE #1952: Quoted MBean names are not split correctly
hawtio/hawtio-integration@f356f788 introduced a RegExp to parse MBean names.
However the rules are more complex and this applies: https://stackoverflow.com/a/1732454

When MBean name uses quoted values, the rules are slightly different than for unquoted values.

I have a fix using a parser instead of magic RegExp.

grgrzybek:
(cherry picked from commit bde75a5e7651c2588d37bb85019bf6dd67aeca3e)

### DIFF
--- a/packages/hawtio/src/plugins/shared/tree/node.ts
+++ b/packages/hawtio/src/plugins/shared/tree/node.ts
@@ -806,8 +806,14 @@ export class PropertyList {
 
   match(properties: Record<string, string>): boolean {
     return Object.entries(properties).every(([key, value]) => {
+      // match by unquoted/unescaped values
       const thisIndex = this.paths.findIndex(e => e.key === key)
-      return thisIndex >= 0 && matchWithWildcard(this.paths[thisIndex]!.value, value)
+      if (thisIndex >= 0 && matchWithWildcard(this.paths[thisIndex]!.value, value)) {
+        return true
+      }
+      // match by quoted/escaped values
+      const thisValue = this.properties[key]
+      return thisValue && matchWithWildcard(thisValue, value)
     })
   }
 
--- a/packages/hawtio/src/plugins/shared/tree/node.ts
+++ b/packages/hawtio/src/plugins/shared/tree/node.ts
@@ -1,6 +1,6 @@
-import { escapeHtmlId, escapeTags } from '@hawtiosrc/util/htmls'
+import { escapeHtmlId } from '@hawtiosrc/util/htmls'
 import { isEmpty } from '@hawtiosrc/util/objects'
-import { matchWithWildcard, stringSorter, trimQuotes } from '@hawtiosrc/util/strings'
+import { matchWithWildcard, stringSorter } from '@hawtiosrc/util/strings'
 import { type TreeViewDataItem } from '@patternfly/react-core'
 import { CubeIcon } from '@patternfly/react-icons/dist/esm/icons/cube-icon'
 import { FolderIcon } from '@patternfly/react-icons/dist/esm/icons/folder-icon'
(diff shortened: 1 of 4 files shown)


packages/hawtio/src/plugins/shared/tree/node.ts
```diff
@@ -598,6 +598,145 @@ export class MBeanNode implements TreeViewDataItem {
...
+  const colon = name.indexOf(':')
+  if (colon === -1) {
+    throw new Error('Illegal ObjectName')
+  }
+
+  // there's no doubt here
+  const domain = name.substring(0, colon)
+  if (domain.includes('\n')) {
+    throw new Error('Illegal Object name - domain should not include new line characters')
+  }
```
tadayosi: do we really need this? (surprise)

packages/hawtio/src/plugins/shared/tree/node.ts
```diff
@@ -619,53 +753,39 @@ export class PropertyList {
...
         case 'type':
-          this.typeName = propValue
-          if (this.domain.findChildren(propValue).length > 0) {
-            // if the type name value already exists in the root node
-            // of the domain then let's move this property around too
-            index = 0
-          } else if (this.properties['name']) {
-            // else if the name key already exists, insert the type key before it
-            index = this.paths.findIndex(p => p.key === 'name')
+          this.typeName = parsedName.properties[kv.key]
```
tadayosi: so, is it the fix that the prop values are ensured to be always quoted instead of using unquoted ones?

tadayosi (approved):
Great!  Comments are primarily for confirmation.

packages/hawtio/src/plugins/shared/tree/node.ts
```diff
@@ -598,6 +598,145 @@ export class MBeanNode implements TreeViewDataItem {
   }
 }
 
+export type MBeanName = {
+  domain?: string
+  // unordered map of key-values directly from javax.management.ObjectName#getKeyPropertyList()
+  // the values, when quoted are kept quoted (see javax.management.ObjectName#quote())
+  properties: Record<string, string>
+  // ordered (sometimes reordered by Hawtio) array of key-value pairs, where values are always unquoted
+  // (for display purpose)
```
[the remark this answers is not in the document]
grgrzybek: but you're right - I did what I wanted for 1.12.x, but for 2.x I'll check with Camel view if the navigation works.

packages/hawtio/src/plugins/shared/tree/node.ts
```diff
@@ -619,53 +753,39 @@ export class PropertyList {
...
-      const [propKey, propValue] = this.parseProperty(match[0])
-      this.properties[propKey] = propValue
+    const parsedName = parseMBeanPropertyList(propList)
+    this.properties = parsedName.properties
+    for (const kv of parsedName.paths) {
       let index = -1
-      const lowerKey = propKey.toLowerCase()
-      const path = { key: lowerKey, value: propValue }
-      switch (lowerKey) {
+      switch (kv.key) {
```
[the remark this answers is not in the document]
grgrzybek: `javax.management.ObjectName#construct(java.lang.String, java.util.Map<java.lang.String,java.lang.String>)` uses case-sensitive map, so I dropped the toLowerCase.

And this works:
image

so I needed to keep it like this.

packages/hawtio/src/plugins/shared/tree/node.ts
grgrzybek: here yes - original value (quoted if was quoted initially)

tadayosi (approved):
we can merge it if you think it's ready

tadayosi:
ok, I'm changing 2.x to main, so let's merge this first.

PR #1973, opened 2026-03-03
Added help.md to Diagnostics plugin

### LINKED ISSUE #1852: Diagnostics - Write help.md
I noticed that currently `help.md` for the plugin is empty. It should have some explanation and instructions on the plugin.

joshiraezcode:
Fix #1852

### DIFF
--- a/packages/hawtio/src/plugins/jvm-diagnostics/help.md
+++ b/packages/hawtio/src/plugins/jvm-diagnostics/help.md
@@ -1,0 +1,10 @@
+## JVM Diagnostics
+
+The **JVM Diagnostics plugin** lets you interact with the JVM to get information
+
+### Flight Recorder
+
+The Java Flight Recorder (JFR) tool interacts with the JFR MBean in the JVM to create recordings of your JVM.
+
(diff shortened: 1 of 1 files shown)


tadayosi (approved):
Looks ok, but when looking at the help from the user's point of view, I don't see the relationship between the words "Diagnostics" and "JFR". Is Diagnostics the name of the plugin and JFR a sub menu of it?  There's also the word "Java Flight Recorder (JFR) plugin". Maybe we can revisit the text after merging it.

tadayosi:
I fully understand your argument. It was only the term "JFR plugin" that caught my attention.

I agree that the name "Diagnostics" plugin is somewhat confusing. I'd propose renaming it as "JVM Diagnostics" plugin and putting JFR and other JVM features under it. And let's not use the term "JFR plugin" in the help. WDYT?

Also, 

> The two features to be ported from Hawtio remaining are
>
> - Class histogram
> - JVM flags

Let's file it as a new issue so that we know there is a plan for the future additions.

tadayosi:
@joshiraezcode Ah is this outdated? Can we close it?

tadayosi (approved):

PR #2044, opened 2026-04-14
fix(ui): Filter MBeanNodes with proper object cloning (fixes #2041)

### LINKED ISSUE #2041: Filtered JMX tree loses MBeanNode prototype
1e26b5c3e2975881e7da801e2e0da66c47948290 changes the filtering code.

We've found strange Typescript quirk - `Object.assign()` method has this signature (in `node_modules/typescript/lib/lib.es2015.core.d.ts`):
```javascript
assign<T extends {}, U>(target: T, source: U): T & U;
```

it satisfies the `tsc` with `MBeanNode` type giving this signature:
```javascript
assign<{}, MBeanNode>(target: {}, source: MBeanNode): MBeanNode
```

However https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign says:
> The Object.assign() static method copies all [enumerable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/propertyIsEnumerable) [own properties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn) from one or more source objects to a target object. It returns the modified target object.

Which effectively means that the prototype of `MBeanNode` class (its methods) are NOT available in the returned object and it leads to hard to find bugs - when the selected node after filtering is used to call methods like `getChildren()`.

grgrzybek:
(cherry picked from commit 5e80ae2a6ae9c29b6428f0e4a465c23c34235861)

### DIFF
--- a/packages/hawtio/src/plugins/shared/tree/tree.ts
+++ b/packages/hawtio/src/plugins/shared/tree/tree.ts
@@ -45,7 +45,7 @@ export class MBeanTree {
         const resultsInSubtree = MBeanTree.filter(parentNode.children || [], filter)
 
         if (resultsInSubtree.length !== 0) {
-          const parentNodeCloned = Object.assign({}, parentNode)
+          const parentNodeCloned = Object.assign(Object.create(Object.getPrototypeOf(parentNode)), parentNode)
           parentNodeCloned.children = resultsInSubtree
 
           results = results.concat(parentNodeCloned)


tadayosi (approved):

tadayosi:
@grgrzybek so do we need 2.0.2?

grgrzybek:
@tadayosi eventually yes - this causes problems in Artemis Console, but for now, I need 1.12.5 for this.

PR #2046, opened 2026-04-15
build: Add build steps involving TS type checking (fixes #2045)

### LINKED ISSUE #2045: Add actual typechecks to the build
Webpack (for `app/package.json`) is using (in our case, because it's extensible) `swc-loader` based on `@swc/core` based on Rust. It [doesn't do type checking](https://github.com/swc-project/pkgs/issues/21).
tsup (for `packages/*/package.json`) is using esbuild which [doesn't do type checking as well](https://tsup.egoist.dev/#what-about-type-checking).

When opening TS files in IDEA, we can see TS errors, because IDEA itself reads `tsconfig.json` files and performs type checking.

The way to do type checking is via `tsc`, though it's slow for actual transpiling TS to JS.
It'd be great to have type checking performed by the _canonical_ TS compiler.

PR is coming.

grgrzybek:


packages/backend-middleware/package.json
```diff
@@ -27,10 +27,13 @@
   ],
   "scripts": {
     "start": "yarn build && node example.mjs",
-    "build": "concurrently \"npm:build:*(!watch)\"",
+    "build": "concurrently \"npm:build:*(!watch|check)\"",
+    "build:check": "yarn build",
```
tadayosi, line 31: maybe you're missing `yarn typecheck &&`?

tadayosi (approved):
one comment but otherwise looks great!

packages/backend-middleware/package.json
grgrzybek: not here, as backend-middleware is already built with `tsc` (not tsup).
So I think it's fine.

grgrzybek: I added `build:check`, so it can be invoked with top-level `yarn workspaces foreach`

PR #2054, opened 2026-04-20
feat(ui): Add option in hawtconfig.json for dark-mode logo (fixes #2053)

### LINKED ISSUE #2053: support "dark mode" variant of appLogoUrl in hawtconfig.json
When testing Patternfly 6 (which, in light mode uses light header - different than in Patternfly 5), I found that the logo may be unreadable.

It'd be good to have two options in `hawtconfig.json`:
 - appLogoUrl - current version
 - appLogoDarkUrl - version of the image to be used in dark mode

cc: @tadayosi

grgrzybek:


app/public/hawtconfig.json
```diff
@@ -2,8 +2,8 @@
   "branding": {
     "appName": "Hawtio Management Console",
     "showAppName": false,
-    "appLogoUrl": "hawtio-logo.svg",
-    "companyLogoUrl": "hawtio-logo.svg",
+    "appLogoUrl": "hawtio-logo-dark.svg",
+    "appLogoDarkUrl": "hawtio-logo.svg",
```
phantomjinx: This reverse logic of name v file colours makes my brain hurt :-)

appLogo is used in light theme so therefore we need the dark version of the logo?
appLogoDarkUrl is used in dark theme so therefore we need the light version of the logo?

phantomjinx (approved):
Looks good but maybe just want to explain the dark / light theme names in conjunction with the logo filenames but not a big issue.

app/public/hawtconfig.json
tadayosi: I agree. I find that using '-dark' for dark mode reduces cognitive load for me :-)

tadayosi (approved):

app/public/hawtconfig.json
grgrzybek: true - I had to stop at this for a while and the worst excuse won - I didn't want to rename existing `hawtio-logo.svg` to `hawtio-logo-light.svg` ;)

I'll fix that.

grgrzybek: How about:
```
"appLogoUrl": "hawtio-logo.svg",
"appLogoDarkModeUrl": "hawtio-logo-darkmode.svg",
```
?

joshiraezcode (approved):

PR #2068, opened 2026-04-28
feat(runtime): Added minutes/seconds for younger apps

### LINKED ISSUE #4199: JVM uptime to show minutes as well
This tab shows days and hours. But for short running apps it would be nice to see minutes as well

Image

joshiraezcode:
Fixes https://github.com/hawtio/hawtio/issues/4199

Added minutes and seconds for younger apps

Captura desde 2026-04-28 12-53-27

It will always show the biggest non-zero representation and the next unit.

### DIFF
--- a/packages/hawtio/src/plugins/runtime/runtime-service.ts
+++ b/packages/hawtio/src/plugins/runtime/runtime-service.ts
@@ -6,13 +6,25 @@ import { eventService } from '@hawtiosrc/core'
 class RuntimeService {
   handlers: number[] = []
 
-  convertMsToDaysHours(ms: number): string {
+  convertMsToHumanReadable(ms: number): string {
     const seconds = Math.floor(ms / 1000)
     const minutes = Math.floor(seconds / 60)
     const hours = Math.floor(minutes / 60)
     const days = Math.floor(hours / 24)
 
-    return `${days} days, ${hours % 24} hours`
+    let toHumanReadable
+
+    if (days != 0) {
+      toHumanReadable = `${days} days, ${hours % 24} hours`
+    } else {
+      if (hours != 0) {
+        toHumanReadable = `${hours} hours, ${minutes % 60} minutes`
+      } else {
+        toHumanReadable = `${minutes} minutes, ${seconds % 60} seconds`
+      }
+    }
+
+    return toHumanReadable
   }
 
   async loadSystemProperties(): Promise<SystemProperty[]> {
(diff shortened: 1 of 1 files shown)


packages/hawtio/src/plugins/runtime/runtime-service.ts
```diff
@@ -6,13 +6,25 @@ import { eventService } from '@hawtiosrc/core'
...
+  convertMsToHumanReadable(ms: number): string {
     const seconds = Math.floor(ms / 1000)
     const minutes = Math.floor(seconds / 60)
     const hours = Math.floor(minutes / 60)
     const days = Math.floor(hours / 24)
 
-    return `${days} days, ${hours % 24} hours`
+    let toHumanReadable
+
+    if (days) {
```
grgrzybek: just one comment - I would avoid this implicit to-boolean conversion and use explicit `days == 0`

grgrzybek (approved):

packages/hawtio/src/plugins/runtime/runtime-service.ts
joshiraezcode: You have a point, let me change it

packages/hawtio/src/plugins/runtime/runtime-service.ts
```diff
@@ -6,13 +6,25 @@ import { eventService } from '@hawtiosrc/core'
...
     const days = Math.floor(hours / 24)
 
-    return `${days} days, ${hours % 24} hours`
+    let toHumanReadable
+
+    if (days != 0) {
+      toHumanReadable = `${days} days, ${hours % 24} hours`
+    } else {
+      if (hours != 0) {
+        toHumanReadable = `${hours % 24} hours, ${minutes % 60} minutes`
```
grgrzybek: Are you sure you need to do `hours % 24`? isn't `hours` already in proper value after `Math.floor(minutes / 60)`?

grgrzybek (approved):

packages/hawtio/src/plugins/runtime/runtime-service.ts
joshiraezcode: You have a point, because of the invariant the they are the higher representation available, none of the % are needed on the higher representation. Minutes % 60 is not needed either below

Regarding your question, yes, but not for this representation. 

If we were to show for example 

1 days, 12 hours, 45 minutes, 20 seconds

The variables would have

days = 1
hours = 12 + 24
minutes = 45 + 2160
seconds = 20 + 132300

Because we make sure that they are the higher representation, we don't need the % operation because the values coincide

grgrzybek (approved):

packages/hawtio/src/plugins/runtime/runtime-service.ts
grgrzybek: thanks - it should be fine now ;)

tadayosi (approved):

PR #2108, opened 2026-05-22
chore(docs): add docs/architecture.md for developers and AI Assistants to make prompting easier (#2109)

grgrzybek:


tadayosi (commented):
that's great. but I was actually thinking to add AGENTS.md
https://agents.md

I remember Camel already has it. what do you think?

tadayosi (commented):
also I'd prefer adding the "Boundaries" section to the ai doc, which tells it *do* and *dont*.

tadayosi:
another consideration is the length of the instructions. the longer it is, the higher it consumes the tokens available for the ai.

grgrzybek:
Sure - I understand all the concerns - I simply 1) never did this, 2) tried to start some AI additions to Hawtio React.

> another consideration is the length of the instructions.

I can try to decrease the size. But I tried to be also as expressive as possible for no ambiguities.

> also I'd prefer adding the "Boundaries" section to the ai doc, which tells it do and dont

that'd be good for `AGENTS.md`, not for the `CONTEXT.md`

> I remember Camel already has it. what do you think?

yes, I checked https://github.com/apache/camel/blob/main/AGENTS.md and it matches my findings - it's about telling agents what to do at the code level. So I decided to go with `CONTEXT.md` to explain some terms for easier prompts in the "ask mode", not "code mode".

WDYT?

tadayosi:
@grgrzybek AGENTS.md is an emerging standard stewarded by Agentic AI Foundation. I believe it covers more editors/tools than another format. If we need only one such file, I think AGENTS.md is better. I don't think we want many similar files for different editors/tools.
https://aaif.io/projects/agents-md/

grgrzybek:
Hmm, let's keep it open.
If I were to add `AGENTS.md` I'd have to include the typical information expected from this file (commit style, code formatting rules, ...) but I wanted to avoid it.

I don't have further ideas about how to make it good for now ;)

tadayosi:
@grgrzybek On Boundaries
https://www.agentpatterns.ai/instructions/agents-md-design-patterns/#pattern-3-three-tier-boundaries

grgrzybek:
> @grgrzybek On Boundaries https://www.agentpatterns.ai/instructions/agents-md-design-patterns/#pattern-3-three-tier-boundaries

I fully agree. But again - I didn't want to help agent with the coding itself, I wanted it to understand what I mean by saying "Remote Jolokia Agent" in _ask mode_ only.

Do you have suggestion what is the _emerging standard_ for such high level information?

grgrzybek:
probably even `ARCHITECTURE.md` or something not explicitly mentioning _AI assistants_ would be good, because the information I put in `.ai/CONTEXT.md` is useful also for developers/users.

tadayosi:
Just for records, I went ahead and created an issue for the topic. We can discuss it further there.
- #2109

tadayosi:
> Do you have suggestion what is the _emerging standard_ for such high level information?

Then, I simply encourage you to write the architecture doc under:
https://github.com/hawtio/hawtio-react/tree/main/docs
and let your AI agent read it for your sessions.

tadayosi:
> Do you have suggestion what is the _emerging standard_ for such high level information?

I believe simply agreeing on the file name/path/format among AI vendors is super beneficial for the ecosystem including us.

AGENTS.md and MCP are two of the three _emerging standards_ by Agentic AI Foundation.

tadayosi (approved):
👍

tadayosi (approved):

PR #2124, opened 2026-06-05
fix(camel): displayed number for ExchangesTotal is wrong

tadayosi:
Fix hawtio/hawtio#4391

The number of total exchanges was calculated as the sum of completed and inflight exchanges. However, judging with `git log`, this calculation does not appear to have been particularly well thought out.

Like the other numbers, this figure should be taken directly from the route statistics.


tadayosi:
cc @joshiraezcode

grgrzybek (approved):

PR #2131, opened 2026-06-11
Theme selector

### LINKED ISSUE #2101: UI - Make light/dark themes customisable via Preferences
Preferences -> Home -> UI section can provide the theme toggle with system/light/dark options.

grgrzybek:


app/public/index.html
```diff
@@ -4,6 +4,21 @@
...
+          const prefersDark = window.matchMedia('(prefers-color-scheme: dark)').matches
+          const theme = saved ?? (prefersDark ? 'dark' : 'light')
+          if (theme === 'dark') {
+            document.documentElement.classList.add('pf-v6-theme-dark')
+          } else {
+            document.documentElement.classList.remove('pf-v6-theme-dark')
+          }
+        }
+      })()
+    </script>
```
tadayosi, line 20: out of curiosity, can't this be in the hawtio/react script?

tadayosi (approved):
awesome!

app/public/index.html
grgrzybek: This is the no-blinking trick from https://www.youtube.com/watch?v=k6fXi6G6Zow
And also I didn't change the `index.html` from hawtio/hawtio/console.
I didn't check if it works without it, but I will.

tadayosi:
It's a new "feature" that justifies 2.3.0, but we can wait for it until hawtio/hawtio#4530 is also addressed.

app/public/index.html
grgrzybek: OK - the blinking is there without this `<script>` snippet. I'd leave it - and definitely it's harmless in hawtio-react/app example.

PR #2150, opened 2026-06-29
feat(middleware): Switch from axios to node fetch() (fixes #2133)

### LINKED ISSUE #2133: Consider removing axios dependencies from @hawtio/backend-middleware
`axios` had a lot of CVE reports recently.
even if we used it only at test time...

grgrzybek:


phantomjinx (approved):
LGTM

tadayosi (approved):
awesome!
so, since it runs on node.js we only use node's native fetch and only need the ts definition for it, correct?

grgrzybek:
> awesome! so, since it runs on node.js we only use node's native fetch and only need the ts definition for it, correct?

exactly!

And I didn't perform any deep tests - I simply analyzed the migration from axios to fetch(), checked that I can normally connect to any "real" Java Jolokia Agent (in Artemis) and normally navigate the MBeans, get the attributes, invoke operations...

Security scanners will be satisfied ;)

PR #2188, opened 2026-07-28
fix(deps): upgrade react-router to v7

### LINKED ISSUE #2187: Upgrade React Router to v7
Currently, NPM audit is failing in CI due to react-router-dom v6 vulnerability:
```
$ yarn npm audit --all
└─ react-router-dom
   ├─ ID: 1124270
   ├─ Issue: React Router: Open redirect leading to XSS
   ├─ URL: https://github.com/advisories/GHSA-jjmj-jmhj-qwj2
   ├─ Severity: moderate
   ├─ Vulnerable Versions: >=6.30.2 <=6.30.4
   │
   ├─ Tree Versions
   │  └─ 6.30.4
   │
   └─ Dependents
      ├─ @hawtio/react@workspace:packages/hawtio
      └─ app@workspace:app
```

We should upgrade it to a version >= 7.13.0. React Router v8 requires React 19+ so it's not yet possible to upgrade to v8.

----
Related issues:
* github/advisory-database#8868
* github/advisory-database#9003

tadayosi:
Fix #2187


tadayosi:
we need to wait for this: https://github.com/remix-run/react-router/issues/15348

grgrzybek (approved):
👍
Let's wait for:
* remix-run/react-router#15348

grgrzybek:
BTW, I checked locally and this warning is still there:
```
You rendered descendant <Routes> (or called `useRoutes()`) at "/camel" (under <Route path="/camel">) but the parent route path has no trailing "*". This means if you navigate deeper, the parent won't match anymore and therefore the child routes will never render.

Please change the parent <Route path="/camel"> to <Route path="/camel/*">.
```

In our case this is a false positive, because we handle this by nested `<Route path='*'>`, but I think it's better to just append `/*` to plugin's `<Route>`.

EDIT: Not that easy - it leads to URLs like http://localhost:3000/hawtio/help/home/preferences/camel/jmx/logs/quartz ...

app/README.md
```diff
@@ -45,7 +45,7 @@ And finally there is one confusing, but recommended configuration - a _peer depe
 ```json
 "devDependencies": {
   "react": "^18.2.0",
-  "react-router-dom": "6.30.1"
+  "react-router": "^7.18.1"
```
grgrzybek: Should be `^7.18.2` even if it's only a readme ;)

grgrzybek (approved):

app/README.md
tadayosi: good catch 😉

Issue #2214, opened 2026-08-21
Selecting "quartz" node selects deeply nested "quartz" when it's earlier in DFS order


grgrzybek:
I had this tree:

<img width="882" height="570" alt="Image" src="https://github.com/user-attachments/assets/f73d43af-7d3c-4bb3-a05c-3c0599226930" />

While the top-level `quartz` is first in BFS order, Camel's nested `quartz` is first in DFS order and clicking the `quartz` domain was selecting `quart` component.

Issue #2198, opened 2026-08-05
Improve navigation and nid parameter handling


grgrzybek:
For now it's just a set of observations, not a strict plan to change anything...

The context for `nid` parameter support:
* #903
* #1918
* #2010

Why?
* #2197

When upgrading to ESLint 10, we've found ~70 issues related to React linting:
* defining components within `rended()` of other components
* using synchronous state changes within effects
* and most importantly: accessing refs (`useRef()`) within component rendering functions.

When I started to investigate the ref misuse, I found that it's mostly related to _selected node_ and I immediately recalled the `nid` parameter.

When testing the application I found some things that need to be improved:
* clicking between Camel and Jmx page is inconsistent - selecting a node in Jmx component reflects the selection in Camel component, but it doesn't work the opposite way
* `nid` parameter is visible in tabs/uris where it's useless
* I didn't do anything yet to handle other _inner navigation_, like preserving the selected tab in Artemis Console (which does it in its own way)
* I see Hawtio's preferences is using `NavLink` component from `react-router-dom` (will it change after #2187 ?)
* I remember Artemis is using just `Tabs`/`Tab` from Patternfly only.
* As @tadayosi mentioned before, URL is not the best place to keep values - even related to navigation. Query parameters are old school...
* I was thinking about some mapping, but `?nid=<base64>` or `?nid=<uuid>` is also not very elegant (and may not be portable)
* `?nid=org.apache.camel-CamelContexts-SampleCamel-endpoints-quartz%3A%2F%2Fcron%3Fcron%3D0%252F10%2B*%2B*%2B*%2B*%2B%253F` is simply ugly

So I have to first get rid of using refs during rendering and selected node should be a state (because it affects rendering), not a ref (which should be used to hold timeout handles or DOM references).

Here's a crude scope analysis of how Hawtio uses states, contexts and providers:
```
.JMX component:
----
<Hawtio>
  <HawtioPage>
    useUser(): username
     - 6x useState() (username, isLogin, isLoginError, loginError, userLoader, loginMethod)
    usePlugins(): plugins
     - 2x useState() (plugins, pluginsLoaded)
    useHawtconfig()
     - 2x useState() (hawtconfig, hawtConfigLoaded)
    usePluginNodeSelected()
     - 1x useState() (selectedNode)
    <PageContext.Provider value={{ username, plugins }}>
      <Page> (patternfly)
[...]

PR #2220, opened 2026-08-28
Eslint10

grgrzybek:
I know it's unacceptable, but I couldn't do it better...

It all started with ESLint 9 -> 10 upgrade and here we are.

Roughly the story was:
* upgrade ESLint to 10
* find that we no longer need `@eslint/compat`
* find that `eslint-plugin-import` is no longer maintained
* find that `eslint-plugin-react` is a leftover from CRA and https://react.dev/reference/eslint-plugin-react-hooks is the _official_ one
* the official one discovered 70+ errors (not just warnings!) in 3 categories:
  * components should not be defined in `render()` of other components - easy one
  * state should not be set synchronously in effects - this is tricky and I had to ignore some of these, but now we know where are the fishy places
  * and most importantly - refs should not be used in render()
* when I was checking the `useRef` problematic usage, I found that we use refs to keep "selected node" in the tree (Jmx and Camel)
* so I thought it should be easy to polish the trees....
* and then I found problems with:
  * nid handling: #2198
  * BFS/DFS navigation: #2214
* the `nid` parameter handling was a very big problem... See this comment: https://github.com/hawtio/hawtio-react/issues/2198#issuecomment-5440052052

And finally I think it's working fine...


packages/hawtio/src/plugins/camel/endpoints/AddEndpointWizard.tsx
```diff
@@ -25,19 +25,20 @@ const placeholder = 'Select Component Name'
 
 export const AddEndpointWizard: React.FunctionComponent = () => {
   const { selectedNode } = useContext(CamelContext)
+  const { componentName, setComponentSchema } = useContext(AddEndpointContext)
   const ctx = useContext(AddEndpointContext)
```
tadayosi, line 28: this is redundant when you've added the above line already and should be merged.

packages/hawtio/src/plugins/camel/endpoints/Endpoints.tsx
```diff
@@ -12,20 +12,21 @@ import * as es from './endpoints-service'
 export const Endpoints: React.FunctionComponent = () => {
   const { selectedNode } = useContext(CamelContext)
   const ctx = useAddEndpointContext()
+  const { setComponentNames } = ctx
```
tadayosi, line 15: this can be merged with the above line if `ctx` is not used elsewhere.

packages/hawtio/src/plugins/camel/rest-services/RestServices.tsx
```diff
@@ -127,6 +126,7 @@ export const RestServices: React.FunctionComponent = () => {
   )
 
   useEffect(() => {
+    // eslint-disable-next-line react-hooks/set-state-in-effect
```
tadayosi, line 129: is this the final fix or a todo?

packages/hawtio/src/plugins/camel/route-diagram/RouteDiagram.tsx
tadayosi, line 1: is disabling lints for the lines the only thing we can do here?
I'm not asking it to urge changes but to understand if it's a todo for the future.

packages/hawtio/src/plugins/camel/routes/Source.tsx
```diff
@@ -17,29 +17,26 @@ loader.config({ monaco })
 export const Source: React.FunctionComponent = () => {
   const { selectedNode } = useContext(CamelContext)
   const isDarkTheme = useIsDarkTheme()
-  const [xmlSource, setXmlSource] = useState('')
+  // const [xmlSource, setXmlSource] = useState('')
```
tadayosi: we can simply remove it instead of commenting out.

packages/hawtio/src/plugins/camel/CamelTreeView.test.tsx
```diff
@@ -67,9 +68,11 @@ describe('CamelTreeView', () => {
...
 
     render(
-      <CamelContext.Provider value={{ tree, selectedNode, setSelectedNode }}>
-        <CamelTreeView />
-      </CamelContext.Provider>,
+      <MemoryRouter>
+        <CamelContext.Provider value={{ tree, selectedNode, setSelectedNode }}>
+          <CamelTreeView />
+        </CamelContext.Provider>
+      </MemoryRouter>,
```
tadayosi, line 75: so now the router is nested for Camel plugin: BrowserRouter -> MemoryRouter?
why do we need it only for the Camel plugin?

tadayosi (approved):

packages/hawtio/src/plugins/camel/exchanges/BlockedExchanges.tsx
```diff
@@ -18,8 +43,10 @@ export const BlockedExchanges: React.FunctionComponent = () => {
    * setTimeout will cache the original state so need to use a ref
    * to allow fetchExchanges() to retrieve the latest value
    */
-  const isConfirmUnblockOpenRef = useRef(isConfirmUnblockOpen)
-  isConfirmUnblockOpenRef.current = isConfirmUnblockOpen
+  const isConfirmUnblockOpenRef = useRef<boolean>()
+  if (isConfirmUnblockOpenRef.current == null) {
+    isConfirmUnblockOpenRef.current = isConfirmUnblockOpen
+  }
```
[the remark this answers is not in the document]
grgrzybek: copilot is suggesting to actually use previous version.
`isConfirmUnblockOpenRef.current = isConfirmUnblockOpen` causes:
```
ESLint:
Error: Cannot access refs during render
```

I'll change to:
```typescript
const isConfirmUnblockOpenRef = useRef(isConfirmUnblockOpen)
if (isConfirmUnblockOpenRef.current == null) {
  isConfirmUnblockOpenRef.current = isConfirmUnblockOpen
}
```

packages/hawtio/src/plugins/jmx/context.ts
```diff
@@ -1,113 +1,154 @@
...
-    return () => eventService.removeListener(EVENT_REFRESH, listener)
+    return () => {
+      apply = false
+      eventService.removeListener(EVENT_REFRESH, listener)
+    }
   }, [])
 
+  // another effect synchronizes the `nid` query parameter and the global state for "selected node"
+  useEffect(() => {
+    const nid = searchParams.get('nid') ?? new URLSearchParams(window.location.href).get('nid')
```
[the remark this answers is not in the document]
grgrzybek: good catch, I'll fix this. My mistake.

packages/hawtio/src/plugins/camel/context.ts
```diff
@@ -1,138 +1,274 @@
...
+    return () => {
+      apply = false
+      eventService.removeListener(EVENT_REFRESH, listener)
+      workspace.clearExpandAll(jmxDomain)
+    }
+  }, [populateTree])
+
+  // synchronize the `nid` query parameter and the global state for "selected node"
+  useEffect(() => {
+    const nid = searchParams.get('nid') ?? new URLSearchParams(window.location.href).get('nid')
```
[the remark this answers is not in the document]
grgrzybek: same as with `jmx/context.ts`

packages/hawtio/src/plugins/shared/workspace.ts
```diff
@@ -219,9 +392,91 @@ class Workspace implements IWorkspace {
...
+        this.expandedFolders.clear()
+      }
+    } else {
+      if (expanded) {
+        this.expandedFolders = this.expandedFolders.union(new Set(this.domainFolders.get(domain)!.keys()))
+        this.domainExpanded.set(domain, ExpansionValue.ExpandAll)
+      } else {
+        this.expandedFolders = this.expandedFolders.difference(new Set(this.domainFolders.get(domain)!.keys()))
+        this.domainExpanded.set(domain, ExpansionValue.CollapseAll)
+      }
```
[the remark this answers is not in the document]
grgrzybek: @copilot, according to MDN:
> Since June 2024, this feature works across the latest devices and browser versions.

you need to refresh your model.

packages/hawtio/src/plugins/camel/endpoints/AddEndpointWizard.tsx
grgrzybek: The reason is:
> ESLint: React Hook useEffect has a missing dependency: 'ctx'. Either include it or remove the dependency array. (react-hooks/exhaustive-deps)

the problem is that you shouldn't use the _object_ as dependency for an effect, because the content of the object doesn't change, but the wrapping object returned from useContext() is always new and triggers `!Object.is()` used by React.

See more explanation in conversation for `packages/hawtio/src/plugins/camel/endpoints/Endpoints.tsx`.

packages/hawtio/src/plugins/camel/endpoints/Endpoints.tsx
grgrzybek: The ESLint error was:
> ESLint: React Hook useEffect has a missing dependency: 'ctx'. Either include it or remove the dependency array. (react-hooks/exhaustive-deps)

I found that the object returned by `useAddEndpointContext()` is ALWAYS new, while containing state-related fields like `setComponentName`.

Adding `ctx` alone to the hooks dependency list was triggerring the effect continuously, because `ctx` was always new (as of `Object.is()`).

To satisfy the linter and the intent, I added pure function as the dependency, so I needed to deconstruct it.

Sure - `ctx.setComponentNames` should be fine, but linter still complains about missing `ctx`.

packages/hawtio/src/plugins/camel/rest-services/RestServices.tsx
grgrzybek: It's kind of "light TODO". I didn't want to refactor the complicated code from Paul added in 2023, because Camel plugin is quite complicated and I didn't want to mess too much with it.

We should remember to have a look at the comments with `eslint-disable-next-line` at some point.

packages/hawtio/src/plugins/camel/CamelTreeView.test.tsx
grgrzybek: no no - this is only in test.tsx file. And this was needed _in test_, because the tested component uses react-router hooks which require to be run within router context.

And the easiest way to provide such context was with MemoryRouter. Only in test. And there's no related `JmxTreeView.test.tsx`.

packages/hawtio/src/plugins/camel/route-diagram/RouteDiagram.tsx
grgrzybek: I disabled the linting only (mostly) in Camel plugin because in order to fix I'd have to involve Paul's memory about why it was done like this in the first place ;)

At some point, when I understand more about how to clean up the state based on another state, I'll get back to it. For now I just marked the lines with warning-disable to remember to get back to it.

tadayosi (approved):

